PR #5, opened 2016-12-01
Replace js-nacl with sodium library

4miners:
Reference: https://github.com/LiskHQ/lisk/pull/321


karmacoma:
Unfortunately, we can not merge as we need to support both server and client environments. The current `js-nacl` implementation allows for this, but `libsodium` will not.

PR #11, opened 2017-02-02
Browserify grunt

Tosch110:
Adding Browserify, Grunt and ESLint for ECMA 5.

Browserify and ESLint can either be executed seperately:
grunt browserify
grunt eslint

The standard grunt will do both:
grunt


Gruntfile.js
```diff
@@ -0,0 +1,22 @@
...
+        },
+        pkg: grunt.file.readJSON('package.json'),
+        browserify: {
+            'app.js': ['index.js']
+        }
+    });
+
+
+    grunt.loadNpmTasks('grunt-browserify');
+    grunt.registerTask("default", ["browserify", "eslint"]);
```
karmacoma: Single quotes please.

package.json
```diff
@@ -20,6 +21,13 @@
     "js-nacl": "LiskHQ/js-nacl#6dc1417"
   },
   "devDependencies": {
```
karmacoma: Please use `=` operator on dependencies.

Issue #10, opened 2017-02-02
Add test coverage metric


karmacoma:
We should use https://istanbul.js.org/ to measure the amount of test coverage.

Tosch110:
This is being done. Later we can add travis to work together with istanbul and coveralls.io

https://istanbul.js.org/docs/tutorials/mocha/

PR #14, opened 2017-02-03
Adding test coverage metric with eslint refinements

Tosch110:
Integrating Istanbul coverage metric.

Updating package.json dependencies.

As mentioned in Issue #10

Refining rules defined in `eslint_ecma5.json`.

### DIFF
--- a/eslint_ecma5.json
+++ b/eslint_ecma5.json
@@ -10,7 +10,7 @@
     "camelcase": [2, {"properties": "always"}],
     "eqeqeq": 2,
     "no-eq-null": 1,
-    "indent": ["error", 2],
+    "indent": ["error", "tab"],
     "no-use-before-define": 1,
     "new-cap": 1,
     "quotes": ["error", "single", { "avoidEscape": true }],
--- a/eslint_ecma5.json
+++ b/eslint_ecma5.json
@@ -7,14 +7,14 @@
     }
   },
   "rules": {
-    "camelcase": [2, {"properties": "always"}],
-    "eqeqeq": 2,
-    "no-eq-null": 1,
+    "camelcase": ["error", {"properties": "always"}],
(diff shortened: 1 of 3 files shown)


eslint_ecma5.json
```diff
@@ -0,0 +1,45 @@
...
+      "jsx": false
+    }
+  },
+  "rules": {
+    // Inspired by airbnb rules: https://github.com/airbnb/javascript/blob/master/linters/SublimeLinter/SublimeLinter.sublime-settings
+    // Force all variable names to use either camelCase style or UPPER_CASE with underscores.
+    "camelcase": [2, {"properties": "always"}],
+
+    // Prohibit use of == and != in favor of === and !==.
+    "eqeqeq": 2,
```
MaciejBaj: in a case of consistency, I think it is better to keep one convention in actions naming, (there is 
`"eqeqeq": 2` and after that `"indent": ["error", "tab"]`). I think that the second notation is more readable, so `"eqeqeq": "error"`

eslint_ecma5.json
```diff
@@ -0,0 +1,45 @@
...
+
+    // Suppress warnings about == null comparisons.
+    "no-eq-null": 1,
+
+    // Enforce tab width of 2 spaces.
```
MaciejBaj: description: enforce tab width of 2 spaces. -> tabs

package.json
```diff
@@ -20,6 +21,14 @@
     "js-nacl": "LiskHQ/js-nacl#6dc1417"
   },
   "devDependencies": {
+    "eslint": "=3.14.1",
```
MaciejBaj: unused if we are using grunt-eslint

eslint_ecma5.json
4miners: I like 4 spaces ;)

PR #18, opened 2017-02-13
add create Lisk Transaction instead of 'createTransaction' which migh…

Tosch110:
…t be confusing for other transactions


karmacoma:
Let's review API design later as one overall task. We will maintain current API for now.

PR #46, opened 2017-03-09
Fix coveralls

Tosch110:


karmacoma:
We should not include `repos_token`. Token regenerated.

PR #51, opened 2017-03-10
Increase test coverage

Tosch110:


lib/api/liskApi.js
```diff
@@ -45,8 +45,8 @@ function LiskAPI (options) {
 
 	this.options = options;
 	this.ssl = options.ssl || false;
+	this.noRandomNode = options.noRandomNode || false;
 	this.testnet = options.testnet || false;
-	this.autoFindNode = options.autoFindNode || true;
```
karmacoma: `noRandomNode` Not sure about negated logic.

`autoFindNode` was better in my opinion. Or: `randomPeer = true` for example.

```
if (this.randomPeer) {
```

Using `peer` instead of `node` is also more consistent with rest of code.

Tosch110: Totally agree! I just was not able to find such a good name :)

Thanks Oliver!

Tosch110:
Changes as requested are integrated. This is now called "randomPeer". Additionally, renaming all "node" variables to "peer"

PR #59, opened 2017-03-24
Removing object wrapper (es6 feature) from transaction object

Tosch110:


slaweet:
There is still one more object wrapper:
https://github.com/LiskHQ/lisk-js/blob/transaction-object-removal/lib/api/parseTransaction.js#L141
and removing it breaks the tests.

And the next ES6 thing that Lisk Nano tests are complaining about is
```
Object.assign({}, { transport: transport }, defaultsOptions)
```
which is in 
https://raw.githubusercontent.com/LiskHQ/lisk-js/transaction-object-removal/dist/lisk-js.js
though I'm not sure where does it come from

Issue #64, opened 2017-03-27
Send LSK doesn't send


senikk:
The reason is the body is missing from request sent. I will post pull request with fix.

Issue #57, opened 2017-03-20
Documentation Show/Hide button does not trigger on Send LSK


Tosch110:
See topic. http://liskhq.github.io/lisk-js/example/experiment.html#send_lsk

PR #65, opened 2017-03-27
fixing #57 and #64

senikk:
- [x] Fixing show/hide - it was used several elements with the same id not allowed in html
- [x] Fixing Send LSK - the options for the transaction request was missing in the popsicle.request body


karmacoma:
Thank you very much @senikk 👍

Tosch110:
So great! Thanks for your help @senikk !

Issue #60, opened 2017-03-27
lisk.api().listTransactions(address, callback) send unexpected random param


slaweet:
steps to replicate:
> address = '16313739661670634666L'
> callback = function (data) { console.log(data);}
> lisk.api({node:'localhost', port: 4000}).listTransactions(address, callback);
VM332:1 Object {success: false, error: "Failed to get transactions: Parameter is not supported: random"}
> lisk.api().listTransactions(address, callback);
VM332:1 Object {success: false, error: "Failed to get transactions: Parameter is not supported: random"}

karmacoma:
Closed by #66

PR #66, opened 2017-03-28
Add missing API features - Fixes #61, #60

### LINKED ISSUE #61: Implement missing features for Lisk Nano
- [x] add `limit` parameter to `lisk.api().listTransactions(address, callback);`
- [x] create a new api method `lisk.api().listPeers();` that returns list of peers for the peer dropdown. It actually contains 3 lists: official, testnet, localhost.
- [ ] provide an API that periodically checks if the active peer is online and calls a callback if peer becomes offline and/or tries to connect to another peer and informs about that in the callback. It could be something like `lisk.api().keepOnline(callback);`. 
- [x] create a new api method `lisk.api().getAccount(address);` that calls `/api/accounts`
- [x] get rid of object wrappers like `{ transaction }`
- [x] create a new api method `lisk.api().sendTransaction(address, amount, passphrase, secondPassphrase);` that creates and sends transaction
- [x] add `orderBy = 'timestamp:desc'` parameter to `lisk.api().listTransactions()`

More on the proposed `lisk.api().keepOnline(callback);`:
Right now Lisk Nano is using /api/loader/status/ call for this, but it also calls two other requests with the same frequency (every ~10 seconds): `/api/accounts` and `/api/transactions`. So lisk-js can be smarter and avoid the `/api/loader/status` call if there was some other successful call recently. In addition, the `/api/transactions` call could be cached and the cache purged only if `balance` or `unconfirmedBalance` in `/api/accounts` changed, provided that `/api/accounts` is called periodically.

Tosch110:


test/api/liskApi.js
```diff
@@ -550,4 +535,38 @@ describe('Lisk.api()', function () {
 			});
 		});
 	});
-});
+
+	describe('#getAccount', function() {
+		it('should get account information', function (done) {
```
karmacoma: `should get account information`

Maybe: `should get an account using a given address`. Please review sibling descriptions for: `#getBlock`, `#getTransaction`.

PR #90, opened 2017-04-06
Adding browser tests - Fixes #89

### LINKED ISSUE #89: Implement tests for the Browser
Tests are currently all running on Node.
The same tests should also be done in the Browser, to get sure it behaves as expected.

Tosch110:
Including tests for the browser in the new folder /browsertests

The browsertest.html can be opened in the Browser and the tests will automatically start running.

Currently all tests are passing for: (tested today)
- Chrome 57.0.2987.133 (64-bit)
- Firefox 52.0 (64-bit)
- Safari 10.0.3 (12602.4.8)

The Browsertest - browsertest.html use the distribution file in /dist/lisk-js.js
The Browsertest - browsertest.min.html use the distribution file in /dist/lisk-js.min.js


browsertest/browsertest.html
```diff
@@ -0,0 +1,47 @@
...
+</head>
+<body>
+
+<div id="mocha"><p><a href=".">Index</a></p></div>
+<div id="messages"></div>
+<div id="fixtures"></div>
+
+<script src="vendor/mocha.js"></script>
+<script src="vendor/chai.js"></script>
+<script src="vendor/should.js"></script>
```
karmacoma: I would prefer if we can include these vendor dependencies via some CDN. Then remove the included source from this PR.

PR #95, opened 2017-04-12
Complete multisignature module - Closes #88

### LINKED ISSUE #88: Complete multisignature module
Multisignature in the current behaviour is not complete in lisk-js.

Missing:

- [ ] Add signature to multisig transaction
- [ ] Route API requests
- [ ] Test verification of multisignature signatures

Tosch110:
Implementing missing / incomplete multisignature features into `lisk-js`

Contains:
- Refactoring of getBytes function to make it better readable, cover all transaction types and redoing the multisignature transaction Byte function
- Adapting the multisignature fees
- Getting multisignature inputs to the right core endpoint and in the right format

Closes #88


test/transactions/multisignature.js
```diff
@@ -94,83 +90,20 @@ describe('multisignature.js', function () {
 	describe('#signTransaction', function () {
```
karmacoma: I think the description of `signTransaction` should be like so. In order to account for object returned with two properties: `transaction` and `signature`.

```js
describe('#signTransaction', function () {

	var secret = '123';
	var transaction = lisk.transaction.createTransaction('58191285901858109L', 1000, 'secret');
	var signTransaction = multisignature.signTransaction(transaction, secret);

	it('should return an object', function () {
		(signTransaction).should.be.type('object');
	});

	it('should have a transaction property', function () {
		(signTransaction.transaction).should.be.type('string');
		(signTransaction.transaction).should.be.equal(transaction.id);
	});

	it('should have a signature property', function () {
		var length = 128; // crypto_sign_BYTES length

		(signTransaction.signature).should.be.type('string');
		(signTransaction.signature).should.have.lengthOf(length);
	});
});
```

Tosch110: The changes you have made to the tests are good, thanks! 👍

PR #97, opened 2017-04-12
Finish YUIdoc for new functions - Closes #71

### LINKED ISSUE #71: Finish YUIdoc for new functions
API wrapper and parseTransaction are not yet included in YUIdoc. This should be fixed.

Tosch110:
Closes #71


lib/api/parseTransaction.js
```diff
@@ -23,6 +23,16 @@ LiskJS.transaction = require('../transactions/transaction');
 LiskJS.transfer = require('../transactions/transfer');
 LiskJS.vote = require('../transactions/vote');
 
+/**
+ * This is the description for the `Lisk ParseOfflineRequest` method.
```
karmacoma: Again, can we give an actual description like other modules?

Something like:

> ParseOfflineRequest module provides automatic routing of new transaction requests which can be signed locally, and then broadcast without any passphrases being transmitted.

lib/api/liskApi.js
```diff
@@ -13,6 +13,27 @@
  *
  */
 
+/**
+ * This is the description for the `Lisk API` module.
```
karmacoma: Can we give an actual description like other modules? Following the existing pattern in other modules.

> LiskAPI module provides functions for interfacing with the Lisk network. Providing mechanisms for:
> 
> - Retrieval of blockchain data: accounts, blocks, transactions.
> - Local signing of transactions and immediate network transmission.

Something like that.

lib/api/parseTransaction.js
Tosch110: Yes, nice summary Oliver, thanks!

PR #122, opened 2017-05-08
verifyMessageWithPublicKey should rather “throw” errors then “return”…

slaweet:
There is a temporary workaround for this in Lisk Nano:
https://github.com/LiskHQ/lisk-nano/pull/184/commits/1edd820ca57d1366ac51bb1bdbd099c7e9969c2e


test/transactions/crypto/index.js
```diff
@@ -141,16 +141,20 @@ describe('crypto/index.js', function () {
...
 		it('should detect invalid publicKeys', function () {
 			var invalidPublicKey = keypair.publicKey + 'ERROR';
-			var invalidVerifyMessage = newcrypto.verifyMessageWithPublicKey(signedMessage, invalidPublicKey);
-
-			(invalidVerifyMessage.message).should.be.equal('Invalid publicKey, expected 32-byte publicKey');
+			try {
+			  newcrypto.verifyMessageWithPublicKey(signedMessage, invalidPublicKey);
+			} catch (error) {
+				(error.message).should.be.equal('Invalid publicKey, expected 32-byte publicKey');
+			}
```
karmacoma: Why not use: http://chaijs.com/api/bdd/#method_throw?

PR #138, opened 2017-06-22
Implement bip39 account generation - Closes #133

### LINKED ISSUE #133: Implement bip39 account generation
Currently, only known accounts are usable with lisk-js. We need a way to generate new 12-24 word passphrases in lisk-js.

alepop:
- Add mnemonic module 
- Add API method for generating new Account

In this implementation, I use some code from `lisk-nano` for generating a random seed. And take functionality from `bitcore-mnemonic` package. Why don't use `bitcore-mnemonic` package instead?
Because there too many dependencies.
Close #133

```
└─┬ bitcore-mnemonic@1.2.5
  └─┬ bitcore-lib@0.14.0
    ├── bn.js@2.0.4
    ├── bs58@2.0.0
    ├── buffer-compare@1.0.0
    ├─┬ elliptic@3.0.3
    │ ├── bn.js@2.0.4  deduped
    │ ├── brorand@1.0.5
    │ ├─┬ hash.js@1.0.3
    │ │ └── inherits@2.0.1  deduped
    │ └── inherits@2.0.1  deduped
    ├── inherits@2.0.1
    └── lodash@3.10.1
```


lib/api/liskApi.js
```diff
@@ -539,6 +539,20 @@ LiskAPI.prototype.getAccount = function (address, callback) {
...
 /**
+ * @method generateAccount
+ * @param secret
+ * @param callback
+ *
+ * @return API object
+ */
+
+LiskAPI.prototype.generateAccount = function (secret, callback) {
+	this.sendRequest('accounts/generatePublicKey', { secret: secret }, function (result) {
```
Tosch110: Please use the cryptographic function instead of an API call:
https://github.com/LiskHQ/lisk-js/blob/development/lib/transactions/crypto/keys.js

lib/api/liskApi.js
```diff
@@ -547,9 +547,9 @@ LiskAPI.prototype.getAccount = function (address, callback) {
  */
 
 LiskAPI.prototype.generateAccount = function (secret, callback) {
-	this.sendRequest('accounts/generatePublicKey', { secret: secret }, function (result) {
-		return callback(result);
-	});
+	var keys = LiskJS.crypto.getPrivateAndPublicKeyFromSecret(secret);
+	callback(keys)
```
Tosch110: The callback is not necessary, as stated the function is synchronous, you can just return it.

alepop: But all API methods receive the callback as the second parameter and return `this` and I follow that design.

PR #143, opened 2017-06-29
Expose options and peers to config file - Closes #111

### LINKED ISSUE #111: Expose peers and options to a configuration file
Current behaviour:
Peers and options to run lisk-js are embedded in the script (liskApi.js). 

Expected behaviour:
Peers and additional options can be included in the repository in a configurable json file.
This way users and developers can have an easier access to configure (default) lisk-js to their needs.

alepop:
- Move default peers and options to the `config.json` file.
- Use it as default params and expose it with passed `options` params.

Closes #111


karmacoma:
Thank you @alepop 👍

PR #177, opened 2017-07-05
Adds signing and verifying of messages with two secrets - Closes #176

### LINKED ISSUE #176: Signing messages with secondPassphrase not supported
Current behaviour:

Lisk-JS lets a user sign messages with a passphrase. Lisk allows to create secondPassphrases in order to secure the account. This is not reflected in the signing process.

Expected behaviour:

Lisk-JS allows to create signature signed by both passphrases.

Solution proposal:

Instead of using just simple two signatures or concatenate signatures, the process should be the following:

**Signing**

1) Create a signature with first passphrase
2) Use the 1) signature as message to create a new signature with secondPassphrase

**Verifying**

3) Use the publicKey of 2nd passphrase to unwrap the first signature created by 1)
4) Use the publicKey of 1st passphrase to unwrap the original message

Tosch110:
Closes #176


lib/transactions/crypto/sign.js
```diff
@@ -27,6 +27,49 @@ function signMessageWithSecret (message, secret) {
...
+
+	if (openSignature) {
+
+		var openSecondSignature = naclInstance.crypto_sign_open(openSignature, publicKeyBytes);
+
+		if(openSecondSignature) {
+			// Returns original message
+			return naclInstance.decode_utf8(openSecondSignature);
+		} else {
+			throw new Error('Invalid signature publicKey2 combination, cannot verify message');
```
willclarktech: Can we have a consistent term for this (`publicKey2` vs `secondPublicKey`)?

lib/transactions/crypto/sign.js
```diff
@@ -27,6 +27,49 @@ function signMessageWithSecret (message, secret) {
...
+		}
+
+	} else {
+		throw new Error('Invalid signature publicKey1 combination, cannot verify message');
```
willclarktech: Likewise.

test/transactions/crypto/index.js
```diff
@@ -249,4 +249,34 @@ describe('crypto/index.js', function () {
...
+		});
+	});
+
+	describe('#verifyMessageWithTwoPublicKeys sign.js', function () {
+
+		it('should verify a message using two publicKeys', function () {
+			var secret = '123';
+			var secondSecret = '1234';
+			var message = 'Hello.';
+			var signature = newcrypto.signMessageWithTwoSecrets(message, secret, secondSecret);
```
willclarktech: Prefer using a static signature rather than depending on separate functionality.

lib/transactions/crypto/sign.js
```diff
@@ -27,6 +27,49 @@ function signMessageWithSecret (message, secret) {
...
+	var signedMessage = naclInstance.crypto_sign(msg, keypair.privateKey);
+	var doubleSignedMessage =  naclInstance.crypto_sign(signedMessage, secondKeypair.privateKey);
+
+	var hexSignedMessage = convert.bufferToHex(doubleSignedMessage);
+
+	return hexSignedMessage;
+
+}
+
+function verifyMessageWithTwoPublicKeys (signedMessage, publicKey, secondPublicKey) {
```
Tosch110: Yes, the signature is assumed to be always hex format

PR #183, opened 2017-07-10
Use should.js doesNotThrow method - Closes #169

### LINKED ISSUE #169: Use should.js's .throw assertion
We can avoid a lot of boilerplate like this. E.g.
```
it('should not accept falsy options like undefined', function (done) {
	try {
		lisk.api().sendRequest('delegates/', {limit:undefined}, function () {});
	} catch (e) {
		(e.message).should.be.equal('parameter value "limit" should not be undefined');
		done();
	}
});
```
can become
```
it('should not accept falsy options like undefined', function () {
	(function() { lisk.api().sendRequest('delegates/', {limit:undefined}, function () {}) })
		.should.throw('parameter value "limit" should not be undefined');
});
```

willclarktech:
Closes #169


test/transactions/dapp.js
```diff
@@ -89,43 +89,38 @@ describe('dapp.js', function () {
...
-				(trs.id).should.be.type('string');
+				(trs).should.have.property('id').and.be.type('string');
 			});
 
 			it('should have type as number and equal 9', function () {
-				(trs.type).should.be.type('number').and.equal(5);
+				(trs).should.have.property('type').and.be.type('number').and.equal(5);
 			});
 
 			it('should have amount as number and eqaul 0', function () {
```
Isabello: Spelling mistake. 'equal'

PR #190, opened 2017-07-11
Standardise test setup - Closes #189

### LINKED ISSUE #189: Tidy up test common.js
Currently we have exports and globals that are used inconsistently.

willclarktech:
Closes #189 
Removing the common import also makes it very clear when our tests are not atomic for future refactors.

I had to change some of the hardcoded public keys to get the tests passing again, which I don't currently understand.

### DIFF
--- a/test/api/liskApi.js
+++ b/test/api/liskApi.js
@@ -1,28 +1,24 @@
-if (typeof module !== 'undefined' && module.exports) {
-	var common = require('../common');
-	var lisk = common.lisk;
-	var privateApi = common.privateApi;
-	var utils = common.utils;
-	var sinon = common.sinon;
-	process.env.NODE_ENV = 'test';
-}
+var liskApi = require('../../lib/api/liskApi');
+var privateApi = require('../../lib/api/privateApi');
+var utils = require('../../lib/api/utils');
+var transactionModule = require('../../lib/transactions/transaction');
 
 describe('Lisk.api()', function () {
 
-	var LSK = lisk.api();
+	var LSK = liskApi();
 
-	describe('lisk.api()', function () {
+	describe('liskApi()', function () {
 
-		it('should create a new instance when using lisk.api()', function () {
+		it('should create a new instance when using liskApi()', function () {
 			(LSK).should.be.ok();
 		});
 
-		it('new lisk.api() should be Object', function () {
+		it('new liskApi() should be Object', function () {
 			(LSK).should.be.type('object');
 		});
 
 		it('should use testnet peer for testnet settings', function () {
-			var TESTLSK = lisk.api({ testnet: true });
+			var TESTLSK = liskApi({ testnet: true });
 
 			(TESTLSK.port).should.be.equal(7000);
 			(TESTLSK.testnet).should.be.equal(true);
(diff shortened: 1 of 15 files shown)


willclarktech:
Grunt is failing because of an es6 feature [here](https://github.com/LiskHQ/lisk-js/blob/189-test_common/lib/utils/mnemonic.js#L44), which is unrelated to this PR, but will be fixed anyway in https://github.com/LiskHQ/lisk-js/pull/194.

test/api/liskApi.js
```diff
@@ -1,28 +1,24 @@
-if (typeof module !== 'undefined' && module.exports) {
```
Tosch110: This part was especially to have the tests also running in a browser. Did you verify if tests in the browser still work? I cannot test right now because the grunt is not working as you mentioned. You can find those tests in browsertest/. It would be good to maintain it because lisk-js is expected to run in nodejs and the browser

willclarktech: I realised that later, and I've created #202 to rethink this. Browser tests are currently broken anyway because they don't have the sinon script from a cdn.

Also, now we have babel compiling our code, I think a more sensible approach is to test against `dist` in the browser, than to force browser testers to have modern es8-friendly browsers. That way the module issue we have here doesn't even arise.

test/api/liskApi.js
```diff
@@ -1,28 +1,24 @@
...
-		it('new lisk.api() should be Object', function () {
+		it('new liskApi() should be Object', function () {
 			(LSK).should.be.type('object');
 		});
 
 		it('should use testnet peer for testnet settings', function () {
-			var TESTLSK = lisk.api({ testnet: true });
+			var TESTLSK = liskApi({ testnet: true });
 
 			(TESTLSK.port).should.be.equal(7000);
```
Isabello: this should be defined as a variable, as testnet does not mean 7000 on end user machines

test/api/liskApi.js
```diff
@@ -130,14 +126,14 @@ describe('Lisk.api()', function () {
 	describe('#selectNode', function () {
 
 		it('should return the node from initial settings when set', function () {
-			var LiskUrlInit = lisk.api({ port: 7000, node: 'localhost', ssl: true, randomPeer: false });
+			var LiskUrlInit = liskApi({ port: 7000, node: 'localhost', ssl: true, randomPeer: false });
```
Isabello: same for all of these hardcodes

test/api/liskApi.js
```diff
@@ -480,7 +476,7 @@ describe('Lisk.api()', function () {
 		it('should give the correct parameters for PUT requests', function () {
 			var requestType = 'signatures';
 			var options = {secret: '123', secondSecret: '1234'};
-			var LSK = lisk.api({ node: 'localhost' });
+			var LSK = liskApi({ node: 'localhost' });
```
Isabello: localhost should be made a variable as well.

test/setup.js
```diff
@@ -1,7 +1,8 @@
-// var Buffer = require('buffer/').Buffer;
-var cryptoLib = require('crypto-browserify');
 var should = require('should');
 var sinon = require('sinon');
+var naclFactory = require('js-nacl');
+
+process.env.NODE_ENV = 'test';
```
Isabello: why declare this inside the setup file instead of as part of the npm run?

willclarktech: Paranoia in case someone runs this outside an npm script.

Isabello: :D

PR #195, opened 2017-07-17
Use offsets in transaction creation - Closes #191

### LINKED ISSUE #191: Mitigate "Timestamp is in the future" error
When user's OS time is a few seconds ahead, all transactions are rejected with "Invalid transaction timestamp. TimeStamp is in the Future" error message.

When this error is obtained in lisk-js, it could try to do the transaction again with timestamp 10 seconds in the past. Maybe try to repeat this step 3 times with 10, 20, 30 seconds?

There are several issues on Lisk Nano reporting this: https://github.com/LiskHQ/lisk-nano/issues/485 https://github.com/LiskHQ/lisk-nano/issues/458 https://github.com/LiskHQ/lisk-nano/issues/365
I heard even more on Lisk Chat, so I consider this to be a big problem. It happens mainly on Windows due to its lack of time sync.

willclarktech:
Closes #191

### DIFF
--- a/lib/api/liskApi.js
+++ b/lib/api/liskApi.js
@@ -168,10 +168,15 @@ LiskAPI.prototype.sendRequest = function (requestType, options, callback) {
 		}
 		return returnAnswer;
 	}).then(function (result) {
-		if (!result.success && result.message.match(/Timestamp is in the future/)) {
+		if (!result.success && result.message.match(/Timestamp is in the future/) && !(options.timeOffset > 40e3)) {
+
+			var newOptions = {};
+			Object.keys(options)
+				.forEach(key => newOptions[key] = options[key]);
+			newOptions.timeOffset = (options.timeOffset || 0) + 10e3;
 			return new Promise(function (resolve, reject) {
 				setTimeout(function () {
-					that.sendRequest(requestType, options, callback)
+					that.sendRequest(requestType, newOptions, typeof callback === 'function' ? callback : null)
 						.then(resolve, reject);
 				}, 1000);
 			});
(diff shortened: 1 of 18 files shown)


lib/api/liskApi.js
```diff
@@ -163,22 +163,35 @@ LiskAPI.prototype.sendRequest = function (requestType, options, callback) {
 	return privateApi.sendRequestPromise.call(this, requestType, options).then(function (requestSuccess) {
 		var returnAnswer = (parseOfflineRequest(requestType, options).requestMethod === 'GET') ? requestSuccess.body : parseOfflineRequest(requestType, options).transactionOutputAfter(requestSuccess.body);
 
-		if (!callback || (typeof callback !== 'function')) {
-			return Promise.resolve(returnAnswer);
-		} else {
-			return callback(returnAnswer);
+		if (callback && (typeof callback === 'function')) {
```
slaweet: You call `callback` before checking for "Timestamp is in the future". The `callback` should be called only if you don't do another api call.

slaweet: Also,`LiskAPI.prototype.sendRequest` function is at the size when it might be worth to extract some part into a separate function.

lib/api/liskApi.js
```diff
@@ -163,22 +163,35 @@ LiskAPI.prototype.sendRequest = function (requestType, options, callback) {
...
+
+			var newOptions = {};
+			Object.keys(options)
+				.forEach(key => newOptions[key] = options[key]);
+			newOptions.timeOffset = (options.timeOffset || 0) + 10e3;
+			return new Promise(function (resolve, reject) {
+				setTimeout(function () {
+					that.sendRequest(requestType, newOptions, typeof callback === 'function' ? callback : null)
+						.then(resolve, reject);
+				}, 1000);
```
slaweet: Do we need this timeout?

lib/api/liskApi.js
willclarktech: Good catch.

lib/api/liskApi.js
willclarktech: Probably not. I put it in before limiting the number of requests that would be made but I've removed it now.

Issue #178, opened 2017-07-05
Make tests atomic


willclarktech:
E.g. [this suite](https://github.com/LiskHQ/lisk-js/blob/development/test/transactions/delegate.js#L39) depends on [this test](https://github.com/LiskHQ/lisk-js/blob/development/test/transactions/delegate.js#L35). It would be better to have a `beforeEach` hook or similar so that individual tests can be run.

- [x] Separate test files for separate api modules - #267 - MERGED
- [x] Dapp transaction - #268 - MERGED
- [x] Delegate transaction - #269 - MERGED
- [x] Multisignature transaction - #270 - MERGED
- [x] Signature transaction - #271 - MERGED
- [x] Transaction transaction - #272 - MERGED
- [x] Transfer transaction - #273 - MERGED
- [x] Vote transaction - #274 - MERGED
- [x] Crypto - #275
- [x] Time - #276 - MERGED
- [x] LiskAPI - #277  - MERGED
- [x] Private API - #278 - MERGED
- [x] API utils - #279 - MERGED
- [x] Transaction utils - #280 - MERGED
- [x] Mnemonic - #281 - MERGED
- [x] Constants (shape only) - #352 - MERGED
- [x] Make sure all transaction tests have an id check - #353 - MERGED
- [x] Remove crypto stubbing - #354 - MERGED

PR #200, opened 2017-07-18
Add functions to create dapp deposits/withdrawals - Closes #145

### LINKED ISSUE #145: Create dapp deposit and withdraw transactions
Related to https://github.com/LiskHQ/lisk/issues/560 and https://github.com/LiskHQ/lisk/issues/457 and #106 

Clarifying this issue a bit about Core's needs:

Two new functions, One for `inTransfer` and one for `outTransfer`

Each should accept an amount in the code from the passed variables.

Historically, dappIn and dappOut logic was done within core, which is why transfer.js was left in the state that it is in now.

willclarktech:
~~Merge #199 first~~
Closes #145

### DIFF
--- a/test/transactions/transfer.js
+++ b/test/transactions/transfer.js
@@ -23,41 +23,58 @@ describe('transfer.js', function () {
 		var dappId = '1234213';
 		var amount = 10e8;
 		var secret = 'secret';
+		var publicKey = '5d036a858ce89f844491762eb89e2bfbd50a4a0a0da658e4b2628b25b117ae09';
 		var secondSecret = 'secondSecret';
-		var transferTransaction = createInTransfer(dappId, amount, secret, secondSecret);
+		var inTransferTransaction = createInTransfer(dappId, amount, secret, secondSecret);
 
 		it('should be a function', function () {
 			(createInTransfer).should.be.type('function');
 		});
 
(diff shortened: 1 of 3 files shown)


test/transactions/transfer.js
```diff
@@ -16,42 +15,168 @@ describe('transfer.js', function () {
...
+	describe('#createOutTransfer', function () {
+		var createOutTransfer = transfer.createOutTransfer;
+		var dappId = '1234213';
+		var transactionId = '9876567';
+		var recipientId = '989234L';
+		var amount = 10e8;
+		var secret = 'secret';
+		var publicKey = '5d036a858ce89f844491762eb89e2bfbd50a4a0a0da658e4b2628b25b117ae09';
+		var secondSecret = 'secondSecret';
+		var outTransferTransaction = createOutTransfer(dappId, transactionId, recipientId, amount, secret, secondSecret);
```
Isabello: Can this be abstracted into its own file? Will be easier to modify later if/when things change if these variables are declared in a single place.

willclarktech: These tests will be restructured as part of https://github.com/LiskHQ/lisk-js/issues/178 - are you happy if we do that then rather than now? (I agree it would be nice.)

PR #226, opened 2017-07-31
Adds fee to additional data field - Closes #146

### LINKED ISSUE #146: Adjust fees for data transaction field
Currently the fees for an optional data field are not included.

Fees should be 0.1 LSK or 10000000 liskies for an additional data field to max 64 bytes.

### LINKED ISSUE #222: Optional data field not null but not present
*placeholder*

Tosch110:
Closes #146


test/transactions/transaction.js
```diff
@@ -229,7 +229,7 @@ describe('transaction.js', () => {
 
 		describe('returned transaction', () => {
```
willclarktech: This test block is a duplicate.

src/transactions/transaction.js
```diff
@@ -43,9 +44,12 @@ function createTransaction(recipientId, amount, secret, secondSecret, data, time
 		senderPublicKey: keys.publicKey,
 		timestamp: slots.getTimeWithOffset(timeOffset),
 		asset: {},
-		data,
 	};
 
+	if (data && data.length > 0) {
```
willclarktech: Should there be any additional validation on `data`?

Tosch110: I first put another check, but then realized that any error message should be coming from core.

willclarktech: What about types? Data here might be stringified when sent to the node, which could lead to unexpected behaviour where the node accepts something the user didn't intend.

Even if validation is happening at the node, do we want to block malformed requests client-side for convenience?

Tosch110: As discussed in private, we will now add a check for utf8 string (else throw an error), what gets expected from core. We might think about more universal validation schemes in the future.

src/transactions/transaction.js
```diff
@@ -33,19 +34,26 @@ import { prepareTransaction } from './utils';
...
 		recipientId,
 		senderPublicKey: keys.publicKey,
 		timestamp: slots.getTimeWithOffset(timeOffset),
 		asset: {},
-		data,
 	};
 
+	if (data && data.length > 0) {
+		const encoding = jschardet.detect(data).encoding;
+		if (encoding !== 'ascii' && encoding !== 'UTF-8') throw new Error(`invalid encoding ${encoding} in transaction data`);
```
willclarktech: I would rather do this without adding the npm package. If we have a check like this:
```js
if (encoding === Buffer.from(encoding).toString()) {
  // ...
}
```
we avoid adding a dependency, and anyway we don't care what the best match for encoding is, we just care if it's `utf-8`-compatible.

willclarktech: (`'utf8'` is the default encoding, but we could specify it to make it explicit).

test/transactions/transaction.js
```diff
@@ -192,6 +192,23 @@ describe('transaction.js', () => {
 		});
 	});
 
+	describe('#createTransaction with data', () => {
+		const createTransaction = transaction.createTransaction;
+		let trs = null;
+
+		it('should create transaction with data', () => {
+			trs = createTransaction('58191285901858109L', 1000, 'secret', '', 'data');
```
willclarktech: Let's store this address, amount and secret in variables since we reuse them, and so it's obvious what they are from the variable names.

test/transactions/transaction.js
```diff
@@ -192,6 +192,23 @@ describe('transaction.js', () => {
...
+			(trs).should.be.ok();
+			(trs.fee).should.be.equal(20000000);
```
willclarktech: Likewise fee amount.

test/transactions/transaction.js
```diff
@@ -192,63 +196,67 @@ describe('transaction.js', () => {
 		});
 	});
 
-	describe('#createTransaction with data', () => {
+	describe('#createTransaction with data @now', () => {
```
willclarktech: `@now`

test/transactions/transaction.js
```diff
@@ -119,7 +123,7 @@ describe('transaction.js', () => {
 		});
 
 		it('should create transaction without second signature', () => {
-			trs = createTransaction('58191285901858109L', 1000, 'secret', secondSecret);
+			trs = createTransaction(testRecipientAddress, 1000, testSecret, secondSecret);
```
willclarktech: amount could be abstracted here too.

test/transactions/transaction.js
```diff
@@ -192,63 +196,67 @@ describe('transaction.js', () => {
...
 			});
 		});
 	});
 
 	describe('#createTransaction with secondSignature and data', () => {
 		const createTransaction = transaction.createTransaction;
 		let trs = null;
-		const secondSecret = 'second secret';
+		const feeWithData = 20000000;
+		const testAmount = 1000;
```
willclarktech: If we're using the same default amount in different places, maybe it should go in an outer `describe` block.

test/transactions/transaction.js
```diff
@@ -3,6 +3,11 @@ import transaction from '../../src/transactions/transaction';
 import cryptoModule from '../../src/transactions/crypto';
 
 describe('transaction.js', () => {
+	const testRecipientAddress = '58191285901858109L';
+	const testSecret = 'secret';
+	const testData = 'data';
+	const secondSecret = 'second secret';
```
willclarktech: Let's make the format of these variable names consistent.

src/transactions/transaction.js
```diff
@@ -33,19 +33,25 @@ import { prepareTransaction } from './utils';
...
+		fee,
 		recipientId,
 		senderPublicKey: keys.publicKey,
 		timestamp: slots.getTimeWithOffset(timeOffset),
 		asset: {},
-		data,
 	};
 
+	if (data && data.length > 0) {
+		if (data !== data.toString('utf8')) throw new Error('invalid encoding in transaction data');
```
willclarktech: `'Invalid encoding in transaction data.'`

test/transactions/transaction.js
```diff
@@ -192,44 +196,67 @@ describe('transaction.js', () => {
 		});
 	});
 
+	describe('#createTransaction with data @now', () => {
```
willclarktech: `@now`

test/transactions/transaction.js
```diff
@@ -192,44 +196,67 @@ describe('transaction.js', () => {
...
 		describe('returned transaction', () => {
 			it('should contain data field with string value', () => {
-				(trs.data).should.be.type('string');
+				(trs.asset.data).should.be.type('string');
+				(trs.fee).should.be.equal(feeWithData);
 			});
 		});
 	});
 
 	describe('#createTransaction with secondSignature and data', () => {
```
willclarktech: This test block is a duplicate.

src/transactions/transaction.js
```diff
@@ -33,19 +33,25 @@ import { prepareTransaction } from './utils';
...
+		fee,
 		recipientId,
 		senderPublicKey: keys.publicKey,
 		timestamp: slots.getTimeWithOffset(timeOffset),
 		asset: {},
-		data,
 	};
 
+	if (data && data.length > 0) {
+		if (data !== data.toString('utf8')) throw new Error('Invalid encoding in transaction data.');
```
willclarktech: `'Data must be utf-8 encoded.'`

src/transactions/transaction.js
```diff
@@ -33,19 +33,25 @@ import { prepareTransaction } from './utils';
  * @return {Object}
  */
 
+
```
willclarktech: Remove spare line.

test/transactions/transaction.js
```diff
@@ -37,15 +43,20 @@ describe('transaction.js', () => {
...
 			it('should use time slots with an offset of -10 seconds to get the time for the timestamp', () => {
 				const offset = -10;
 
-				trs = createTransaction('58191285901858109L', 1000, 'secret', null, null, offset);
+				trs = createTransaction(testRecipientAddress,
+					testAmountThousand,
+					testSecret,
+					null,
+					null,
+					offset);
```
willclarktech: I prefer the uniformity of
```js
trs = createTransaction(
  testRecipientAddress,
  ...
  offset,
);
```
and we've used that format already elsewhere.

test/transactions/transaction.js
```diff
@@ -119,7 +129,10 @@ describe('transaction.js', () => {
 		});
 
 		it('should create transaction without second signature', () => {
-			trs = createTransaction('58191285901858109L', 1000, 'secret', secondSecret);
+			trs = createTransaction(testRecipientAddress,
+				testAmountThousand,
+				testSecret,
+				testSecondSecret);
```
willclarktech: As above.

test/transactions/transaction.js
```diff
@@ -192,44 +205,48 @@ describe('transaction.js', () => {
...
-	describe('#createTransaction with secondSignature and data', () => {
+	describe('#createTransaction with data', () => {
 		const createTransaction = transaction.createTransaction;
 		let trs = null;
-		const secondSecret = 'second secret';
-
-		it('should be a function', () => {
-			(createTransaction).should.be.type('function');
-		});
+		const feeWithData = 20000000;
```
willclarktech: This is so many zeroes. Can we put it in terms of `fixedPoint`?

test/transactions/transaction.js
```diff
@@ -192,44 +205,48 @@ describe('transaction.js', () => {
...
+		const testAmount = 1000;
```
willclarktech: Should be able to reuse `testAmountThousand` from the outer block.

test/transactions/transaction.js
willclarktech: Also since the same value is reused later let's move this into the outer block.

test/transactions/transaction.js
```diff
@@ -192,44 +205,48 @@ describe('transaction.js', () => {
...
 			(createTransaction).should.be.type('function');
 		});
 
 		it('should create transaction with second signature and data', () => {
-			trs = createTransaction('58191285901858109L', 1000, 'secret', secondSecret, 'data');
+			trs = createTransaction(testRecipientAddress,
+				testAmountThousand,
+				testSecret,
+				testSecondSecret,
+				testData);
```
willclarktech: As above.

test/transactions/transaction.js
```diff
@@ -108,7 +119,6 @@ describe('transaction.js', () => {
 	describe('#createTransaction with second secret', () => {
 		const createTransaction = transaction.createTransaction;
 		let trs = null;
-		const secondSecret = 'second secret';
 		const keys = {
```
willclarktech: Maybe `keys` should be defined in the outer block to keep things together.

PR #257, opened 2017-08-08
Remove Object.entries/values - Closes #255

### LINKED ISSUE #255: Use babel-polyfill correctly
Should be in source code, i.e. dependencies not dev dependencies.

willclarktech:
Closes #255


Tosch110:
We have found another solution with babel

PR #294, opened 2017-08-16
Renames list to get functions - Closes #244

### LINKED ISSUE #244: Rename functions, get functions start with get...
Part of #139

Tosch110:
Closes #244

### DIFF
--- a/src/api/liskApi.js
+++ b/src/api/liskApi.js
@@ -281,7 +281,7 @@ LiskAPI.prototype.getStandbyDelegates = privateApi.wrapSendRequest(GET, 'delegat
  * @return API object
  */
 
-LiskAPI.prototype.searchDelegateByUsername = privateApi.wrapSendRequest(GET, 'delegates/search', username => ({ username }));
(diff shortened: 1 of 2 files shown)


src/api/liskApi.js
```diff
@@ -284,26 +284,26 @@ LiskAPI.prototype.listStandbyDelegates = privateApi.wrapSendRequest(GET, 'delega
 LiskAPI.prototype.searchDelegateByUsername = privateApi.wrapSendRequest(GET, 'delegates/search', username => ({ username }));
```
willclarktech: `getDelegateByUsername`?

Tosch110: I prefer to keep it `searchDelegateByUsername`.

Because it does not give a single response, even when a delegate name is exactly matched, it will display others that also contain the same strings in their delegate name.

willclarktech: OK, then plural makes more sense? `searchDelegatesByUsername`?

PR #307, opened 2017-08-17
Add tests for transactions utils

### LINKED ISSUE #280: [178] Make transaction utils tests atomic
Part of #178

willclarktech:
Closes #280

### DIFF
--- a/test/transactions/utils.js
+++ b/test/transactions/utils.js
@@ -1,0 +1,83 @@
+import utils from '../../src/transactions/utils';
+
+describe('transactions utils module', () => {
+	describe('exports', () => {
+		it('should be an object', () => {
+			(utils).should.be.type('object');
+		});
+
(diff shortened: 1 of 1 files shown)


test/transactions/utils.js
```diff
@@ -0,0 +1,83 @@
...
+		beforeEach(() => {
+			inputTransaction = Object.assign({}, defaultTransaction);
+		});
+
+		describe('without second signature', () => {
+			beforeEach(() => {
+				preparedTransaction = prepareTransaction(inputTransaction, keys);
+			});
+
+			it('should not mutate the original transaction', () => {
```
Tosch110: For me it would make sense to have this before the describe block.
Currently this test is double in both describe blocks.

willclarktech: Both situations - with or without a second secret - give rise to the opportunity to mutate the original transaction.

PR #342, opened 2017-08-31
Encrypt Passphrase - Closes #325

### LINKED ISSUE #325: AES256-cbc encryption / decryption
For the forging passphrase

Tosch110:
Closes #325


src/crypto/sign.js
```diff
@@ -301,3 +302,66 @@ export function verifyTransaction(transaction, secondPublicKey) {
...
+ * @return {string}
+ */
+
+function encryptAES256CBCWithPassword(plaintext, password) {
+	const iv = crypto.randomBytes(16);
+	const passHash = getSha256Hash(password, 'utf8');
+	const cipherInit = crypto.createCipheriv('aes-256-cbc', passHash, iv);
+	const textCipher = cipherInit.update(plaintext, 'utf8');
+	const finishedCipher = Buffer.concat([textCipher, cipherInit.final()]);
+	const cipherText = ''.concat(iv.toString('hex'), '$', finishedCipher.toString('hex'));
```
reyraa: Prefer using string templates

Tosch110: I would agree, when the delimiter in between would not be a dollar sign, making it not looking obvious anymore that there _is_ a delimiter:

`${iv.toString('hex')}$${finishedCipher.toString('hex')}`

willclarktech: I think it's still ok.

src/crypto/sign.js
```diff
@@ -301,3 +302,66 @@ export function verifyTransaction(transaction, secondPublicKey) {
...
+ * @method encryptAES256CBCWithPassword
+ * @param plaintext utf8
+ * @param password
+ *
+ * @return {string}
+ */
+
+function encryptAES256CBCWithPassword(plaintext, password) {
+	const iv = crypto.randomBytes(16);
+	const passHash = getSha256Hash(password, 'utf8');
```
willclarktech: Prefer `passwordHash`

src/crypto/sign.js
```diff
@@ -301,3 +302,66 @@ export function verifyTransaction(transaction, secondPublicKey) {
...
+	const cipherInit = crypto.createCipheriv('aes-256-cbc', passHash, iv);
```
willclarktech: Prefer `cipher`

src/crypto/sign.js
```diff
@@ -301,3 +302,66 @@ export function verifyTransaction(transaction, secondPublicKey) {
...
+	const textCipher = cipherInit.update(plaintext, 'utf8');
```
willclarktech: Prefer something like `initialOutput` or `firstBlock` or `output1`...

src/crypto/sign.js
willclarktech: Misleading name: the cipher text is the output of the cipher. This is the concatenation of IV and cipher text.

src/crypto/sign.js
```diff
@@ -301,3 +302,66 @@ export function verifyTransaction(transaction, secondPublicKey) {
...
+	const finishedCipher = Buffer.concat([textCipher, cipherInit.final()]);
```
willclarktech: Prefer `cipherOutput` or `encrypted`

src/crypto/sign.js
willclarktech: It's weird to mix hex and `$` here. What are the other options?

src/crypto/sign.js
```diff
@@ -301,3 +302,66 @@ export function verifyTransaction(transaction, secondPublicKey) {
...
+	const cipherText = ''.concat(iv.toString('hex'), '$', finishedCipher.toString('hex'));
+
+	return cipherText;
+}
+
+/**
+ * @method decryptAES256CBCWithPassword
+ * @param cipherText
```
willclarktech: What format are we expecting for this?

src/crypto/sign.js
```diff
@@ -301,3 +302,66 @@ export function verifyTransaction(transaction, secondPublicKey) {
...
+
+function encryptAES256CBCWithPassword(plaintext, password) {
+	const iv = crypto.randomBytes(16);
+	const passHash = getSha256Hash(password, 'utf8');
+	const cipherInit = crypto.createCipheriv('aes-256-cbc', passHash, iv);
+	const textCipher = cipherInit.update(plaintext, 'utf8');
+	const finishedCipher = Buffer.concat([textCipher, cipherInit.final()]);
+	const cipherText = ''.concat(iv.toString('hex'), '$', finishedCipher.toString('hex'));
+
+	return cipherText;
```
willclarktech: Do we definitely want to return a string, rather than an object here in `lisk-js`?

src/crypto/sign.js
```diff
@@ -301,3 +302,66 @@ export function verifyTransaction(transaction, secondPublicKey) {
...
+ * @param password
+ *
+ * @return {string}
+ */
+
+function decryptAES256CBCWithPassword(cipherText, password) {
+	const passHash = getSha256Hash(password, 'utf8');
+	const encryptedParts = cipherText.split('$');
+	const iv = Buffer.from(encryptedParts.shift(), 'hex');
+	const encryptedPass = Buffer.from(encryptedParts.join('$'), 'hex');
```
willclarktech: If there's more than 1 member of `encryptedParts` left then something has gone wrong. Why not throw an error, or just use the first member? Otherwise, `Buffer.from` will effectively do one of these anyway but it will depend on where the `$` appears in the string.

src/crypto/sign.js
```diff
@@ -301,3 +302,66 @@ export function verifyTransaction(transaction, secondPublicKey) {
...
+	const decipherInit = crypto.createDecipheriv('aes-256-cbc', Buffer.from(passHash), iv);
+	const decryptedPass = decipherInit.update(encryptedPass);
```
willclarktech: As above. Also prefer `decryptedPassword`

src/crypto/sign.js
```diff
@@ -301,3 +302,66 @@ export function verifyTransaction(transaction, secondPublicKey) {
...
+ *
+ * @return {string}
+ */
+
+function decryptAES256CBCWithPassword(cipherText, password) {
+	const passHash = getSha256Hash(password, 'utf8');
+	const encryptedParts = cipherText.split('$');
+	const iv = Buffer.from(encryptedParts.shift(), 'hex');
+	const encryptedPass = Buffer.from(encryptedParts.join('$'), 'hex');
+	const decipherInit = crypto.createDecipheriv('aes-256-cbc', Buffer.from(passHash), iv);
```
willclarktech: As above.

src/crypto/sign.js
```diff
@@ -301,3 +302,66 @@ export function verifyTransaction(transaction, secondPublicKey) {
...
+/**
+ * @method encryptPassphraseWithPassword
+ * @param plaintext utf8
+ * @param password
+ *
+ * @return {string}
+ */
+
+export function encryptPassphraseWithPassword(plaintext, password) {
+	return encryptAES256CBCWithPassword(plaintext, password);
```
willclarktech: Maybe this function should trim the plaintext?

src/crypto/sign.js
```diff
@@ -301,3 +302,66 @@ export function verifyTransaction(transaction, secondPublicKey) {
...
+
+/**
+ * @method encryptPassphraseWithPassword
+ * @param plaintext utf8
+ * @param password
+ *
+ * @return {string}
+ */
+
+export function encryptPassphraseWithPassword(plaintext, password) {
```
willclarktech: Let's call the first parameter `passphrase` since this function is specific to this use case.

test/crypto/sign.js
```diff
@@ -260,24 +262,65 @@ ${defaultSignature}
...
+	describe('encrypting passphrase integration test', () => {
+		const secretPassphrase = 'minute omit local rare sword knee banner pair rib museum shadow juice';
+		const password = 'myTotal53cr3t%&';
+		const encryptString = encryptPassphraseWithPassword(secretPassphrase, password);
+
+		describe('#encryptPassphraseWithPassword', () => {
+			it('should encrypt a given secret with a password', () => {
+				(encryptString).should.be.ok();
+				(encryptString).should.be.type('string');
+				(encryptString).should.containEql('$');
```
willclarktech: This is not an integration test.

test/crypto/sign.js
```diff
@@ -260,24 +262,65 @@ ${defaultSignature}
...
-					id: '13987348420913138422',
-				};
-
-				const multiSignature = multiSignTransaction(multiSigtransaction, firstSecret);
-
-				(multiSignature).should.be.eql(expectedMultiSignature);
+		describe('#decryptPassphraseWithPassword', () => {
+			it('should decrypt a given cipher with a password', () => {
+				const decryptedString = decryptPassphraseWithPassword(encryptString, password);
+				(decryptedString).should.be.eql(secretPassphrase);
```
willclarktech: This test should be atomic. It should encrypt, decrypt, and assert equality of original and output all in one place.

test/crypto/sign.js
```diff
@@ -260,24 +262,65 @@ ${defaultSignature}
...
+		});
+	});
+
+	describe('#encryptPassphraseWithPassword', () => {
+		it('should encrypt a message', () => {
+			const plainText = 'Hello Lisk';
+			const password = '123';
+			const cipher = encryptPassphraseWithPassword(plainText, password);
+
+			(cipher).should.be.type('string').and.length(65);
```
willclarktech: This should test the output against a hard-coded string.

test/crypto/sign.js
```diff
@@ -260,24 +262,65 @@ ${defaultSignature}
...
+
+			const multiSignature = multiSignTransaction(multiSigtransaction, firstSecret);
+
+			(multiSignature).should.be.eql(expectedMultiSignature);
+		});
+	});
+
+	describe('#encryptPassphraseWithPassword', () => {
+		it('should encrypt a message', () => {
+			const plainText = 'Hello Lisk';
```
willclarktech: The passphrase should be realistically long in this case.

test/crypto/sign.js
```diff
@@ -260,24 +262,65 @@ ${defaultSignature}
...
+			const cipher = encryptPassphraseWithPassword(plainText, password);
+
+			(cipher).should.be.type('string').and.length(65);
+		});
+	});
+
+	describe('#decryptPassphraseWithPassword', () => {
+		it('should decrypt a message', () => {
+			const cipher = 'eca6ab05837efb15dcbd500b211d95b9$4a378aebc4969d74c1db0c55420bf909';
+			const password = '123';
```
willclarktech: Since this is reused it can be hoisted to the outer block.

test/crypto/sign.js
```diff
@@ -42,6 +44,7 @@ describe('sign', () => {
 	const defaultSecondSecret = 'second secret';
 	const defaultPublicKey = '5d036a858ce89f844491762eb89e2bfbd50a4a0a0da658e4b2628b25b117ae09';
 	const defaultSecondPublicKey = '0401c8ac9f29ded9e1e4d5b6b43051cb25b22f27c7b7b35092161e851946f82f';
+	const firstSecret = '123';
```
willclarktech: Why do we need this as well as `defaultSecret`?

Tosch110: This is used in signTransaction and verify . Moved to the top no relevance for this PR

willclarktech: But why not use `defaultSecret` for all of these instead since we're refactoring anyway?

src/crypto/sign.js
```diff
@@ -301,3 +302,70 @@ export function verifyTransaction(transaction, secondPublicKey) {
...
+
+/**
+ * @method encryptAES256CBCWithPassword
+ * @param {String} passphrase utf8 - twelve word secret passphrase
+ * @param {String} password utf8 - the password used to encrypt the passphrase
+ *
+ * @return {string}
+ */
+
+function encryptAES256CBCWithPassword(passphrase, password) {
```
willclarktech: This function is generic so we should be using something like `plaintext` instead of `passphrase`.

src/crypto/sign.js
```diff
@@ -301,3 +302,70 @@ export function verifyTransaction(transaction, secondPublicKey) {
...
+
+	return {
+		cipher: encrypted.toString('hex'),
+		nonce: nonce.toString('hex'),
+	};
+}
+
+/**
+ * @method decryptAES256CBCWithPassword
+ * @param {Object} cipherAndNonce.
```
willclarktech: Remove `.`

src/crypto/sign.js
```diff
@@ -301,3 +302,70 @@ export function verifyTransaction(transaction, secondPublicKey) {
...
+/**
+ * @method encryptAES256CBCWithPassword
+ * @param {String} passphrase utf8 - twelve word secret passphrase
+ * @param {String} password utf8 - the password used to encrypt the passphrase
+ *
+ * @return {string}
+ */
+
+function encryptAES256CBCWithPassword(passphrase, password) {
+	const nonce = crypto.randomBytes(16);
```
willclarktech: We should be more accurate with our terms here. This is an initialisation vector, not a nonce.
https://crypto.stackexchange.com/a/16002

src/crypto/sign.js
```diff
@@ -301,3 +302,70 @@ export function verifyTransaction(transaction, secondPublicKey) {
...
+
+	return decrypted.toString();
+}
+
+/**
+ * @method encryptPassphraseWithPassword
+ * @param {String} passphrase utf8 - twelve word secret passphrase
+ * @param {String} password utf8 - the password used to encrypt the passphrase
+ *
+ * @return {String}
```
willclarktech: Output type

test/crypto/sign.js
```diff
@@ -260,24 +264,60 @@ ${defaultSignature}
...
+			const multiSignature = multiSignTransaction(multiSigtransaction, firstSecret);
+			(multiSignature).should.be.eql(expectedMultiSignature);
+		});
+	});
+
+	describe('#encryptPassphraseWithPassword', () => {
+		it('should encrypt a message', () => {
+			const cipher = encryptPassphraseWithPassword(secretPassphrase, defaultPassword);
+			(cipher).should.be.type('object').and.have.property('cipher');
+			(cipher).should.be.type('object').and.have.property('nonce');
```
willclarktech: Types/constraints for the properties?

test/crypto/sign.js
```diff
@@ -260,24 +264,60 @@ ${defaultSignature}
...
-
-				const multiSignature = multiSignTransaction(multiSigtransaction, firstSecret);
-
-				(multiSignature).should.be.eql(expectedMultiSignature);
+		describe('#decryptPassphraseWithPassword', () => {
+			it('should decrypt a given cipher with a password', () => {
+				const decryptedString = decryptPassphraseWithPassword(encryptString, defaultPassword);
+				(decryptedString).should.be.eql(secretPassphrase);
 			});
 		});
```
willclarktech: My previous comments about this integration test still apply.

src/crypto/sign.js
```diff
@@ -301,3 +302,70 @@ export function verifyTransaction(transaction, secondPublicKey) {
 
 	return secondSignaturePresent ? verifyTransaction(transactionWithoutSignature) : verified;
 }
+
+/**
+ * @method encryptAES256CBCWithPassword
+ * @param {String} passphrase utf8 - twelve word secret passphrase
+ * @param {String} password utf8 - the password used to encrypt the passphrase
+ *
+ * @return {string}
```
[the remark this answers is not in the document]
reyraa: In JSDocs notation all interfaces should be written in Capital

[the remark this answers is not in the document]
willclarktech: We're currently using Yuidoc in `lisk-js`.

src/crypto/sign.js
```diff
@@ -301,3 +302,70 @@ export function verifyTransaction(transaction, secondPublicKey) {
...
+ * @param {Object} cipherAndIv
+ * The cipher text resulting from the AES-256-CBC encryption,
+ * including the nonce { cipher: ..., nonce: ..., }
+ * @param {String} password utf8 - the password used to encrypt the passphrase
+ *
+ * @return {String}
+ */
+
+function decryptAES256CBCWithPassword(cipherAndIv, password) {
+	const { cipher, iv } = cipherAndIv;
```
willclarktech: Consider destructuring directly from the argument.

src/crypto/sign.js
```diff
@@ -301,3 +302,70 @@ export function verifyTransaction(transaction, secondPublicKey) {
...
+	const passwordHash = getSha256Hash(password, 'utf8');
+	const decipherInit = crypto.createDecipheriv('aes-256-cbc', passwordHash, hexToBuffer(iv));
+	const decryptedPassword = decipherInit.update(hexToBuffer(cipher));
```
willclarktech: As above, `firstBlock`.

src/crypto/sign.js
```diff
@@ -301,3 +302,70 @@ export function verifyTransaction(transaction, secondPublicKey) {
...
+}
+
+/**
+ * @method decryptAES256CBCWithPassword
+ * @param {Object} cipherAndIv
+ * The cipher text resulting from the AES-256-CBC encryption,
+ * including the nonce { cipher: ..., nonce: ..., }
+ * @param {String} password utf8 - the password used to encrypt the passphrase
+ *
+ * @return {String}
```
willclarktech: Maybe specify the encoding here too?

test/crypto/sign.js
```diff
@@ -260,25 +264,51 @@ ${defaultSignature}
...
+				iv: '89d0fa0b955219a0e6239339fbb8239f',
+			};
+			const decrypted = decryptPassphraseWithPassword(cipherAndNonce, defaultPassword);
+			(decrypted).should.be.eql(secretPassphrase);
+		});
+	});
+
+	describe('encrypting passphrase integration test', () => {
+		it('should encrypt a given secret with a password and decrypt it back to the original passphrase', () => {
+			const encryptString = encryptPassphraseWithPassword(secretPassphrase, defaultPassword);
```
willclarktech: `encryptedString`

webmaster128:
Does this still run in the Browser? `crypto` is a Node module, isn't it?

willclarktech:
@webmaster128 The module has been renamed `cryptography` since this PR.

willclarktech:
@webmaster128 We have a build process for the browser distribution...

webmaster128:
Thanks. I now found that this is a built-in feature by browserify

>  Get browser versions of the node core libraries events, stream, path, url, assert, buffer, util, querystring, http, vm, and crypto when you require() them

which implements the node crypto module using https://github.com/crypto-browserify/crypto-browserify.

PR #372, opened 2017-09-11
Return false instead of error in decrypt - Closes #341

### LINKED ISSUE #341: verifyMessage should return false on correct format
Verifying messages should throw error only on wrong input formats.

When everything is in the right format but the signature does not match the publicKey, we should return false for a non-verified message.

toschdev:
Closes #341


test/crypto/sign.js
```diff
@@ -162,7 +162,9 @@ ${defaultSignature}
 
 		it('should throw on invalid primary signature', () => {
 			const invalidTwoSignSignature = defaultTwoSignSignature.slice(0, 20);
-			(verifyMessageWithTwoPublicKeys.bind(null, invalidTwoSignSignature, publicKey1, publicKey2)).should.throw('Invalid signature second publicKey, cannot verify message');
+			(verifyMessageWithTwoPublicKeys(
+				invalidTwoSignSignature, publicKey1, publicKey2,
+			)).should.be.false();
 		});
 
 		it('should throw on invalid secondary signature', () => {
```
willclarktech: Test descriptions

src/crypto/sign.js
```diff
@@ -89,7 +89,7 @@ export function verifyMessageWithPublicKey(signedMessage, publicKey) {
 	if (signatureVerified) {
 		return naclInstance.decode_utf8(signatureVerified);
 	}
-	throw new Error('Invalid signature publicKey combination, cannot verify message');
+	return false;
```
willclarktech: Yuidoc return types. Do we want to always return a boolean instead of sometimes a boolean, sometimes a string?

willclarktech (changes requested):

Tosch110:
New solution in https://github.com/LiskHQ/lisk-js/pull/379

PR #374, opened 2017-09-12
move utils out of private api - closes #331

### LINKED ISSUE #331: Move more utils out of privateApi
Functions like `checkOptions` and `serialiseHTTPData` are pure and would make more sense in the utils file than in the already bloated `privateApi.js`.

Tschakki:
closes #331

### DIFF
--- a/src/api/liskApi.js
+++ b/src/api/liskApi.js
@@ -40,6 +40,7 @@
  * @main lisk
  */
 import privateApi from './privateApi';
+import utils from './utils';
 import config from '../../config.json';
 import cryptoModule from '../crypto';
 
@@ -208,7 +209,7 @@ LiskAPI.prototype.sendRequest = function sendRequest(
 ) {
 	const callback = callbackIfOptions || optionsOrCallback;
 	const options = (typeof optionsOrCallback !== 'function' && typeof optionsOrCallback !== 'undefined')
-		? privateApi.checkOptions.call(this, optionsOrCallback)
+		? utils.checkOptions.call(this, optionsOrCallback)
 		: {};
 
 	return privateApi.sendRequestPromise.call(this, requestMethod, requestType, options)
@@ -217,7 +218,7 @@ LiskAPI.prototype.sendRequest = function sendRequest(
 			this, requestMethod, requestType, options,
 		))
 		.catch(privateApi.handleSendRequestFailures.bind(this, requestMethod, requestType, options))
-		.then(privateApi.optionallyCallCallback.bind(this, callback));
+		.then(utils.optionallyCallCallback.bind(this, callback));
 };
 
 /**
@@ -229,7 +230,7 @@ LiskAPI.prototype.sendRequest = function sendRequest(
  * @return API object
  */
 
-LiskAPI.prototype.getAccount = privateApi.wrapSendRequest(GET, 'accounts', address => ({ address }));
+LiskAPI.prototype.getAccount = utils.wrapSendRequest(GET, 'accounts', address => ({ address }));
 
 /**
  * @method getActiveDelegates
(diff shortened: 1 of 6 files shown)


src/api/privateApi.js
```diff
@@ -299,12 +262,6 @@ function handleSendRequestFailures(requestMethod, requestType, options, error) {
...
 }
 
-function optionallyCallCallback(callback, result) {
-	if (typeof callback === 'function') {
-		callback(result);
-	}
-	return result;
-}
 
 module.exports = {
```
willclarktech: Functions should also be exported from the utils file instead of this one.

Tschakki: That's right, fixed that in the new commit, and updated tests

src/api/liskApi.js
```diff
@@ -40,6 +40,7 @@
  * @main lisk
  */
 import privateApi from './privateApi';
+import utils from './utils';
```
willclarktech: Since we're here anyway, it would be nice to destructure the functions out of `utils` directly here.

src/api/liskApi.js
```diff
@@ -208,7 +209,7 @@ LiskAPI.prototype.sendRequest = function sendRequest(
 ) {
 	const callback = callbackIfOptions || optionsOrCallback;
 	const options = (typeof optionsOrCallback !== 'function' && typeof optionsOrCallback !== 'undefined')
-		? privateApi.checkOptions.call(this, optionsOrCallback)
+		? utils.checkOptions.call(this, optionsOrCallback)
```
willclarktech: All these `utils` functions can be called directly instead of using `.call` or `.bind`.

src/api/privateApi.js
```diff
@@ -254,21 +211,16 @@ function sendRequestPromise(requestMethod, requestType, options) {
...
- * @param getDataFn
+ * @method handleTimestampIsInFutureFailures
+ * @param requestMethod
+ * @param requestType
+ * @param options
+ * @param result
+ * @private
  *
- * @return function wrappedSendRequest
+ * @return result object
```
willclarktech: This returns a `Promise`

src/api/privateApi.js
```diff
@@ -280,6 +232,17 @@ function handleTimestampIsInFutureFailures(requestMethod, requestType, options,
...
 
+/**
+ * @method handleSendRequestFailures
+ * @param requestMethod
+ * @param requestType
+ * @param options
+ * @param result
+ * @private
+ *
+ * @return error promise
```
willclarktech: This returns a `Promise`

src/api/utils.js
```diff
@@ -14,6 +14,67 @@
  */
 
 /**
+ * @method optionallyCallCallback
+ * @param calback
```
willclarktech: `callback`

test/api/utils.js
```diff
@@ -78,4 +127,211 @@ describe('api utils module', () => {
...
+
+		beforeEach(() => {
+			data = {
+				' key/1 ': '  value %2',
+				key3: 4,
+			};
+			serialisedData = serialiseHTTPData(data);
+		});
+
+		it('should prepend a question mark to the query string', () => {
```
willclarktech: `should trim, escape, and prepend a question mark to the query string`

test/api/utils.js
```diff
@@ -15,6 +15,55 @@
...
+				foo: 'bar',
+			},
+
+			parseOfflineRequests: () => ({
+				requestMethod: GET,
+			}),
+			setTestnet: () => {},
+			setNode: () => {},
+			sendRequest: () => {},
+		};
```
willclarktech: This `LSK` mock can be cut down a lot. Try to get rid of everything we don't need for testing utils.

src/api/privateApi.js
```diff
@@ -211,7 +181,7 @@ const serialiseHTTPData = (data) => {
 function createRequestObject(method, requestType, providedOptions) {
 	const options = providedOptions || {};
 	const url = method === GET
-		? `${getFullURL.call(this)}/api/${requestType}${serialiseHTTPData.call(this, options)}`
+		? `${getFullURL.call(this)}/api/${requestType}${utils.serialiseHTTPData.call(this, options)}`
```
willclarktech: Would be nice to destructure this directly from the import too.

src/api/utils.js
```diff
@@ -14,6 +14,67 @@
...
 
 /**
+ * @method optionallyCallCallback
+ * @param callback
+ * @param result
+ *
+ * @return result object
+ */
+
+function optionallyCallCallback(callback, result) {
```
willclarktech: Let's make this a `const` arrow function to match the others.

PR #379, opened 2017-09-15
Sign message functions return format - Closes #341

### LINKED ISSUE #341: verifyMessage should return false on correct format

Tosch110:
Closes #341


src/crypto/convert.js
```diff
@@ -36,7 +36,7 @@ export function bufferToHex(buffer) {
  */
 
 export function hexToBuffer(hex) {
-	return naclInstance.from_hex(hex);
+	return Buffer.from(naclInstance.from_hex(hex));
```
willclarktech: `Buffer.from(hex, 'hex')` - no need to use the `naclInstance` if we're only going to convert the result anyway.

src/crypto/sign.js
```diff
@@ -22,74 +22,81 @@ import {
...
+		signature: Buffer.from(signature).toString('base64'),
+		secondSignature: Buffer.from(secondSignature).toString('base64'),
+	};
 }
 
 /**
  * @method verifyMessageWithPublicKey
- * @param signedMessage
- * @param publicKey
+ * @param {Object} Object - Object with message (utf8), signature (base64) and publicKey (hex)
```
willclarktech: This description is made redundant by the subsequent lines.

src/crypto/sign.js
```diff
@@ -22,74 +22,81 @@ import {
...
 
 /**
  * @method signMessageWithSecret
- * @param message
- * @param secret
+ * @param message - utf8
+ * @param secret - utf8
  *
- * @return {string}
+ * @return {Object} - message, signature and publicKey
```
willclarktech: Prefer an order that's consistent with the code.

src/crypto/sign.js
```diff
@@ -22,74 +22,81 @@ import {
...
  * @method signMessageWithTwoSecrets
- * @param message
- * @param secret
- * @param secondSecret
+ * @param message - utf8
+ * @param secret - utf8
+ * @param secondSecret - utf8
  *
- * @return {string}
+ * @return {Object} - message, signature and publicKey
```
willclarktech: This object has additional properties.

src/crypto/sign.js
```diff
@@ -22,74 +22,81 @@ import {
...
  */
 
 export function signMessageWithTwoSecrets(message, secret, secondSecret) {
 	const msgBytes = naclInstance.encode_utf8(message);
```
willclarktech: In the verify function we use `Buffer.from` - probably best to be consistent.

src/crypto/sign.js
```diff
@@ -114,22 +133,15 @@ export function verifyMessageWithTwoPublicKeys(signedMessage, publicKey, secondP
...
-		secondSignatureVerified, publicKeyBytes,
+	const secondSignatureVerified = naclInstance.crypto_sign_verify_detached(
+		secondSignatureBytes, messageBytes, secondPublicKeyBytes,
 	);
 
-	if (!firstSignatureVerified) {
-		throw new Error('Invalid signature first publicKey, cannot verify message');
-	}
-	return naclInstance.decode_utf8(firstSignatureVerified);
+	return firstSignatureVerified && secondSignatureVerified;
```
willclarktech: Not sure how noticeable it would be, but more performant would be to have these as functions so that you don't have to wait for both verifications if the first one fails. I.e.:
```js
const verifyFirstSignature = () => naclInstance.crypto_sign_verify_detached(signatureBytes, messageBytes, publicKeyBytes);
const verifySecondSignature = () => naclInstance.crypto_sign_verify_detached(secondSignatureBytes, messageBytes, secondPublicKeyBytes);
return verifyFirstSignature() && verifySecondSignature();

File: src/crypto/sign.js [hunk 7]
```diff
@@ -141,7 +153,7 @@ export function verifyMessageWithTwoPublicKeys(signedMessage, publicKey, secondP
  * @return {string}
  */
 
-export function printSignedMessage(message, signedMessage, publicKey) {
+export function printSignedMessage({ message, signature, publicKey }) {
```
willclarktech: This function should also handle second-signed messages, right?

test/crypto/sign.js
```diff
@@ -50,28 +46,39 @@ describe('sign', () => {
 
 	describe('#signMessageWithSecret', () => {
 		const signedMessage = signMessageWithSecret(notSecretMessage, defaultSecret);
-
 		it('should signTransaction the message correctly', () => {
-			(signedMessage).should.be.equal(defaultSignature);
+			(signedMessage).should.have.property('message').be.equal(notSecretMessage);
+			(signedMessage).should.have.property('signature').be.equal(defaultSignature);
+			(signedMessage).should.have.property('publicKey').be.equal(defaultPublicKey);
```
willclarktech: Either put these in separate tests or assert `eql`ity with an object (to exclude extraneous properties).

test/crypto/sign.js
```diff
@@ -50,28 +46,39 @@ describe('sign', () => {
...
 		});
 	});
 
 	describe('#verifyMessageWithPublicKey', () => {
-		const signedMessage = signMessageWithSecret(notSecretMessage, defaultSecret);
-		const verifyMessage = verifyMessageWithPublicKey(signedMessage, defaultPublicKey);
+		let signature;
```
willclarktech: `signedMessage` or something that conveys that there's more than just the signature here.

test/crypto/sign.js
```diff
@@ -50,28 +46,39 @@ describe('sign', () => {
...
+		it('should throw on invalid signature length', () => {
+			signature.signature = `ERROR${defaultSignature}`;
+			(verifyMessageWithPublicKey.bind(null, signature)).should.throw('Invalid signature length, expected 64-byte signature');
 		});
 
-		it('should detect not verifiable signature', () => {
-			const invalidSignedMessage = `${signMessageWithSecret(notSecretMessage, defaultSecret)}ERROR`;
-			(verifyMessageWithPublicKey.bind(null, invalidSignedMessage, defaultPublicKey)).should.throw('Invalid signature publicKey combination, cannot verify message');
+		it('should return false on invalid signature', () => {
+			signature.signature = `W${defaultSignature.slice(1, -1)}S`;
```
willclarktech: This is invalid base64. Is this a test for an *invalid* signature (not valid base64 string) or a *valid but incorrect* signature (was not generated by the relevant private key for this message)?

test/crypto/sign.js
```diff
@@ -99,6 +110,68 @@ ${defaultSignature}
...
+		it('should signTransaction a message using two secrets', () => {
+			const signature = signMessageWithTwoSecrets(
+				notSecretMessage, defaultSecret, defaultSecondSecret,
+			);
+
+			(signature).should.have.property('message').be.equal(notSecretMessage);
+			(signature).should.have.property('publicKey').be.equal(defaultPublicKey);
+			(signature).should.have.property('secondPublicKey').be.equal(defaultSecondPublicKey);
+			(signature).should.have.property('signature').be.equal(defaultSignature);
+			(signature).should.have.property('secondSignature').be.equal(defaultSecondSignature);
```
willclarktech: As above

test/crypto/sign.js
```diff
@@ -99,6 +110,68 @@ ${defaultSignature}
...
+		});
+	});
+
+	describe('#verifyMessageWithTwoPublicKeys', () => {
+		let signature;
```
willclarktech: As above

test/crypto/sign.js
```diff
@@ -99,6 +110,68 @@ ${defaultSignature}
...
+			(verifyMessageWithTwoPublicKeys.bind(null, signature)).should.throw('Invalid first signature length, expected 64-byte signature');
+		});
+
+		it('should throw on invalid secondary signature length', () => {
+			signature.secondSignature = `ERR${defaultSecondSignature}`;
+			(verifyMessageWithTwoPublicKeys.bind(null, signature)).should.throw('Invalid second signature length, expected 64-byte signature');
+		});
+
+		it('should return false on wrong signature', () => {
+			signature.signature = `W${defaultSignature.slice(1, -1)}S`;
```
willclarktech: As above

test/crypto/sign.js
```diff
@@ -190,15 +215,7 @@ ${defaultSignature}
...
-			const alterTransaction = {
-				type: 0,
-				amount: '100',
-				recipientId: '58191285901858109L',
-				timestamp: 141738,
-				asset: {},
-				id: '13987348420913138422',
-				senderPublicKey: keys.publicKey,
-			};
+			const alterTransaction = Object.assign({}, transaction, { amount: '100' });
```
willclarktech: `alteredTransaction` - otherwise it sounds like a function name

test/crypto/sign.js
```diff
@@ -211,39 +228,31 @@ ${defaultSignature}
...
-				signSignature: 'e88b4bd56a80de3b15220bdf0d1df0aa024a7a127ef07b8dc36a4e12d50e8eb338bc61ebe510ab15839e23f073cffda2a8c8b3d1fc1f0db5eed114230ecffe0a',
-				id: '6950565552966532158',
-			};
-
+			const recipientId = '13356260975429434553L';
+			const type = 0;
+			const amount = '10';
+			const fee = '10000000';
+			const senderPublicKey = '215b667a32a5cd51a94c9c2046c11fffb08c65748febec099451e3b164452bca';
+			const senderSecondPublicKey = '922fbfdd596fa78269bbcadc67ec2a1cc15fc929a19c462169568d7a3df1a1aa';
```
willclarktech: Can't we reuse some of the variables in the outer block?

Tosch110: This is possible but not related to this PR and every signature for these tests would have to be redone.

src/crypto/sign.js
Tosch110: Not yet, this should be addressed for the future. I will create an issue for it

test/crypto/sign.js
```diff
@@ -190,17 +222,9 @@ ${defaultSignature}
...
-				recipientId: '58191285901858109L',
-				timestamp: 141738,
-				asset: {},
-				id: '13987348420913138422',
-				senderPublicKey: keys.publicKey,
-			};
+			const alteredTransaction = Object.assign({}, transaction, { amount: '100' });
 			const signature = signTransaction(transaction, defaultSignatureFirstSecret);
-			const alterSignature = signTransaction(alterTransaction, defaultSignatureFirstSecret);
+			const alterSignature = signTransaction(alteredTransaction, defaultSignatureFirstSecret);
```
willclarktech: `alteredTransactionSignature`

test/crypto/sign.js
```diff
@@ -211,39 +235,31 @@ ${defaultSignature}
...
-				fee: 10000000,
-				recipientId: '13356260975429434553L',
-				senderPublicKey: '215b667a32a5cd51a94c9c2046c11fffb08c65748febec099451e3b164452bca',
-				timestamp: 39541109,
-				asset: {},
-				signature: 'e7027dbe9bb8ebcc1738c560fe0a09161d781d9bfc5df4e9b4ccba2d7a1febcd25ba663938c8d22d4902d37435be149cfb0fd69e7a59daf53469abe8f6509e0c',
-				id: '6950565552966532158',
+				type, amount, fee, recipientId, senderPublicKey, timestamp, asset, signature, id,
 			};
+			const transactionForVerifyTwoSignatures = Object.assign(
```
willclarktech: Prefer reinitialising this in a `beforeEach` hook.

Issue #296, opened 2017-08-16
Remove publicKey from createDelegate method


diego-G:
The method `createDelegate` is returning an object with an asset which has a publickKey property. This property is not needed. Below, the related line from Core:
https://github.com/LiskHQ/lisk/blob/28201773982425bfde2126bd471846a58fb05364/logic/delegate.js#L68

PR #383, opened 2017-09-20
Refactor transaction bytes - Closes #366

### LINKED ISSUE #366: Refactor transactionBytes
This whole module can need a revisit and a refactor.

Tosch110:
Closes #366


test/transactions/transactionBytes.js
```diff
@@ -15,50 +15,250 @@
...
+				recipientId: defaultRecipient,
+				timestamp: defaultTimestamp,
+				asset: {},
+				senderPublicKey: defaultSenderPublicKey,
+				signature: defaultSignature,
+				id: defaultTransactionId,
+			};
+		});
+
+		it('should return Buffer of type 0 (send LSk) transaction', () => {
```
willclarktech: A few `LSk`s in here.

test/transactions/transactionBytes.js
```diff
@@ -15,50 +15,250 @@
...
+		it('should return Buffer of type 0 (send LSk) with data', () => {
+			defaultTransaction.asset.data = 'Hello Lisk! Some data in here!...';
+			const expectedBuffer = Buffer.from('AKopAgBdA2qFjOifhESRdi64niv71QpKCg2mWOSyYoslsReuCQDOvKqNNBU96AMAAAAAAABIZWxsbyBMaXNrISBTb21lIGRhdGEgaW4gaGVyZSEuLi5hilSXUhLq2T34yIFlXGJVRLzo7XzN/m8IpC7s+xrevQUTB75QFLsFFhe694FdUPYhKecJGBkDYeXU3UeWVBsK', 'base64');
+			const transactionBytes = getTransactionBytes(defaultTransaction);
+
+			(transactionBytes).should.be.eql(expectedBuffer);
+		});
+
+		it('should throw on type 0 with too much data', () => {
+			defaultTransaction.asset.data = '0123456789012345678901234567890123456789012345678901234567890123456789';
```
willclarktech: Prefer `new Array(maxDataLength + 1).fill('1').join('')` so we can see more easily see how long it is.

test/transactions/transactionBytes.js
```diff
@@ -15,50 +15,250 @@
...
+			(getTransactionBytes.bind(null, defaultTransaction)).should.throw('Transaction asset data exceeds size of 64.');
+		});
+
+		it('should return Buffer of transaction with second signature', () => {
+			defaultTransaction.signSignature = defaultSecondSignature;
+			const expectedBuffer = Buffer.from('AKopAgBdA2qFjOifhESRdi64niv71QpKCg2mWOSyYoslsReuCQDOvKqNNBU96AMAAAAAAABhilSXUhLq2T34yIFlXGJVRLzo7XzN/m8IpC7s+xrevQUTB75QFLsFFhe694FdUPYhKecJGBkDYeXU3UeWVBsKsAxK0ZiLyiRddENWYKJ4v+a/L176i9qW2Sf6v4tPb8/cspU/arrKoRnWiAmHpV3qDmNUvINmBStF+iMUVSICDw==', 'base64');
+			const transactionBytes = getTransactionBytes(defaultTransaction);
+
+			(transactionBytes).should.be.eql(expectedBuffer);
+		});
```
willclarktech: What about
1. Missing properties
1. Extra properties
1. Invalid values (e.g. non-hex strings, non-integers etc., asset not an object etc.)

test/transactions/transactionBytes.js
```diff
@@ -15,50 +15,250 @@
...
+			fee: 500000000,
+			recipientId: null,
+			senderPublicKey: defaultSenderPublicKey,
+			timestamp: defaultTimestamp,
+			asset: { signature: { publicKey: defaultSenderSecondPublicKey } },
+			signature: defaultSignature,
+			id: defaultTransactionId,
+		};
+
+		it('should create correct transactionBytes from signature transaction', () => {
```
willclarktech: `should return Buffer of type 1 (register second signature) transaction` would be more consistent with previous test

test/transactions/transactionBytes.js
```diff
@@ -15,50 +15,250 @@
...
-	it('should be ok', () => {
-		(getTransactionBytes).should.be.ok();
+	describe('send transaction, type 0', () => {
+		let defaultTransaction;
+
+		beforeEach(() => {
+			defaultTransaction = {
+				type: 0,
+				amount: defaultAmount,
+				recipientId: defaultRecipient,
```
willclarktech: `fee`?

willclarktech: Also https://docs.lisk.io/docs/the-lisk-protocol-transactions lists `senderId`, but not sure if that's still valid.

test/transactions/transactionBytes.js
```diff
@@ -15,50 +15,250 @@
...
+		});
 	});
 
-	it('should be a function', () => {
-		(getTransactionBytes).should.be.type('function');
+	describe('signature transaction, type 1', () => {
+		const signatureTransaction = {
+			type: 1,
+			amount: defaultNoAmount,
+			fee: 500000000,
```
willclarktech: Prefer specifying large numbers relative to a `fixedPoint` constant.

test/transactions/transactionBytes.js
```diff
@@ -15,50 +15,250 @@
...
-			id: '13987348420913138422',
+	describe('delegate registration transaction, type 2', () => {
+		const delegateRegistrationTransaction = {
+			type: 2,
+			amount: defaultNoAmount,
+			fee: 2500000000,
+			recipientId: null,
+			senderPublicKey: defaultSenderPublicKey,
+			timestamp: defaultTimestamp,
+			asset: { delegate: { username: 'MyDelegateUsername' } },
```
willclarktech: https://docs.lisk.io/docs/the-lisk-protocol-transactions lists `publicKey` as well as `username` for `asset.delegate`

test/transactions/transactionBytes.js
```diff
@@ -15,50 +15,250 @@
...
+			(transactionBytes).should.be.eql(expectedBuffer);
+		});
+	});
+
+	describe('vote transaction, type 3', () => {
+		const voteTransaction = {
+			type: 3,
+			amount: 0,
+			fee: 100000000,
+			recipientId: '18160565574430594874L',
```
willclarktech: Why not `defaultRecipient`?

test/transactions/transactionBytes.js
```diff
@@ -15,50 +15,250 @@
...
+		};
+
+		it('should create correct transactionBytes from create multisignature transaction', () => {
+			const expectedBuffer = Buffer.from('BKopAgBdA2qFjOifhESRdi64niv71QpKCg2mWOSyYoslsReuCQAAAAAAAAAAAAAAAAAAAAACBSs1ZDAzNmE4NThjZTg5Zjg0NDQ5MTc2MmViODllMmJmYmQ1MGE0YTBhMGRhNjU4ZTRiMjYyOGIyNWIxMTdhZTA5KzA0MDFjOGFjOWYyOWRlZDllMWU0ZDViNmI0MzA1MWNiMjViMjJmMjdjN2I3YjM1MDkyMTYxZTg1MTk0NmY4MmZhilSXUhLq2T34yIFlXGJVRLzo7XzN/m8IpC7s+xrevQUTB75QFLsFFhe694FdUPYhKecJGBkDYeXU3UeWVBsK', 'base64');
+			const transactionBytes = getTransactionBytes(createMultiSignatureTransaction);
+
+			(transactionBytes).should.be.eql(expectedBuffer);
+		});
+
+		it('should create correct transactionBytes from multisignature send transaction', () => {
```
willclarktech: This belongs in the `type 0` section

src/transactions/transactionBytes.js
```diff
@@ -203,132 +148,110 @@ function getEmptyBytesForTransaction(transaction) {
...
-			buffer[i] = arrayBuffer[i];
-		}
+	/**
+	 * @method concatTransactionBytes
+	 * @private
+	 * @return {Buffer}
+	 */
 
-		return Buffer.from(buffer);
+	concatTransactionBytes() {
```
willclarktech: Does this need to be a separate method or can we put this logic directly in `get transactionBytes`?

src/transactions/transactionBytes.js
```diff
@@ -203,132 +148,110 @@ function getEmptyBytesForTransaction(transaction) {
...
 
-function createTransactionBuffer(transaction) {
-	function assignHexToTransactionBytes(partTransactionBuffer, hexValue) {
-		const hexBuffer = Buffer.from(hexValue, 'hex');
-		for (let i = 0; i < hexBuffer.length; i++) {
-			partTransactionBuffer.writeByte(hexBuffer[i]);
-		}
-		return partTransactionBuffer;
-	}
+export class Transaction {
```
willclarktech: What are the benefits of using a class over a function here?

src/transactions/transactionBytes.js
```diff
@@ -203,132 +148,110 @@ function getEmptyBytesForTransaction(transaction) {
...
+			: Buffer.alloc(this.byteSizes.RECIPIENT_ID).fill(0);
 
-		assignHexToTransactionBytes(transactionBuffer, transaction.senderPublicKey);
+		this.transactionAmount = Buffer.alloc(this.byteSizes.AMOUNT);
+		this.transactionAmount.writeInt32LE(transaction.amount, 0, this.byteSizes.AMOUNT);
 
-		if (transaction.requesterPublicKey) {
-			assignHexToTransactionBytes(transactionBuffer, transaction.requesterPublicKey);
-		}
+		this.transactionAssetData = this.getAssetBytes(transaction);
```
willclarktech: `this.getAssetBytes` doesn't take any arguments.

src/transactions/transactionBytes.js
```diff
@@ -12,181 +12,126 @@
  * Removal or modification of this copyright notice is prohibited.
  *
  */
-/* eslint-disable no-plusplus */
-import ByteBuffer from 'bytebuffer';
 import bignum from 'browserify-bignum';
 
+const byteSizes = {
```
willclarktech: Maybe capitalise this too?

src/transactions/transactionBytes.js
```diff
@@ -203,132 +148,110 @@ function getEmptyBytesForTransaction(transaction) {
...
+	 * @throws
+	 * @return {}
+	 */
 
-function getTransactionBytes(transaction) {
-	return createTransactionBuffer(transaction);
+	checkTransaction() {
+		if (this.transaction.type === 0 && this.transaction.asset.data) {
+			if (this.transaction.asset.data.length > byteSizes.DATA
+				|| this.transactionAssetData.length > byteSizes.DATA) {
```
willclarktech: If `this.transactionAssetData.length` somehow ends up less than or equal to `byteSizes.DATA` does it matter if `this.transaction.asset.data.length > byteSizes.DATA`? If the transaction ends up with `this.transactionAssetData`, does it matter if it's not true that `this.transaction.type === 0 && this.transaction.asset.data`?

Alternatively, a check for `this.transaction.asset.data.length` could be performed at the start of the constructor, rather than after all the redundant processing.

src/transactions/transactionBytes.js
```diff
@@ -203,132 +148,110 @@ function getEmptyBytesForTransaction(transaction) {
...
-	);
+	/**
+	 * @method getAssetBytes
+	 * @private
+	 * @return {Buffer}
+	 */
 
-	return assignedTransactionBuffer;
-}
+	getAssetBytes() {
```
willclarktech: Does this need to be a method?

src/transactions/transactionBytes.js
```diff
@@ -203,132 +148,110 @@ function getEmptyBytesForTransaction(transaction) {
...
+				throw new Error(`Transaction asset data exceeds size of ${byteSizes.DATA}.`);
+			}
+		}
+	}
 }
 
-module.exports = {
-	getTransactionBytes,
-};
+export function getTransactionBytes(transaction) {
```
willclarktech: `export default`?

src/transactions/transactionBytes.js
```diff
@@ -203,132 +148,110 @@ function getEmptyBytesForTransaction(transaction) {
...
+	constructor(transaction) {
+		this.byteSizes = byteSizes;
+		this.transaction = transaction;
 
-	/**
-	 * @method createEmptyTransactionBuffer
-	 * @param assetSize number
-	 * @return {buffer}
-	 */
+		this.transactionType = Buffer.alloc(this.byteSizes.TYPE);
```
willclarktech: Prefer allocating when creating the Buffer: `Buffer.alloc(this.byteSizes.TYPE, transaction.type)`

src/transactions/transactionBytes.js
```diff
@@ -203,132 +148,110 @@ function getEmptyBytesForTransaction(transaction) {
...
-	function assignHexToTransactionBytes(partTransactionBuffer, hexValue) {
-		const hexBuffer = Buffer.from(hexValue, 'hex');
-		for (let i = 0; i < hexBuffer.length; i++) {
-			partTransactionBuffer.writeByte(hexBuffer[i]);
-		}
-		return partTransactionBuffer;
-	}
+export class Transaction {
+	constructor(transaction) {
+		this.byteSizes = byteSizes;
```
willclarktech: Does this need to be stored in the `Transaction`?

src/transactions/transactionBytes.js
```diff
@@ -12,181 +12,126 @@
...
 
-function getEmptyBytesForTransaction(transaction) {
+function getAssetBytesHelper(transaction) {
 	/**
 	 * @method isSendTransaction
-	 * @return {object}
+	 * @return {Buffer}
 	 */
 
 	function isSendTransaction() {
```
willclarktech: Can we extract these outside `getAssetBytesHelper` into self-contained, pure functions for readability?

src/transactions/transactionBytes.js
```diff
@@ -12,181 +12,126 @@
...
 	 */
 
 	function isVoteTransaction() {
-		const voteTransactionBytes = (Buffer.from(transaction.asset.votes.join('')) || null);
-
-		return {
-			assetBytes: voteTransactionBytes,
-			assetSize: (voteTransactionBytes.length || 0),
-		};
+		return Buffer.from(transaction.asset.votes.join(''));
```
willclarktech: I guess these are `utf8` because of the `+` but I like that you specify `utf8` in e.g. `isDelegateTransaction` (even though it's not strictly necessary).

src/transactions/transactionBytes.js
```diff
@@ -12,181 +12,126 @@
...
-
-		bb.toBuffer();
-		const multiSigBuffer = new Uint8Array(bb.toArrayBuffer());
-
-		return {
-			assetBytes: multiSigBuffer,
-			assetSize: multiSigBuffer.length,
-		};
+		const multisigTransactionAsset = transaction.asset.multisignature;
+		const minBuffer = Buffer.alloc(1);
```
willclarktech: Prefer allocating directly when creating the Buffer: `Buffer.alloc(1, multisigTransactionAsset.min)`

src/transactions/transactionBytes.js
```diff
@@ -12,181 +12,126 @@
...
+		minBuffer.writeInt8(multisigTransactionAsset.min);
+		const lifetimeBuffer = Buffer.alloc(1);
```
willclarktech: As above.

src/transactions/transactionBytes.js
```diff
@@ -12,181 +12,126 @@
...
+		lifetimeBuffer.writeInt8(multisigTransactionAsset.lifetime);
+		const keysgroupBuffer = Buffer.from(multisigTransactionAsset.keysgroup.join(''));
```
willclarktech: As above (`utf8` comment)

src/transactions/transactionBytes.js
```diff
@@ -12,181 +12,126 @@
...
-		bb.flip();
-
-		buf = Buffer.concat([buf, bb.toBuffer()]);
-
-		return {
-			assetBytes: buf,
-			assetSize: buf.length,
-		};
+		const dappNameBuffer = Buffer.from(dapp.name);
+		const dappDescriptionBuffer = dapp.description ? Buffer.from(dapp.description) : Buffer.from('');
```
willclarktech: Prefer `Buffer.alloc(0)`

src/transactions/transactionBytes.js
```diff
@@ -12,181 +12,126 @@
...
+		const dappIconBuffer = dapp.icon ? Buffer.from(dapp.icon) : Buffer.from('');
+		const dappTypeBuffer = Buffer.alloc(4);
+		dappTypeBuffer.writeInt8(dapp.type);
+		const dappCategoryBuffer = Buffer.alloc(4);
+		dappCategoryBuffer.writeInt8(dapp.category);
+
+		return Buffer.concat([
+			dappNameBuffer,
+			dappDescriptionBuffer,
+			dappTagsBuffer, dappLinkBuffer,
```
willclarktech: Prefer everything on its own line

src/transactions/transactionBytes.js
```diff
@@ -12,181 +12,126 @@
...
-	/**
-	 * `transactionType` describes the available transaction types.
-	 *
-	 * @property transactionType
-	 * @type object
-	 */
+		return Buffer.concat([dappOutAppIdBuffer, dappOutTransactionIdBuffer]);
+	}
 
 	const transactionType = {
```
willclarktech: This needs to be renamed, along with the functions for getting asset data for a given transaction type. Perhaps `assetDataGetters` and `getAssetDataForSendTransaction` etc?

test/transactions/transactionBytes.js
Tosch110: PublicKey will not be part of the delegate asset anymore:
https://github.com/LiskHQ/lisk-js/issues/296

src/transactions/transactionBytes.js
```diff
@@ -12,323 +12,252 @@
...
-		const bb = new ByteBuffer(32, true);
-		const publicKey = transaction.asset.signature.publicKey;
-		const publicKeyBuffer = Buffer.from(publicKey, 'hex');
-
-		for (let i = 0; i < publicKeyBuffer.length; i++) {
-			bb.writeByte(publicKeyBuffer[i]);
+export function checkRequiredFields(requiredFields, data) {
+	requiredFields.forEach((parameter) => {
+		const dataFields = Object.keys(data);
+		if (!dataFields.includes(parameter)) {
```
willclarktech: What if the key is there but the value is `undefined`?

src/transactions/transactionBytes.js
```diff
@@ -12,323 +12,252 @@
...
+	checkRequiredFields(['name', 'link', 'type', 'category'], dapp);
+	const { name, description, tags, link, icon, type, category } = dapp;
+	const dappNameBuffer = Buffer.from(name, 'utf8');
+	const dappLinkBuffer = Buffer.from(link, 'utf8');
+	const dappTypeBuffer = Buffer.alloc(4, type);
+	const dappCategoryBuffer = Buffer.alloc(4, category);
+
+	const dappDescriptionBuffer = description ? Buffer.from(description, 'utf8') : Buffer.alloc(0);
+	const dappTagsBuffer = tags ? Buffer.from(tags, 'utf8') : Buffer.alloc(0);
+	const dappIconBuffer = icon ? Buffer.from(icon, 'utf8') : Buffer.alloc(0);
```
willclarktech: Since we're in the context of a dapp transaction, better to remove the `dapp` prefix from all of these for readability (as in multisignature function above).

src/transactions/transactionBytes.js
```diff
@@ -12,323 +12,252 @@
...
 
-	/**
-	 * @method isDappOutTransferTransaction
-	 * @return {object}
-	 */
+export function getAssetDataForDappOutTransaction({ outTransfer }) {
+	checkRequiredFields(['dappId', 'transactionId'], outTransfer);
+	const { dappId, transactionId } = outTransfer;
+	const dappOutAppIdBuffer = Buffer.from(dappId, 'utf8');
+	const dappOutTransactionIdBuffer = Buffer.from(transactionId, 'utf8');
```
willclarktech: As above.

src/transactions/transactionBytes.js
```diff
@@ -12,323 +12,252 @@
...
-		0: isSendTransaction,
-		1: isSignatureTransaction,
-		2: isDelegateTransaction,
-		3: isVoteTransaction,
-		4: isMultisignatureTransaction,
-		5: isDappTransaction,
-		6: isDappInTransferTransaction,
-		7: isDappOutTransferTransaction,
+export function getAssetBytesHelper(transaction) {
+	const assetDataGetters = {
```
willclarktech: It looks odd to have an object whose keys are `0`–`7`, but I guess it makes more sense than an array in this case.

src/transactions/transactionBytes.js
```diff
@@ -12,323 +12,252 @@
...
-			assignHexToTransactionBytes(transactionBuffer, transaction.signSignature);
-		}
-
-		transactionBuffer.flip();
-		const arrayBuffer = new Uint8Array(transactionBuffer.toArrayBuffer());
-		const buffer = [];
-
-		for (let i = 0; i < arrayBuffer.length; i++) {
-			buffer[i] = arrayBuffer[i];
+export function checkTransaction(transaction) {
```
willclarktech: Return `true` to match `checkRequiredFields` pattern.

src/transactions/transactionBytes.js
```diff
@@ -12,323 +12,252 @@
...
+	if (transaction.asset.data) {
```
willclarktech: Destructure `data` out of `transaction` and you can do:
```js
if (data && data.length > BYTESIZES.DATA) {
    // ...
}
```

src/transactions/transactionBytes.js
```diff
@@ -12,323 +12,252 @@
...
+];
 
-function getTransactionBytes(transaction) {
-	return createTransactionBuffer(transaction);
+/**
+* A utility class to get transaction byteSizes
+*
+* @class TransactionBytes
+* @param {Object} transaction
+* @constructor
```
willclarktech: Needs updating

src/transactions/transactionBytes.js
```diff
@@ -12,323 +12,252 @@
...
-	return assignedTransactionBuffer;
 }
 
-/**
- * @method getBytes
- * @param transaction Object
- *
- * @return {buffer}
- */
+const REQUIRED_TRANSACTION_PARAMETER = [
```
willclarktech: `REQUIRED_TRANSACTION_PARAMETER`

src/transactions/transactionBytes.js
```diff
@@ -12,323 +12,252 @@
...
+		? Buffer.from(transaction.requesterPublicKey, 'hex')
+		: Buffer.alloc(0);
+
+	const transactionRecipientID = transaction.recipientId
+		? Buffer.from(
+			bignum(
+				transaction.recipientId.slice(0, -1),
+			).toBuffer({ size: BYTESIZES.RECIPIENT_ID }),
+		)
+		: Buffer.alloc(BYTESIZES.RECIPIENT_ID).fill(0);
```
willclarktech: Unnecessary `fill`

src/transactions/transactionBytes.js
```diff
@@ -12,323 +12,252 @@
...
-	const transactionType = {
-		0: isSendTransaction,
-		1: isSignatureTransaction,
-		2: isDelegateTransaction,
-		3: isVoteTransaction,
-		4: isMultisignatureTransaction,
-		5: isDappTransaction,
-		6: isDappInTransferTransaction,
-		7: isDappOutTransferTransaction,
+export function getAssetBytesHelper(transaction) {
```
willclarktech: I think it makes more sense to call this `getAssetBytes`

src/transactions/transactionBytes.js
```diff
@@ -12,323 +12,252 @@
...
+* A utility class to get transaction byteSizes
+*
+* @class TransactionBytes
+* @param {Object} transaction
+* @constructor
+*/
+
+export function getTransactionBytes(transaction) {
+	checkTransaction(transaction);
+	checkRequiredFields(REQUIRED_TRANSACTION_PARAMETER, transaction);
```
willclarktech: Maybe this should be called inside `checkTransaction` since that sounds like it's going to check everything.

src/transactions/transactionBytes.js
```diff
@@ -12,323 +12,252 @@
...
-	return createTransactionBuffer(transaction);
+/**
+* A utility class to get transaction byteSizes
+*
+* @class TransactionBytes
+* @param {Object} transaction
+* @constructor
+*/
+
+export function getTransactionBytes(transaction) {
```
willclarktech: `export default`?

src/transactions/transactionBytes.js
```diff
@@ -12,323 +12,252 @@
...
+	checkTransaction(transaction);
+	checkRequiredFields(REQUIRED_TRANSACTION_PARAMETER, transaction);
+
+	const transactionType = Buffer.alloc(BYTESIZES.TYPE);
+	transactionType.writeInt8(transaction.type);
```
willclarktech: Is there a reason to do this rather than on one line?
```js
const transactionType = Buffer.alloc(BYTESIZES.TYPE, transaction.type);
```

test/transactions/transactionBytes.js
```diff
@@ -12,53 +12,549 @@
...
+			const transactionBytes = getTransactionBytes(outTransferTransaction);
+
+			(transactionBytes).should.be.eql(expectedBuffer);
+		});
+	});
+});
+
+describe('getTransactionBytes helper functions', () => {
+	describe('#checkRequiredFields', () => {
+		const arrayToCheck = ['OneValue', 'SecondValue', 'ThirdValue'];
```
willclarktech: How about some non-string keys?

test/transactions/transactionBytes.js
```diff
@@ -12,53 +12,549 @@
...
+			const objectParameter = {
+				OneValue: '1',
+				SecondValue: '2',
+			};
+
+			(checkRequiredFields.bind(null, arrayToCheck, objectParameter)).should.throw('ThirdValue is a required parameter.');
+		});
+	});
+
+	const defaultEmptyBuffer = Buffer.alloc(0);
```
willclarktech: Prefer defining constants at the top of a `describe` block.

test/transactions/transactionBytes.js
```diff
@@ -12,53 +12,549 @@
...
+			});
+
+			(assetSignaturesPublicKeyBuffer).should.be.eql(expectedBuffer);
+		});
+
+		it('should throw on missing publicKey in the signature asset', () => {
+			(getAssetDataForSignatureTransaction.bind(null, { signature: {} })).should.throw(
+				'publicKey is a required parameter.',
+			);
+		});
```
willclarktech: As above: what if the key is present, but the value is `undefined`?

test/transactions/transactionBytes.js
```diff
@@ -12,53 +12,549 @@
...
+		const defaultType = 0;
+		const defaultLink = 'https://github.com/MaxKK/guestbookDapp/archive/master.zip';
+		const defaultIcon = 'https://raw.githubusercontent.com/MaxKK/guestbookDapp/master/icon.png';
+		const dappNameBuffer = Buffer.from(defaultDappName, 'utf8');
+		const dappDescriptionBuffer = Buffer.from(defaultDescription, 'utf8');
+		const dappTagsBuffer = Buffer.from(defaultTags, 'utf8');
+		const dappLinkBuffer = Buffer.from(defaultLink, 'utf8');
+		const dappIconBuffer = Buffer.from(defaultIcon, 'utf8');
+		const dappTypeBuffer = Buffer.alloc(4, defaultType);
+		const dappCategoryBuffer = Buffer.alloc(4, defaultCategory);
```
willclarktech: I think with these more complicated cases, it makes more sense to calculate the expected value and store it as a Buffer from hex string to avoid simply replicating the logic under test.

test/transactions/transactionBytes.js
```diff
@@ -12,53 +12,549 @@
...
+				asset: {
+					data: new Array(maxDataLength + 1).fill('1').join(''),
+				},
+				senderPublicKey: defaultSenderPublicKey,
+				senderId: defaultSenderId,
+				signature: defaultSignature,
+				id: defaultTransactionId,
+			};
+			(checkTransaction.bind(null, defaultTransaction)).should.throw('Transaction asset data exceeds size of 64.');
+		});
```
willclarktech: We need a test for if there isn't too much data.

test/transactions/transactionBytes.js
Tosch110: good catch!

test/transactions/transactionBytes.js
Tosch110: Addressed in utils - fn isValidValue

src/transactions/utils.js
```diff
@@ -35,6 +35,19 @@ const prepareTransaction = (transaction, secret, secondSecret) => {
 	return transactionWithId;
 };
 
+const isValidValue = (value) => {
+	if (
+		// eslint-disable-next-line no-self-compare
+		value !== value
```
willclarktech: Is this for `NaN`? Prefer explicit checks. Or even
```js
if ([undefined, false, NaN, /* ... */].includes(value)) {
    / ...
}
```

willclarktech: Actually, since this is just a bunch of boolean checks it should probably return a value directly rather than having a body at all.

src/transactions/transactionBytes.js
```diff
@@ -12,323 +12,250 @@
...
+	checkRequiredFields(['dappId'], inTransfer);
+	const { dappId } = inTransfer;
+	return Buffer.from(dappId, 'utf8');
+}
 
-		if (transaction.requesterPublicKey) {
-			assignHexToTransactionBytes(transactionBuffer, transaction.requesterPublicKey);
-		}
+/**
+ * @method getASsetDataForDappOutTransaction
```
willclarktech: Weird capitalisation

src/transactions/transactionBytes.js
```diff
@@ -12,323 +12,250 @@
...
 /**
- * @method getBytes
- * @param transaction Object
- *
- * @return {buffer}
- */
-
-function getTransactionBytes(transaction) {
-	return createTransactionBuffer(transaction);
+* A utility method to get transaction byteSizes
```
willclarktech: This doesn't just get byte sizes now.

test/transactions/transactionBytes.js
```diff
@@ -12,53 +12,586 @@
...
+			const objectParameter = {
+				0: 0,
+				1: 10,
+				2: 20,
+				3: 30,
+			};
+			(checkRequiredFields(arrayToCheckNonString, objectParameter)).should.be.true();
+		});
+
+		it('should work with non-string keys', () => {
```
willclarktech: Duplicate test

PR #394, opened 2017-10-03
Print signed message for with signing with two signatures - Closes #385

### LINKED ISSUE #385: Print signed message for with signing with two signatures
Currently the print function only supports one publicKey and one signature. 

The signMessageWithTwoPublicKeys should also have a print function or the old ones support this format.

alepop:
closes #385

### DIFF
--- a/src/crypto/sign.js
+++ b/src/crypto/sign.js
@@ -25,6 +25,14 @@ import {
 } from './keys';
 import { getTransactionHash, getSha256Hash } from './hash';
 
+const signedMessageHeader = '-----BEGIN LISK SIGNED MESSAGE-----';
+const messageHeader = '-----MESSAGE-----';
+const publicKeyHeader = '-----PUBLIC KEY-----';
+const secondPublicKeyHeader = '-----SECOND PUBLIC KEY-----';
+const signatureHeader = '-----SIGNATURE-----';
+const secondSignatureHeader = '-----SECOND SIGNATURE-----';
+const signatureFooter = '-----END LISK SIGNED MESSAGE-----';
+
 /**
  * @method signMessageWithSecret
  * @param message - utf8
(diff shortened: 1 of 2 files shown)


willclarktech:
Hey @alepop, thanks for the PR. Can you think of a use case for calling `printSignedMessage` rather than `printSignedMessageWithTwoSecrets` on a message with a second signature? I was imagining `printSignedMessage` would just check if `secondSignature` is present and conditionally include the relevant sections. Then we only need one function, and whatever's calling it doesn't need to know that messages with one vs two signatures get printed differently.

alepop:
@willclarktech when I started implementing this feature I thought the same. In my opinion than function is easier then better (avoid if statements as much as possible). But if you think that the better way is to implement it in one function then I do it.

willclarktech:
@alepop I think so. That way we can also put `signedMessage` and `result` in `const`s rather than having to use `let` in `signAndPrintMessage`. We don't need any `if` statements in `printSignedMessage`, we can just conditionally include elements in the `outputArray`:
```js
const outputArray = [
	signedMessageHeader,
	messageHeader,
	message,
	publicKeyHeader,
	publicKey,
	secondPublicKey ? secondPublicKeyHeader : null,
	secondPublicKey,
	signatureHeader,
	signature,
	secondSignature ? secondSignatureHeader : null,
	secondSignature,
	signatureFooter,
].filter(Boolean);
```

alepop:
@willclarktech agree. I'll fix it.

src/crypto/sign.js
```diff
@@ -148,27 +156,28 @@ export function verifyMessageWithTwoPublicKeys(
  * @param message
  * @param signedMessage
  * @param publicKey
+ * @param secondSignature
+ * @param secondPublicKey
```
willclarktech: Actually these Yuidoc annotations are incorrect - there's just one parameter.

test/crypto/sign.js
```diff
@@ -115,8 +115,22 @@ ${defaultSignature}
 		});
 	});
 
-	describe('#signMessageWithTwoSecrets', () => {
-		it('should signTransaction a message using two secrets', () => {
+	describe('signTransaction and print messages with two secrets', () => {
+		const signedMessageWithTwoSecretsExample = `
+-----BEGIN LISK SIGNED MESSAGE-----
+-----MESSAGE-----
+not secret message
```
willclarktech: Prefer this from the variable `notSecretMessage`

test/crypto/sign.js
```diff
@@ -115,8 +115,22 @@ ${defaultSignature}
...
+${defaultPublicKey}
+-----SECOND PUBLIC KEY-----
+${defaultSecondPublicKey}
+-----SIGNATURE-----
+${defaultSignature}
+-----SECOND SIGNATURE-----
+${defaultSecondSignature}
+-----END LISK SIGNED MESSAGE-----
+`.trim();
+		it('#signMessageWithTwoSecret should signTransaction using two secrets', () => {
```
willclarktech: `signMessageWithTwoSecrets`

PR #400, opened 2017-10-05
Use tweetnacl instead of js-nacl  - Closes #393

### LINKED ISSUE #393: React-Native bundler halts after include
I'm currently trying to create an app with `react-native`.

After I try to include lisk-js with `import lisk from 'lisk-js'` the bundler stops working while including lisk. It's a plain project and after removing the line it's working again!

Do you have any idea what's going on - I will continue trying to find a solution

Thanks!

tobiaslins:
Closes #393


Tosch110:
Thank you @tobiaslins for the Pull Request. We already had a chat on this topic, to sum up, even though we appreciate the effort, we will not include this PR for several reasons.
The crypto module has already diverged a lot since 0.5.0 and has been refactored, even though we are still using `js-nacl` in the 1.0.0 version.
But our plans are different than switching to another 'swiss army knife' of crypto libraries, we want to specifically focus on the functions we really need and therefore slim the application this way. Hopefully also make it more compatible with other libraries. 
Also, changing the overall `crypto` library will involve more in-depth discussions and probably benchmarks, security inspections etc.
Thanks for your understanding.

Tosch110:
Btw. if this approach works for you feel free to use a modified version in your application. This is the strength of open-source ;)

PR #430, opened 2017-10-19
Update transactions and tests for strings - Closes #409

### LINKED ISSUE #409: Update transactions and tests so amount and fee are a string
Reflecting upcoming change in lisk-core

Tosch110:
Closes #409

### DIFF
--- a/src/transactions/0_send.js
+++ b/src/transactions/0_send.js
@@ -46,8 +46,8 @@ export default function send({
 	const fee = data ? SEND_FEE + DATA_FEE : SEND_FEE;
 	const transaction = {
 		type: 0,
-		amount,
-		fee,
+		amount: amount.toString(),
+		fee: fee.toString(),
 		recipientId,
 		senderPublicKey: keys.publicKey,
 		timestamp: getTimeWithOffset(timeOffset),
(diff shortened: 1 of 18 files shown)


src/transactions/0_send.js
```diff
@@ -46,8 +46,8 @@ export default function send({
 	const fee = data ? SEND_FEE + DATA_FEE : SEND_FEE;
```
willclarktech: @Isabello Should we still be doing maths with fees as vanilla numbers? If not it would make more sense to have the fee constants themselves as strings.

test/transactions/3_castVotes.js
```diff
@@ -75,18 +75,18 @@ describe('#castVotes transaction', () => {
...
 				castVotesTransaction.should.have
 					.property('amount')
-					.and.be.type('number')
-					.and.equal(0);
+					.and.be.type('string')
+					.and.equal('0');
 			});
 
-			it('should have fee number equal to 0', () => {
+			it('should have fee string equal to 0', () => {
```
willclarktech: Test description doesn't match test.

src/transactions/0_send.js
Tosch110: I would prefer keeping internal (Lisk-JS) fee in numbers for the calculations. As soon as we put it in a transaction object we can make sure it is a string.

PR #439, opened 2017-10-26
:white_check_mark: Revisit second secret tests - Closes #407

### LINKED ISSUE #407: Revisit second secret tests for transactions
Some tests aren't doing what they say they are. E.g.
https://github.com/LiskHQ/lisk-js/blob/246033fb80f6ed865aa9b038f888f6264a025abc/test/transactions/0_sendFromMultisignatureAccount.js#L147-L160

Tosch110:
Closes #407

### DIFF
--- a/test/transactions/0_send.js
+++ b/test/transactions/0_send.js
@@ -43,7 +43,7 @@ describe('#send transaction', () => {
 			.returns(timeWithOffset);
 	});
 
-	describe('without second secret', () => {
+	describe('with one secret', () => {
 		describe('without data', () => {
 			beforeEach(() => {
 				sendTransaction = send({
@@ -145,6 +145,10 @@ describe('#send transaction', () => {
 						.and.be.type('object')
 						.and.be.empty();
 				});
+
+				it('should not have the second signature property', () => {
+					sendTransaction.should.not.have.property('signSignature');
+				});
 			});
 		});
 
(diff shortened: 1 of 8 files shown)


test/transactions/0_send.js
```diff
@@ -199,64 +203,56 @@ describe('#send transaction', () => {
...
-		describe('returned send transaction', () => {
-			it('should have second signature hex string', () => {
-				sendTransaction.should.have
-					.property('signSignature')
-					.and.be.hexString();
+		it('should have the second signature property as hex string', () => {
+			sendTransaction.should.have.property('signSignature').and.be.hexString();
+		});
+
+		describe('verification of the created transaction', () => {
```
willclarktech: Do these tests belong here? Seems like they're testing `cryptoModule.verifyTransaction`. Maybe the more appropriate test would be to see if `signature` or `signSignature` is equal to the result of calling (a stubbed) `cryptoModule.signTransaction`.

test/transactions/1_registerSecondPassphrase.js
```diff
@@ -178,6 +163,23 @@ describe('#registerSecondPassphrase transaction', () => {
 					emptyStringPublicKey,
 				);
 			});
+
+			describe('verification of the created transaction', () => {
```
willclarktech: As above.

test/transactions/0_send.js
```diff
@@ -145,6 +145,10 @@ describe('#send transaction', () => {
 						.and.be.type('object')
 						.and.be.empty();
 				});
+
+				it('should not have the second signature property', () => {
+					sendTransaction.should.not.have.property('signSignature');
+				});
```
willclarktech: In other test files you've moved these tests out of the `returned transaction block`. Best to be consistent.

PR #444, opened 2017-11-02
Add recipient public key to transaction creator - Closes #203

### LINKED ISSUE #203: Allow transaction objects to be generated with pubKey instead of address
Currently, in order to create a transaction we must pass the address of the recipientId. It would be nice if we could alternatively pass the publickey, and Lisk-JS would subsequently generate the address for us and create the transaction as normal. In this way, we could send "directly" to publickeys without an address on hand.

Tosch110:
Closes #203

### DIFF
--- a/src/transactions/0_send.js
+++ b/src/transactions/0_send.js
@@ -20,7 +20,7 @@ import cryptoModule from '../crypto';
 import { SEND_FEE, DATA_FEE } from '../constants';
 import {
 	prepareTransaction,
-	checkRecipientIdAndRecipientPublicKey,
+	getAddressAndPublicKeyFromRecipientData,
 } from './utils';
 import { getTimeWithOffset } from './utils/time';
 
@@ -47,7 +47,7 @@ export default function send({
 	data,
 	timeOffset,
 }) {
-	const { address, publicKey } = checkRecipientIdAndRecipientPublicKey({
+	const { address, publicKey } = getAddressAndPublicKeyFromRecipientData({
 		recipientId,
 		recipientPublicKey,
 	});
--- a/src/transactions/0_send.js
+++ b/src/transactions/0_send.js
@@ -18,13 +18,17 @@
  */
 import cryptoModule from '../crypto';
 import { SEND_FEE, DATA_FEE } from '../constants';
-import { prepareTransaction } from './utils';
+import {
+	prepareTransaction,
+	checkRecipientIdAndRecipientPublicKey,
+} from './utils';
 import { getTimeWithOffset } from './utils/time';
 
 /**
  * @method createTransaction
  * @param {Object} Object - Object
  * @param {String} Object.recipientId
+ * @param {String} Object.recipientPublicKey
  * @param {String} Object.amount
  * @param {String} Object.secret
  * @param {String} Object.secondSecret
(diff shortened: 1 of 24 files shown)


src/transactions/0_send.js
```diff
@@ -36,19 +40,25 @@ import { getTimeWithOffset } from './utils/time';
...
 export default function send({
 	recipientId,
+	recipientPublicKey,
 	amount,
 	secret,
 	secondSecret,
 	data,
 	timeOffset,
 }) {
+	const { address, publicKey } = checkRecipientIdAndRecipientPublicKey({
```
willclarktech: Since we go on to use the output of this function I think `check` is misleading. Maybe `getAddressAndPublicKey` or `getAddressAndPublicKeyFromRecipientData`?

src/transactions/utils/checkRecipientIdAndRecipientPublicKey.js
```diff
@@ -0,0 +1,49 @@
...
+}) {
+	if (recipientId && recipientPublicKey) {
+		const addressFromPublicKey = crypto.getAddressFromPublicKey(
+			recipientPublicKey,
+		);
+		if (recipientId === addressFromPublicKey) {
+			return { address: recipientId, publicKey: recipientPublicKey };
+		}
+		throw new Error(
+			'RecipientId and recipientPublicKey do not match. Please check.',
```
willclarktech: `Could not create transaction: recipientId does not match recipientPublicKey.`

src/transactions/utils/checkRecipientIdAndRecipientPublicKey.js
```diff
@@ -0,0 +1,49 @@
...
+		);
+	}
+
+	if (recipientId === undefined && recipientPublicKey) {
```
willclarktech: Why not just `!recipientId && recipientPublicKey`?

test/transactions/utils/checkRecipientIdAndRecipientPublicKey.js
```diff
@@ -0,0 +1,60 @@
...
+import { checkRecipientIdAndRecipientPublicKey } from '../../../src/transactions/utils';
+
+describe('#checkRecipientIdAndRecipientPublicKey', () => {
+	const recipientId = '18160565574430594874L';
+	const recipientPublicKey =
+		'5d036a858ce89f844491762eb89e2bfbd50a4a0a0da658e4b2628b25b117ae09';
+	const malformedRecipientPublicKey =
+		'12345a858ce89f844491762eb89e2bfbd50a4a0a0da658e4b2628b25b117ae09';
+
+	describe('it should check if a given recipientId and recipientPublicKey match', () => {
```
willclarktech: `When both recipientPublicKey and an address are given`

test/transactions/utils/checkRecipientIdAndRecipientPublicKey.js
```diff
@@ -0,0 +1,60 @@
...
+ * Removal or modification of this copyright notice is prohibited.
+ *
+ */
+import { checkRecipientIdAndRecipientPublicKey } from '../../../src/transactions/utils';
+
+describe('#checkRecipientIdAndRecipientPublicKey', () => {
+	const recipientId = '18160565574430594874L';
+	const recipientPublicKey =
+		'5d036a858ce89f844491762eb89e2bfbd50a4a0a0da658e4b2628b25b117ae09';
+	const malformedRecipientPublicKey =
```
willclarktech: This isn't malformed, it's just not the right one.

PR #451, opened 2017-11-07
:seedling: Export transaction utils with lisk-js - Closes #445

### LINKED ISSUE #445: Make publicly available prepareTransactions() from utils
For testing purposes in core, export `prepareTransaction` function is really useful.

Tosch110:
Closes #445

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -24,7 +24,6 @@ import api from './api/liskApi';
 import crypto from './crypto';
 import * as time from './transactions/utils/time';
 import transaction from './transactions';
-import * as utils from './transactions/utils';
 
 global.naclFactory = naclFactory;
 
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -24,6 +24,7 @@ import api from './api/liskApi';
 import crypto from './crypto';
 import * as time from './transactions/utils/time';
 import transaction from './transactions';
+import * as utils from './transactions/utils';
```
willclarktech: Prefer exporting it within the transactions module for access via `require('lisk-js').transactions.utils`, otherwise we'll have to refactor when it comes to switching to a monorepo.

src/index.js
```diff
@@ -32,4 +33,4 @@ naclFactory.instantiate(nacl => {
 	naclInstance = nacl;
 });
 
-export { crypto, transaction, api, time, Mnemonic };
+export { crypto, transaction, api, time, Mnemonic, utils };
```
willclarktech: Not directly related to this pr but this should be a default export no?

PR #455, opened 2017-11-14
Bitcore to bitcoinjs mnemonics - Closes #453

### LINKED ISSUE #453: Moving from bitcore to bitcoinjs mnemonics
The Bitcore (https://github.com/bitpay/bitcore-mnemonic) library is bigger than bitcoinjs library (https://github.com/bitcoinjs/bip39) and adds a lot of extra features for HD wallet or more languages. Which we all do not need. Therefore the smaller bitcoinjs mnemonic library fits better to our needs. When we want more languages we could still add them as needed.

Tosch110:
Closes #453

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,6 @@
 	"license": "GPL-3.0",
 	"dependencies": {
 		"babel-polyfill": "=6.23.0",
-		"bitcore-mnemonic": "LiskHQ/bitcore-mnemonic#v1.2.5",
 		"browserify-bignum": "=1.3.0-2",
 		"ed2curve": "=0.2.1",
 		"js-nacl": "LiskHQ/js-nacl#6dc1417",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -29,7 +29,7 @@
 	"license": "GPL-3.0",
 	"dependencies": {
 		"babel-polyfill": "=6.23.0",
-		"bitcore-mnemonic": "LiskHQ/bitcore-mnemonic#v1.2.5",
+		"bip39": "=2.4.0",
```
willclarktech: Do we not want to fork this ourselves?

Tosch110: We will probably do so in the future. Especially when we think about adding a new language to the wordlists (e.g. german). But that will probably not happen until lisk-js 1.1. Therefore for now we use the published version from bitcoinjs

PR #462, opened 2017-11-16
Purify functions - Closes #219

### LINKED ISSUE #219: Purify functions
In a lot of places we have functions which mutate an object passed as an argument and may or may not return the value that was assigned as well. This couples code and makes it difficult to reason about. A better approach would be to have pure functions, the result of which is assigned as needed.

E.g. https://github.com/LiskHQ/lisk-js/blob/development/lib/transactions/crypto.js#L396-L404

Instead of adding a signature to a transaction in an opaque way like this:
```js
crypto.sign(transaction); // has transaction been mutated? how?
```
we could simply return the signature from the function without mutation the transaction directly, and then assign it:
```js
transaction.signature = crypto.sign(transaction);
```
Or if we manage to do this pervasively we can just create new objects instead of mutating anything:
```js
const signedTransaction = {
  ...transaction,
  signature: crypto.sign(transaction),
};
```

willclarktech:
Closes #219 for now.
We should revisit this once we've thought more about state management in the API module.

### DIFF
--- a/src/transactions/1_registerSecondPassphrase.js
+++ b/src/transactions/1_registerSecondPassphrase.js
@@ -27,10 +27,10 @@ import { prepareTransaction, getTimeWithOffset } from './utils';
  * @return {Object}
  */
 
-function newSignature(secondSecret) {
+const createAsset = secondSecret => {
 	const { publicKey } = cryptoModule.getKeys(secondSecret);
-	return { publicKey };
-}
+	return { signature: { publicKey } };
+};
 
 /**
  * @method registerSecondPassphrase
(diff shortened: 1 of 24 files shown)


src/transactions/1_registerSecondPassphrase.js
```diff
@@ -27,10 +27,10 @@ import { prepareTransaction, getTimeWithOffset } from './utils';
  * @return {Object}
  */
 
-function newSignature(secondSecret) {
+const createSecondSignature = secondSecret => {
```
Tosch110: This name is still confusing, it just returns the publicKey of a secret.

src/transactions/1_registerSecondPassphrase.js
```diff
@@ -42,14 +42,10 @@ function newSignature(secondSecret) {
...
-export default function registerSecondPassphrase({
-	secret,
-	secondSecret,
-	timeOffset,
-}) {
+const registerSecondPassphrase = ({ secret, secondSecret, timeOffset }) => {
 	const keys = cryptoModule.getKeys(secret);
 
-	const signature = newSignature(secondSecret);
+	const signature = createSecondSignature(secondSecret);
```
Tosch110: Also here, prefer `secondPublicKey` instead of `signature`

PR #481, opened 2017-11-27
Add node build check - Closes #466

### LINKED ISSUE #466: Add an automated check for node distribution
Currently we have automated tests for source code and for browserified code, but have to manually check the node distribution.

willclarktech:
Closes #466

This just adds a basic check to see if requiring the build distribution for node errors. In the issue it suggested running the full test suite against the built files, so we might need to discuss this before merging.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -38,6 +38,7 @@ pipeline {
 			steps {
 				sh '''
 				npm run build
+				npm run build:check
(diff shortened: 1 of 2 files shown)


Tosch110:
First solution is to differentiate on the import with env variables. Then either run the tests against source code or distribution code. Would only need to be tested on release branches.

PR #493, opened 2017-12-11
Make constant ports a string - Closes #474

### LINKED ISSUE #474: Make constant ports a string
This is more uniform, we want to concatenate strings to form a URL

alepop:
close #474


Tosch110:
Yes, it should apply to all port values.

PR #497, opened 2017-12-13
:seedling: validate public keys - Closes #488

### LINKED ISSUE #488: Add check for valid public key
So that every public key passed into the transaction creator functions are checked and valid.

Tosch110:
Closes #488

### DIFF
--- a/src/transactions/utils/index.js
+++ b/src/transactions/utils/index.js
@@ -25,3 +25,4 @@ export {
 	verifyTransaction,
 } from './signAndVerify';
 export { getTimeFromBlockchainEpoch, getTimeWithOffset } from './time';
+export { validatePublicKey, validatePublicKeys } from './validatePublicKey';
--- a/src/transactions/utils/validatePublicKey.js
+++ b/src/transactions/utils/validatePublicKey.js
@@ -1,0 +1,32 @@
+/*
+ * Copyright © 2017 Lisk Foundation
+ *
+ * See the LICENSE file at the top-level directory of this distribution
+ * for licensing information.
+ *
+ * Unless otherwise agreed in a custom licensing agreement with the Lisk Foundation,
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ *
+ */
+import { bufferToHex, hexToBuffer } from '../../crypto/convert';
+
+export const validatePublicKey = publicKey => {
+	const publicKeyBuffer = hexToBuffer(publicKey);
+	if (bufferToHex(publicKeyBuffer) !== publicKey) {
+		throw new Error('Public key must be a hex string.');
+	}
+	if (publicKey.length !== 64 || publicKeyBuffer.length !== 32) {
+		throw new Error(
+			`Public key ${publicKey} length differs from the expected 64 hex characters (32 bytes) for a public key.`,
+		);
+	}
+
+	return publicKey;
+};
+
+export const validatePublicKeys = publicKeys =>
+	publicKeys.map(publicKey => validatePublicKey(publicKey));
--- a/src/transactions/utils/validatePublicKey.js
+++ b/src/transactions/utils/validatePublicKey.js
@@ -19,14 +19,14 @@ export const validatePublicKey = publicKey => {
 	if (bufferToHex(publicKeyBuffer) !== publicKey) {
 		throw new Error('Public key must be a valid hex string.');
 	}
-	if (publicKey.length !== 64 || publicKeyBuffer.length !== 32) {
+	if (publicKeyBuffer.length !== 32) {
 		throw new Error(
-			`Public key ${publicKey} length differs from the expected 64 hex characters (32 bytes) for a public key.`,
+			`Public key ${publicKey} length differs from the expected 32 bytes for a public key.`,
 		);
 	}
 
-	return publicKey;
+	return true;
 };
 
 export const validatePublicKeys = publicKeys =>
-	publicKeys.map(publicKey => validatePublicKey(publicKey));
+	publicKeys.every(publicKey => validatePublicKey(publicKey));
--- a/src/transactions/utils/validatePublicKey.js
+++ b/src/transactions/utils/validatePublicKey.js
@@ -29,4 +29,4 @@ export const validatePublicKey = publicKey => {
 };
 
 export const validatePublicKeys = publicKeys =>
-	publicKeys.every(publicKey => validatePublicKey(publicKey));
+	publicKeys.every(validatePublicKey);
--- a/src/transactions/utils/validatePublicKey.js
+++ b/src/transactions/utils/validatePublicKey.js
@@ -17,7 +17,7 @@ import { bufferToHex, hexToBuffer } from '../../crypto/convert';
 export const validatePublicKey = publicKey => {
 	const publicKeyBuffer = hexToBuffer(publicKey);
 	if (bufferToHex(publicKeyBuffer) !== publicKey) {
-		throw new Error('Public key must be a hex string.');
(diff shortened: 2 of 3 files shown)


src/transactions/utils/index.js
```diff
@@ -25,3 +25,4 @@ export {
 	verifyTransaction,
 } from './signAndVerify';
 export { getTimeFromBlockchainEpoch, getTimeWithOffset } from './time';
+export { validatePublicKey, validatePublicKeys } from './validatePublicKey';
```
willclarktech: We can't do `export * from './validatePublicKey';` until we support Babel 7 because of babel-transform, correct?

src/transactions/utils/validatePublicKey.js
```diff
@@ -0,0 +1,32 @@
...
+ *
+ */
+import { bufferToHex, hexToBuffer } from '../../crypto/convert';
+
+export const validatePublicKey = publicKey => {
+	const publicKeyBuffer = hexToBuffer(publicKey);
+	if (bufferToHex(publicKeyBuffer) !== publicKey) {
+		throw new Error('Public key must be a valid hex string.');
+	}
+	if (publicKey.length !== 64 || publicKeyBuffer.length !== 32) {
```
willclarktech: The important check here is the buffer length. If the public key made it past the first check, then checking for the string length is redundant here.

test/transactions/utils/validatePublicKey.js
```diff
@@ -0,0 +1,109 @@
...
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ *
+ */
+import {
+	validatePublicKey,
+	validatePublicKeys,
+} from '../../../src/transactions/utils/validatePublicKey';
+
+describe('validate public key', () => {
```
willclarktech: `public key validation`

test/transactions/utils/validatePublicKey.js
```diff
@@ -0,0 +1,109 @@
...
+	describe('#validatePublicKey', () => {
+		describe('Given an invalid hex public key', () => {
```
willclarktech: `Given a hex string with odd length`

test/transactions/utils/validatePublicKey.js
```diff
@@ -0,0 +1,109 @@
...
+			const invalidHexPublicKey =
+				'215b667a32a5cd51a94c9c2046c11fffb08c65748febec099451e3b164452bc';
+			it('should throw an error', () => {
+				validatePublicKey
+					.bind(null, invalidHexPublicKey)
+					.should.throw('Invalid hex string');
+			});
+		});
+
+		describe('Given an invalid public key', () => {
```
willclarktech: `Given a hex string with additional non-hex characters`

test/transactions/utils/validatePublicKey.js
```diff
@@ -0,0 +1,109 @@
...
+			it('should throw an error', () => {
+				validatePublicKey
+					.bind(null, tooShortPublicKey)
+					.should.throw(
+						'Public key 215b667a32a5cd51a94c9c2046c11fffb08c65748febec099451e3b164452b length differs from the expected 64 hex characters (32 bytes) for a public key.',
+					);
+			});
+		});
+
+		describe('Given a correct public key', () => {
```
willclarktech: `Given a valid public key`

test/transactions/utils/validatePublicKey.js
```diff
@@ -0,0 +1,109 @@
...
+			const publicKey =
+				'215b667a32a5cd51a94c9c2046c11fffb08c65748febec099451e3b164452bca';
+			it('should return the public key', () => {
+				validatePublicKey(publicKey).should.be.equal(publicKey);
+			});
+		});
+
+		describe('Given a correct, all-numeric public key', () => {
```
willclarktech: As above

src/transactions/utils/validatePublicKey.js
```diff
@@ -0,0 +1,32 @@
...
+		throw new Error(
+			`Public key ${publicKey} length differs from the expected 64 hex characters (32 bytes) for a public key.`,
+		);
+	}
+
+	return publicKey;
```
willclarktech: Do you think it makes more sense to return the public key or `true`?

src/transactions/utils/validatePublicKey.js
```diff
@@ -0,0 +1,32 @@
...
+};
+
+export const validatePublicKeys = publicKeys =>
+	publicKeys.map(publicKey => validatePublicKey(publicKey));
```
willclarktech: As above re return value

src/transactions/utils/index.js
Tosch110: Yes, looks like it
https://github.com/babel/babel/issues/2877

src/transactions/utils/validatePublicKey.js
Tosch110: `true` also works for me. From the name it seems to be more consistent. thx

src/transactions/utils/validatePublicKey.js
```diff
@@ -19,14 +19,14 @@ export const validatePublicKey = publicKey => {
...
 		);
 	}
 
-	return publicKey;
+	return true;
 };
 
 export const validatePublicKeys = publicKeys =>
-	publicKeys.map(publicKey => validatePublicKey(publicKey));
+	publicKeys.every(publicKey => validatePublicKey(publicKey));
```
willclarktech: 👌 for the `.every` but `publicKeys.every(validatePublicKey)` is even 👌 er.

test/transactions/utils/validatePublicKey.js
```diff
@@ -58,24 +58,38 @@ describe('validate public key', () => {
...
 			const publicKey =
 				'215b667a32a5cd51a94c9c2046c11fffb08c65748febec099451e3b164452bca';
 			it('should return the public key', () => {
-				validatePublicKey(publicKey).should.be.equal(publicKey);
+				validatePublicKey(publicKey).should.be.true();
 			});
 		});
 
-		describe('Given a correct, all-numeric public key', () => {
+		describe('Given a valid public key', () => {
```
willclarktech: Still needs the `all-numeric` part: `Given a valid public key with only numeric characters`?

test/transactions/utils/validatePublicKey.js
```diff
@@ -58,24 +58,38 @@ describe('validate public key', () => {
...
+				validatePublicKeys
+					.bind(null, publicKeys)
+					.should.throw('Invalid hex string');
+			});
+		});
+
+		describe('Given a valid public key', () => {
 			const publicKey =
 				'215b667a32a5cd51a94c9c2046c11fffb08c65748febec099451e3b164452bca';
 			it('should return the public key', () => {
```
willclarktech: Descriptions need updating

test/transactions/utils/validatePublicKey.js
```diff
@@ -0,0 +1,109 @@
...
+			it('should throw an error', () => {
+				validatePublicKey
+					.bind(null, tooShortPublicKey)
+					.should.throw(
+						'Public key 215b667a32a5cd51a94c9c2046c11fffb08c65748febec099451e3b164452b length differs from the expected 32 bytes for a public key.',
+					);
+			});
+		});
+
+		describe('Given an array of public keys with one invalid public key', () => {
```
willclarktech: Why did this get moved here?

test/transactions/utils/validatePublicKey.js
```diff
@@ -0,0 +1,109 @@
...
+			const publicKey =
+				'1234567812345678123456781234567812345678123456781234567812345678';
+			it('should return the public key', () => {
+				validatePublicKey(publicKey).should.be.true();
+			});
+		});
+	});
+
+	describe('#validatePublicKeys', () => {
+		describe('Given an array of public keys', () => {
```
willclarktech: `valid public keys`

Issue #475, opened 2017-11-23
Improve communication of headers


Tosch110:
- [ ] Consistency in capital naming (nethash)
- [ ] Rename functions and objects that are about headers rather than nethash
- [ ] See how much we can strip out / update on the headers
- [ ] Move nethash into constants
- [ ] Get rid of broadhash
- [ ] Structure of the returned object (netHashOptions)

willclarktech:
@Tosch110 I think we can close this as it should now be covered as part of https://github.com/LiskHQ/lisk-js/issues/491

PR #512, opened 2017-12-22
Improve API class structure - Closes #491

### LINKED ISSUE #491: Create a coherent and secure structure in liskAPI
As discussed in the meeting, liskAPI should be clear and logical on types and functions. 

How and when they get initialised, what type each of the variables has and what they will be used for.

willclarktech:
Closes #491 

Very much WIP. Tests are failing currently and incomplete.

There are a bunch of ideas here:
1. Tidied up the constructor re types, ordering, defaults, new arrays rather than references to the config export etc
1. Alphabetized some things
1. Moved a bunch of methods from privateApi into the class if they seemed like they might be useful. Could be followed up by moving the remaining functions into utils or making them private in the style of getters/setters described below
1. Updated a few names here and there to make it clearer
1. Made definition of prototype methods more succinct/DRY
1. Made setting of ssl and testnet match each other
1. Used `get` and `set` where it seemed like it might be appropriate
1. Upgraded sinon so we can stub getters and setters via the sandbox (also lets us use fake timers with the sandbox, although see https://github.com/sinonjs/sinon/issues/550)
1. Used symbols for private properties
1. Ignored nethash (which will be covered by #475)

I'm not fully convinced by all of these. In particular, the getters, setters and symbols seem to make things more opaque and less easy to test in isolation. Also some of those private api methods maybe should remain private. It would be good to go through exactly what we want to expose/conceal and why.


Reshetnyak:
I think, you could leverage some code from my PRs

mattsecrets (commented):
1. At first I thought it was odd that many of the functions check for nodes on the mainnet and testnet together. But I realize that this probably reduces code clutter and the necessity of each network having their own, very similar, functions. It also might make it easier for users/developers to switch between the testnet and the mainnet.

2. I notice there is use of properties like `this.ssl` vs `this[ssl]`.  I see that Symbols are being utilized at the top of the file, so it's not clear if we are working with two different properties, or the same, or if they are being mixed up.

3. In general, I did not see much in this commit that warrants a refactor to ES6/ES7.  To me, most of the code looks well arranged and mostly consistent.  

4. Also, certain items that I recommend, such as destructuring properties from objects, may be reasonable from a one-off perspective. For example, most of this code uses the `this` keyword.
        To only destructure properties off of `this` for one example may go against a Lisk style-guide for consistency. 
        So I hesitate to make some of these comments, because certain refactors might violate the consistency of the code in other places.

5. Overall, I think that the code is well organized, the naming patterns are descriptive and semantic, and the tests seem comprehensive. I think that in certain places, the format could be more consistent with the rest of the codebase, but this is issue is minor.

src/api/liskApi.js
```diff
@@ -12,183 +12,177 @@
...
-* @class LiskAPI
-* @param {Object} options - Initialization Object for the LiskAPI instance.
- */
+const bannedNodes = Symbol('bannedNodes');
+const defaultNodes = Symbol('defaultNodes');
+const defaultSSLNodes = Symbol('defaultSSLNodes');
+const defaultTestnetNodes = Symbol('defaultTestnetNodes');
+const nethash = Symbol('nethash');
+const ssl = Symbol('ssl');
+const testnet = Symbol('testnet');
```
shuse2: what is the reason to set Symbol for these?
It might be solved by introducing flow?
https://github.com/LiskHQ/lisk-js/pull/556/files#diff-3aefae204ccf5f3f6e65ce8890b8c25cR45

src/api/liskApi.js
```diff
@@ -12,183 +12,177 @@
...
 		};
-
 		return privateApi.sendRequestPromise
 			.call(this, POST, request)
 			.then(result => result.body)
 			.then(utils.optionallyCallCallback.bind(null, callback));
 	}
 
-	broadcastSignatures(signatures, callback) {
+	broadcastSignedTransaction(transaction, callback) {
```
shuse2: Are we leaving callback for compatibility?

src/api/privateApi.js
```diff
@@ -208,11 +115,11 @@ export function handleSendRequestFailures(
 	error,
 ) {
 	const that = this;
-	if (hasAvailableNodes.call(that)) {
+	if (that.hasAvailableNodes()) {
```
shuse2: I like this format =)

src/api/liskApi.js
willclarktech: It's a way of making them private since there's no way to access the symbol outside of the module.

src/api/liskApi.js
willclarktech: No, this PR was made before https://github.com/LiskHQ/lisk-js/pull/545 was merged and will need to be updated.

shuse2:
- getter, setter => only when it requires, virtual variables, do PR and go on from there
- private variable => no private variable
- private api class, createReqObj, sendReqObj => util, sendReqPromise => merge with sendReq
- renaming Promise function and wrapper function
- moving private api to utils or api(if it uses this)
- definition of prototype
- option type def => create get, update method
- nethash is more about header

Separate PR:
- line 105 liskAPI test port / ssl port logic
- think about the timestamp in transaction
- transaction API body

src/api/liskApi.js
```diff
@@ -13,211 +13,300 @@
...
 	}
 
-	broadcastTransactions(transactions) {
-		return privateApi.sendRequestPromise
-			.call(this, POST, 'transactions', transactions)
-			.then(result => result.body);
+	banActiveNode() {
+		if (!this.isBanned(this.node)) {
+			this.bannedNodes.push(this.node);
+			this.node = this.selectNewNode();
```
shuse2: Selecting new node is ambiguous side effect.

src/api/liskApi.js
```diff
@@ -13,211 +13,300 @@
...
-	getNethash(providedNethash) {
-		const { port } = this;
-		const NetHash = this.testnet
-			? utils.netHashOptions({ port }).testnet
-			: utils.netHashOptions({ port }).mainnet;
+	get allNodes() {
+		return {
+			current: this.node,
+			default: this.defaultNodes,
+			ssl: this.defaultSSLNodes,
```
shuse2: why are these all ports are saved?

src/api/liskApi.js
willclarktech: Might be solvable by renaming?

shuse2:
I think great sample for this PR would be
[https://github.com/stripe/stripe-node](https://github.com/stripe/stripe-node)

shuse2:
Changes done
- Followed [https://github.com/stripe/stripe-node](https://github.com/stripe/stripe-node)
- handleTimestampInFuture is deleted for now
- API retry can be used in any API now by setting retry to true
- api methods are defined like `api({}).accounts.get(query);`
- new structure supports `params`, like `/accounts//{address}`
- new structure supports any `query` or `body`
- new structure supports validator function
- new structure supports defining default `body` or `query`

test/api/liskApi.js
```diff
@@ -307,73 +528,86 @@ describe('Lisk API module', () => {
...
 			});
 
 			describe('when initially on mainnet', () => {
 				it('should reset banned nodes when switching from mainnet to testnet', () => {
 					LSK.setTestnet(true);
-					return LSK.should.have
-						.property('bannedNodes')
-						.and.be.Array()
-						.and.be.empty();
+					return defaultNodes.every(node =>
```
willclarktech: Why `.every` rather than `.forEach`?

test/api/liskApi.js
```diff
@@ -307,73 +528,86 @@ describe('Lisk API module', () => {
...
+					const bannedNodes = defaultNodes.filter(node => LSK.isBanned(node));
 					LSK.setTestnet(false);
-					return LSK.should.have
-						.property('bannedNodes')
-						.and.be.eql(defaultbannedNodes);
+					return bannedNodes.every(node => LSK.isBanned(node).should.be.true());
 				});
 			});
 
 			describe('when initially on testnet', () => {
```
willclarktech: The tests in this suite are meaningless now.

test/api/liskApi.js
```diff
@@ -383,11 +617,11 @@ describe('Lisk API module', () => {
 				});
 
 				it('should not change bannedNodes', () => {
-					LSK.bannedNodes = [].concat(defaultbannedNodes);
+					LSK.bannedNodes = [].concat(defaultBannedNodes);
```
willclarktech: Prefer `[...defaultBannedNodes]`

test/api/liskApi.js
```diff
@@ -307,73 +528,86 @@ describe('Lisk API module', () => {
...
+		beforeEach(() => {
+			selectNewNodeStub = sandbox
+				.stub(LSK, 'selectNewNode')
+				.returns(defaultSelectedNode);
+		});
+
 		describe('when ssl is initially true', () => {
 			beforeEach(() => {
-				LSK.ssl = true;
+				LSK.setSSL(true);
```
willclarktech: Should be set directly for setup.

test/api/liskApi.js
```diff
@@ -398,18 +632,29 @@ describe('Lisk API module', () => {
 			});
 
 			describe('when set to false', () => {
+				beforeEach(() => {
+					LSK.setSSL(true);
```
willclarktech: Doesn't match suite description.

test/api/liskApi.js
```diff
@@ -398,18 +632,29 @@ describe('Lisk API module', () => {
...
+				});
+				it('should set port to ssl port', () => {
+					LSK.setSSL(false);
+					return LSK.should.have.property('port').and.be.equal(livePort);
```
willclarktech: Doesn't match test description.

test/api/resources/accounts.js
```diff
@@ -0,0 +1,62 @@
...
+			fullURL: defaultBasePath,
+			hasAvailableNodes: () => {},
+			randomizeNodes: () => {},
+			banActiveNodeAndSelect: () => {},
+		};
+		resource = new AccountResource(LiskAPI);
+	});
+
+	describe('#constructor', () => {
+		it('should throw error without LiskAPI input', () => {
```
willclarktech: What if it's called with something, but it's not an instance of `LiskAPI`?

test/api/resources/accounts.js
```diff
@@ -0,0 +1,62 @@
...
+			return (() => new AccountResource()).should.throw(
+				'Require LiskAPI instance to be initialized.',
```
willclarktech: `AccountResource constructor requires a LiskAPI instance for initialization.`

test/api/resources/accounts.js
```diff
@@ -0,0 +1,62 @@
...
+			);
+		});
+
+		it('should be instance of APIResource', () => {
+			return resource.should.be.instanceOf(APIResource);
+		});
+
+		it('should have correcrt full path', () => {
```
willclarktech: Typo

test/api/resources/accounts.js
```diff
@@ -0,0 +1,62 @@
...
+ * Removal or modification of this copyright notice is prohibited.
+ *
+ */
+
+import APIResource from 'api/apiResource';
+import AccountResource from 'api/resources/accounts';
+
+describe('AccountResouce', () => {
+	const defaultBasePath = 'http://localhost:1234';
+	const defaultPath = '/accounts';
```
willclarktech: This is a specific value, so should be named something like `path` or `accountsPath`.

test/api/resources/accounts.js
```diff
@@ -0,0 +1,62 @@
...
+			return resource.should.be.instanceOf(APIResource);
+		});
+
+		it('should have correcrt full path', () => {
+			return resource.resourcePath.should.eql(
+				`${defaultBasePath}/api${defaultPath}`,
+			);
+		});
+
+		it('should set resource path', () => {
```
willclarktech: This test belongs on the parent, not here.

test/api/resources/accounts.js
```diff
@@ -0,0 +1,62 @@
...
+			return resource.path.should.equal(defaultPath);
+		});
+
+		it('should have methods', () => {
```
willclarktech: We should test that it calls `apiMethod` correctly.

test/api/resources/accounts.js
```diff
@@ -0,0 +1,62 @@
...
+			return resource.should.have.keys('path', 'get');
```
willclarktech: We already tested `path`.

test/api/resources/votes.js
```diff
@@ -0,0 +1,62 @@
...
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ *
+ */
+
+import APIResource from 'api/apiResource';
+import VoteResource from 'api/resources/votes';
+
+describe('VoteResouce', () => {
```
willclarktech: Typo.

test/api/utils.js
```diff
@@ -14,109 +14,12 @@
...
-			return result.should.equal(`${URLPrefix}://${LSK.node}:${defaultPort}`);
-		});
-
-		it('should not include a port if not set', () => {
-			delete LSK.port;
-			result = getFullURL(LSK);
-			return result.should.equal(`${URLPrefix}://${LSK.node}`);
-		});
-	});
+const defaultURL = 'http://localhost:8080/api/resources';
```
willclarktech: Prefer nesting variables under the relevant suite.

test/api/utils.js
```diff
@@ -14,109 +14,12 @@
...
 
+describe('api utils module', () => {
 	describe('#toQueryString', () => {
-		const { toQueryString } = utils;
-
 		it('should create a query string from an object', () => {
-			const queryString = toQueryString({
+			const queryString = utils.toQueryString({
```
willclarktech: Why remove the destructuring?

test/api/utils.js
```diff
@@ -125,103 +28,44 @@ describe('api utils module', () => {
...
-					badKey: NaN,
-				},
-				goodOptions,
-			);
-
-			return checkOptions
-				.bind(null, optionsWithNaN)
-				.should.throw('"badKey" option should not be NaN');
+		it('should throw error if url has variable but no param', () => {
+			const fn = () => utils.solveURLParams(`${defaultURL}/{id}`);
```
willclarktech: Prefer `.bind` syntax.

test/api/utils.js
```diff
@@ -125,103 +28,44 @@ describe('api utils module', () => {
...
-				return returnedFunction.call(LSK, value).then(result => {
-					result.should.be.eql(sendRequestResult);
-				});
+		it('should replace variable with correct id and encode special characters', () => {
+			const solvedURL = utils.solveURLParams(`${defaultURL}/{id}`, {
+				id: '456ß1234sd',
+				accountId: '123',
 			});
+			return solvedURL.should.be.equal(`${defaultURL}/456%C3%9F1234sd`);
 		});
```
willclarktech: What about multiple parameters?

willclarktech (commented):
I'm the original PR author so I can't select "Request changes".

src/api/apiMethod.js
```diff
@@ -0,0 +1,76 @@
...
+const apiMethod = params => {
+	const spec = params || {};
+	const method = spec.method || GET;
+	const path = spec.path || '';
+	const urlParams = spec.urlParams || [];
+	const validator = spec.validator || null;
+	const defaultData = spec.defaultData || {};
+	const retry = spec.retry || false;
+
+	return function apiHandler(...args) {
```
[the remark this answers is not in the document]
shuse2: I think use of `Promise` and `async` should be consistent.
Should we start to use `async` syntax?

src/api/apiMethod.js
```diff
@@ -0,0 +1,76 @@
...
+			}
+			const urlData = {};
+			urlParams.forEach(param => {
+				urlData[param] = args.shift();
+			});
+			fullURL = utils.solveURLParams(fullURL, urlData);
+		}
+		// last argument is data
+		// # same as length of urlParams is the arguments, and the last one is data(body or query)
+		const data = Object.assign(
```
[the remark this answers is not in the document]
shuse2: at this point, this `data` can be either `query` or `body`.
Im not sure if `requestData` is `metadata` coz it's not supportive data itself.
Maybe, it would be improved by changing `requestData` to just `req`?

src/api/index.js
```diff
@@ -0,0 +1,50 @@
...
+
+api({}).signatures.create(body);
+
+api({}).transactions.get(query);
+api({}).transactions.create(body);
+
+api({}).votes.get(query);
+api({}).voters.get(query);
+
+*/
```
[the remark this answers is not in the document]
shuse2: Will be deleted, we should put this in `README` or other document

src/api/liskApi.js
```diff
@@ -12,212 +12,184 @@
...
+	'Content-Type': 'application/json',
+	os: 'lisk-js-api',
+	version: '1.0.0',
+	minVersion: '>=0.5.0',
+};
 
 export default class LiskAPI {
 	constructor(providedOptions) {
 		const options = Object.assign({}, config.options, providedOptions);
+		this.ssl = options.ssl !== undefined ? options.ssl : true;
```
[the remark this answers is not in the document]
shuse2: Although needs of `ssl` option should be discussed

src/api/liskApi.js
```diff
@@ -12,212 +12,184 @@
...
 		}
+		return this.defaultNodes;
+	}
 
-		return NetHash;
+	get urlPrefix() {
+		if (this.ssl) {
+			return 'https';
+		}
+		return 'http';
```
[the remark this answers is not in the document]
shuse2: Needs of global change in `ssl` should be discussed

src/api/liskApi.js
```diff
@@ -12,212 +12,184 @@
...
 	}
 
-	setTestnet(testnet) {
-		if (this.testnet !== testnet) {
-			this.bannedNodes = [];
+	getDefaultPort() {
+		if (this.testnet) {
+			return TEST_PORT;
+		}
+		if (this.ssl) {
```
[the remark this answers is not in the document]
shuse2: Also, need to confirm if all the `testnet` nodes cannot be used with `ssl`

src/api/liskApi.js
```diff
@@ -12,212 +12,184 @@
...
 		}
-		this.testnet = testnet;
-		this.port = testnet ? TEST_PORT : LIVE_PORT;
 
-		privateApi.selectNewNode.call(this);
+		const randomIndex = Math.floor(Math.random() * nodes.length);
+		return nodes[randomIndex];
+	}
+
+	setSSL(newSSLValue) {
```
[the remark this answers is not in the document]
shuse2: why do we need default value for setter?

src/api/utils.js
```diff
@@ -83,3 +23,21 @@ export const toQueryString = obj => {
 
 	return parts.join('&');
 };
+
+const urlParamRegex = /{[^}]+}/i;
+export const solveURLParams = (url, params) => {
```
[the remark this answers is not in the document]
shuse2: Yes, this function will be used when we have
`/accounts/{accountId}`
like this.

src/api/utils.js
```diff
@@ -83,3 +23,21 @@ export const toQueryString = obj => {
 
 	return parts.join('&');
 };
+
+const urlParamRegex = /{[^}]+}/i;
```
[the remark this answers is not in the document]
shuse2: Are we going to allow special character in `dapp` name?
but in any case, the url params should be solved before injecting query, so it should not be a problem

test/api/apiMethod.js
```diff
@@ -0,0 +1,184 @@
...
+			it('should return function', () => {
+				return handler.should.be.type('function');
+			});
+
+			it('should be rejected with error without param', () => {
+				return handler().should.be.rejectedWith(errorArgumentNumber);
+			});
+
+			it('should be rejected with error without enough param', () => {
+				return handler().should.be.rejectedWith(errorArgumentNumber);
```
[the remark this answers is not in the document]
shuse2: mmm it does look the same

test/api/apiResource.js
```diff
@@ -0,0 +1,191 @@
...
+	beforeEach(() => {
+		LiskAPI = {
+			headers: defaultHeaders,
+			fullURL: defaultBasePath,
+			hasAvailableNodes: () => {},
+			randomizeNodes: () => {},
+			banActiveNodeAndSelect: () => {},
+		};
+
+		banActiveNodeAndSelectStub = sandbox.stub(
```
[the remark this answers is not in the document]
shuse2: you mean for `fullURL` right?

test/api/apiResource.js
```diff
@@ -0,0 +1,191 @@
...
+			LiskAPI,
+			'banActiveNodeAndSelect',
+		);
+		resource = new APIResource(LiskAPI);
+		handleRetrySpy = sandbox.spy(resource, 'handleRetry');
```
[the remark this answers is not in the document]
shuse2: we want to make sure the handleRetry is called, but we dont't need the behavior of the handleRetry

test/api/resources/accounts.js
shuse2: That would be a problem.
However, importing lisk-api from resource will have circular dependency.
Any suggestions?

test/api/resources/accounts.js
shuse2: This behavior is defined in `APIResource`, so I am not sure if we should extend this just to add message

test/api/resources/accounts.js
shuse2: This is checking if the path of AccountResource is set to `/accounts`.
It is predefined value, so it might not worth checking though.

test/api/resources/accounts.js
shuse2: It is true, but once we move to the `chai` test, we will change this to `all.keys`.
Therefore, we should try to list all of them

test/api/utils.js
shuse2: No idea, but probably we should be consistent with `require` vs `import` and the way of exporting

src/api/apiMethod.js
```diff
@@ -0,0 +1,76 @@
...
+	const path = spec.path || '';
+	const urlParams = spec.urlParams || [];
+	const validator = spec.validator || null;
+	const defaultData = spec.defaultData || {};
+	const retry = spec.retry || false;
+
+	return function apiHandler(...args) {
+		// this refers to resource class
+		const self = this;
+		let fullURL = self.resourcePath + path;
```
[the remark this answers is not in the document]
shuse2: Any suggestion?
Also, why avoid let where `fullURL` can be changed afterwords?

src/api/liskApi.js
```diff
@@ -12,212 +12,184 @@
...
 			options.port === '' || options.port
 				? options.port
-				: utils.getDefaultPort(options);
-		this.nethash = this.getNethash(options.nethash);
+				: this.getDefaultPort();
+		this.randomizeNodes =
+			options.randomizeNodes !== undefined ? options.randomizeNodes : true;
+		this.providedNode = options.node || null;
+		this.node = options.node || this.selectNewNode();
+		this.headers = this.getDefaultHeaders(nethash);
```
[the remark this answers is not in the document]
shuse2: Could you elaborate?
It was originally like `this.nethash = this.getNethash(options.nethash);`

test/api/liskApi.js
```diff
@@ -13,85 +13,50 @@
...
 		version: '1.0.0',
 		minVersion: '>=0.5.0',
 		port: sslPort,
-		Accept: 'application/json',
 	};
-	const testnetNethash = Object.assign({}, defaultNethash, {
+	const testnetHeaders = Object.assign({}, defaultHeaders, {
 		nethash: testnetHash,
-		broadhash: testnetHash,
+		version: '0.0.0a',
```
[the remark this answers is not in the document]
shuse2: I just kept current behavior
`NetHash.version = '0.0.0a';`

test/api/liskApi.js
```diff
@@ -192,109 +158,364 @@ describe('Lisk API module', () => {
...
+
+			it('should set bannedNodes when passed as an option', () => {
+				const bannedNodes = ['a', 'b'];
+				LSK = new LiskAPI({ bannedNodes });
+				return LSK.should.have.property('bannedNodes').be.eql(bannedNodes);
+			});
+		});
+
+		describe('headers', () => {
+			it('should set with passed nethash', () => {
```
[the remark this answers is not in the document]
shuse2: In this case, the paragraph is right, input of `LiskAPI` should be `{ headers: {nethash: '123'}}` will be more clear

test/api/liskApi.js
```diff
@@ -192,109 +158,364 @@ describe('Lisk API module', () => {
...
+		});
+	});
+
+	describe('get fullURL', () => {
+		it('should return with set port', () => {
+			LSK = new LiskAPI({ port: '8080', node: localNode });
+			return LSK.fullURL.should.be.equal('https://localhost:8080');
+		});
+
+		it('should return default port if it is not set', () => {
```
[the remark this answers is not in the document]
shuse2: should not include port in URL if port is not set

test/api/liskApi.js
```diff
@@ -192,109 +158,364 @@ describe('Lisk API module', () => {
...
-			nodes.should.have
-				.property('ssl')
-				.have.property('length')
-				.be.equal(8);
-			return nodes.ssl.forEach(node => {
-				node.should.have.property('node').and.be.type('string');
-				node.should.have.property('ssl').and.be.true();
+		describe('if a node was not provided in the options', () => {
+			beforeEach(() => {
+				LSK.providedNode = undefined;
```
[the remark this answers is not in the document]
shuse2: In this case, we are testing `if no option was provided`, which will not set the providedNode.
That's why it's `undefined` rather than null.

test/api/resources/accounts.js
shuse2: `apiMethod` is tested separately. If we stub the method, is it meaningful to test it here?

src/api/liskApi.js
```diff
@@ -12,212 +12,184 @@
...
-		this.node = node || privateApi.selectNewNode.call(this);
-		return this.node;
+	getDefaultHeaders(providedNethash) {
+		const headers = Object.assign({}, commonHeaders, {
+			port: this.port,
+			nethash: this.testnet ? TESTNET_NETHASH : MAINNET_NETHASH,
+		});
+		if (providedNethash) {
+			headers.nethash = providedNethash;
+			headers.version = '0.0.0a';
```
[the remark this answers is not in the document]
shuse2: should we change it to be able to customize whole header?

src/api/liskApi.js
```diff
@@ -12,212 +12,184 @@
...
+	setSSL(newSSLValue) {
+		if (this.ssl !== newSSLValue) {
+			const nonSSLPort = this.testnet ? TEST_PORT : LIVE_PORT;
+
+			this.ssl = newSSLValue;
+			this.port = newSSLValue ? SSL_PORT : nonSSLPort;
+			this.bannedNodes = [];
+
+			this.selectNewNode();
+		}
```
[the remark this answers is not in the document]
shuse2: Also, should we return some value if this is `setter`?

src/api/liskApi.js
```diff
@@ -12,212 +12,184 @@
...
+		if (this.testnet !== newTestnetValue) {
+			const nonTestnetPort = this.ssl ? SSL_PORT : LIVE_PORT;
+
+			this.testnet = newTestnetValue;
+			this.port = newTestnetValue ? TEST_PORT : nonTestnetPort;
 			this.bannedNodes = [];
-			privateApi.selectNewNode.call(this);
+
+			this.selectNewNode();
 		}
```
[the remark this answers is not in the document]
shuse2: same as above

src/api/liskApi.js
```diff
@@ -12,212 +12,184 @@
...
+	selectNewNode() {
+		if (this.randomizeNodes) {
+			return this.getRandomNode();
+		} else if (this.providedNode) {
+			if (this.isBanned(this.providedNode)) {
+				throw new Error(
+					'Cannot select node: provided node has been banned and randomizeNodes is not set to true.',
+				);
+			}
+			return this.providedNode;
```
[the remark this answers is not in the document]
shuse2: it's true it's very confusing, but throwing Error would be confusing at the same time.
Probably, we should determine what is actually `providedNode, nodes, randomNodes` intended  to do

test/api/apiMethod.js
```diff
@@ -0,0 +1,184 @@
...
+			});
+
+			it('should be rejected with no data', () => {
+				return handler('r-123', 'id-123').should.be.rejectedWith(
+					Error('no data'),
+				);
+			});
+
+			it('should be request with the given data', () => {
```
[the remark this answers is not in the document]
shuse2: In this case, `should call request with given data` since it's referring method `request`?

src/api/apiMethod.js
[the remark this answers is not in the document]
willclarktech: The question is: why make something that can change, rather than two things that don't change?

```js
// defined outside the function?
const addParamsToURL = (params, url, args) => {
    if (args.length < urlParams.length) {
        throw new Error('xxx');
    }
    const data = params.reduce((accumulator, param, i) => Object.assign({}, accumulator, { [param]: args[i] }), {});
    return utils.solveURLParams(url, data);
};

// inside the function
// ...
const urlWithoutParams = `${this.resourcePath}${path}`;
const fullURL = urlParams.length
    ? addParamsToURL(urlParams, urlWithoutParams, args)
    : urlWithoutParams;
```

test/api/utils.js
willclarktech: The rule is: use `import` unless you need to use `require` for stubbing purposes.

test/api/liskApi.js
```diff
@@ -192,109 +158,364 @@ describe('Lisk API module', () => {
...
+		});
+
 		describe('to true', () => {
 			beforeEach(() => {
-				LSK.setTestnet(true);
+				LSK.setTestnet(false);
 			});
 
 			it('should set testnet to true', () => {
+				LSK.setTestnet(true);
```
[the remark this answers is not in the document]
shuse2: This should be in `it` to make it parallel with another test below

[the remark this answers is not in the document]
willclarktech: The `callCount` could also be recorded in the `beforeEach` hook, but if `LSK.setTestnet(false)` is changed to `LSK.testnet = false` it should be removable anyway.

src/api/apiMethod.js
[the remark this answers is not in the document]
shuse2: I came up with the idea without using `let` or two variables 👍

src/api/apiMethod.js
```diff
@@ -0,0 +1,80 @@
...
+		defaultData = {},
+		retry = false,
+	} = params;
+
+	return function apiHandler(...args) {
+		if (urlParams.length > 0 && args.length < urlParams.length) {
+			return Promise.reject(
+				new Error(
+					'Arguments must include Params defined. Required: ',
+					urlParams.toString(),
```
willclarktech: I think you want a string template here.

src/api/apiMethod.js
```diff
@@ -0,0 +1,80 @@
...
+// Bind to resource class
+const apiMethod = (params = {}) => {
+	const {
+		method = GET,
+		path = '',
+		urlParams = [],
+		validator = null,
+		defaultData = {},
+		retry = false,
+	} = params;
```
willclarktech: Why not do this directly in the parameters brackets for the function?

src/api/apiMethod.js
```diff
@@ -0,0 +1,80 @@
...
+
+	return function apiHandler(...args) {
+		if (urlParams.length > 0 && args.length < urlParams.length) {
+			return Promise.reject(
+				new Error(
+					'Arguments must include Params defined. Required: ',
```
willclarktech: `This endpoint must be supplied with the following parameters: `

src/api/apiMethod.js
```diff
@@ -0,0 +1,80 @@
...
+					urlParams.toString(),
+				),
+			);
+		}
+
+		const data = Object.assign(
+			{},
+			defaultData,
+			args.length > urlParams.length &&
+			typeof args[urlParams.length] === 'object'
```
willclarktech: What if there are additional args beyond `urlParams.length`?

src/api/apiMethod.js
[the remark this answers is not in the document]
willclarktech: Maybe not if it's just this function.

src/api/apiMethod.js
shuse2: Not used!
Always, args[urlPrams.length] is the body or query object.

src/api/apiMethod.js
shuse2:
```
const apiMethod = ({
		method = GET,
		path = '',
		urlParams = [],
		validator = null,
		defaultData = {},
		retry = false,
	}) => function apiHandler(...args) {
```
wouldn't it be harder to read?

src/api/liskApi.js
```diff
@@ -12,212 +12,176 @@
...
+const commonHeaders = {
+	'Content-Type': 'application/json',
+	os: 'lisk-js-api',
+	version: '1.0.0',
+	minVersion: '>=0.5.0',
+};
 
 export default class LiskAPI {
 	constructor(providedOptions) {
 		const options = Object.assign({}, config.options, providedOptions);
```
willclarktech: This needs to change to `Object.assign({ headers: {} }, config.options, providedOptions);` otherwise the code below can error.

src/api/liskApi.js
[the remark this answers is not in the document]
willclarktech: It's just confusing having such a combination of headers and nethash. Maybe (especially if `version` is configurable) we can change this to an object which takes a `nethash` key?

src/api/liskApi.js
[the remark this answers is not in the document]
willclarktech: I would say so. Also this whole function could be much cleaner.

src/api/liskApi.js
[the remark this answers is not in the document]
willclarktech: This isn't strictly a `setter`. I could imaging that people might call `.setSSL()` thinking it would default to `true`. And if they do that, `undefined` is going to trigger a new node selection whether `ssl` is currently `true` or `false`.

src/api/liskApi.js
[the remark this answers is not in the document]
willclarktech: Maybe!

src/api/resources/accounts.js
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ *
+ */
+
+import { GET } from 'constants';
+import APIResource from '../apiResource';
+import apiMethod from '../apiMethod';
+
+export default class AccountResource extends APIResource {
```
willclarktech: These class names should match the plural names in `index.js`.

src/api/utils.js
```diff
@@ -83,3 +23,23 @@ export const toQueryString = obj => {
...
+
+const urlParamRegex = /{[^}]+}/;
+export const solveURLParams = (url, params = {}) => {
+	if (Object.keys(params).length === 0) {
+		if (url.match(urlParamRegex)) {
+			throw new Error('URL is not completely solved');
+		}
+		return url;
+	}
+	const solvedURL = Object.keys(params).reduce(
```
willclarktech: Prefer `Object.entries` since you need both the key and the value each time.

src/api/liskApi.js
shuse2: do we not rely on `config.json`?
if we set the first object to `{ headers: {}}`, then we need to delete `headers` option from `config.options`

src/api/utils.js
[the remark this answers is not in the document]
willclarktech: Right, that was a terrible example in two respects. I guess we're unlikely to want those characters in specific API endpoints, so fine for now and if for some reason the API changes on that front then we can address it then.

src/api/apiMethod.js
```diff
@@ -0,0 +1,80 @@
...
+		if (urlParams.length > 0 && args.length < urlParams.length) {
+			return Promise.reject(
+				new Error(
+					'Arguments must include Params defined. Required: ',
+					urlParams.toString(),
+				),
+			);
+		}
+
+		const data = Object.assign(
```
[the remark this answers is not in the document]
shuse2: `request` is ambiguous since we already have `this.request`, and `urlParamsResolvers` sounds like a function.
maybe `urlData` => `urlObject` and `requestData` => `requestContent`?

test/api/resources/accounts.js
willclarktech: This should apply to `APIResource`. My issue is with the wording, which is currently ungrammatical.

test/api/resources/accounts.js
willclarktech: I can see that.

test/api/resources/accounts.js
willclarktech: Not sure, let's discuss with @Tosch110

test/api/resources/accounts.js
willclarktech: Not sure what difference `all` makes: that means that of the keys you list it should have all of them, not that the keys you list should be all the keys it has. At the very least the description of this test would have to change if it wasn't specifically about methods.

test/api/resources/transactions.js
```diff
@@ -0,0 +1,65 @@
...
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ *
+ */
+
+import APIResource from 'api/apiResource';
+import TransactionResource from 'api/resources/transactions';
+
+describe('TransactionResouce', () => {
```
willclarktech: All of these specs have a typo. Also they'll need to update to plural.

test/api/utils.js
```diff
@@ -12,109 +12,14 @@
...
-
-		it('should not include a port if not set', () => {
-			delete LSK.port;
-			result = getFullURL(LSK);
-			return result.should.equal(`${URLPrefix}://${LSK.node}`);
-		});
-	});
+const defaultBaseURL = 'http://localhost:8080';
+const path = '/resources';
+const defaultURL = `${defaultBaseURL}/api${path}`;
```
willclarktech: Prefer nesting these under the suite.

willclarktech (commented):
Again, can't request changes.

src/api/apiMethod.js
willclarktech: For me it's easier, but happy to defer to the majority if @Tosch110 agrees with you.

src/api/apiMethod.js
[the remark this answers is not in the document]
willclarktech: Let's talk it through with @Tosch110

src/api/liskApi.js
willclarktech: Oh sorry! Didn't realise there was already a `headers` object in `config.options`. Ah, then maybe we do need the nested check lower down in case someone changes that to `undefined` or something.

test/api/utils.js
shuse2: `const defaultURL = `${defaultBaseURL}/api${path}`;` is not exactly related to the test here.
What is the point of defining it separately and putting it under suite?

test/api/apiResource.js
```diff
@@ -0,0 +1,191 @@
...
+	});
+
+	describe('get headers', () => {
+		it('should return header set to liskAPI', () => {
+			return resource.headers.should.eql(defaultHeaders);
+		});
+	});
+
+	describe('get resourcePath', () => {
+		it("should return the resource's full path", () => {
```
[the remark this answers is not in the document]
shuse2: I don't think I used `"` here.
It might be prettier...

shuse2:
Additional discussion required for
- How user should handle saved nodes, defaultNodes, testnetNodes
- How user should handle headers
- How user should handle ssl, testnet

shuse2:
## Additional Refactor ideas
### Limit constructor options
- nodes: [string] (required)
- nethash: string (required)
- node: string default nodes[0]
- bannedNodes: [string] default []
- verion: string default 1.0.0
- minVersion: string default 1.0.0 (shouldn't be hyphen separated?)

### Reasons
#### Separate headers
- `version`, `minVersion` and `nethash` is the only thing should be changed for headers
#### ssl/testnets and default settings
- `ssl` and `testnet` should be indicated by `nodes` (e.g. ['https://testnet.io:443`, 'http://testnet.io:7001'])
- it is a library, so it should not depends on `config.json`, it should be default constants. (or read it from parent config file, too?)
#### Deleting randomizeNode
- When API failed, it doesn't make sense to use `providedNode` when randomizeNode is `off since providedNode it will be banned by the time choosing.
- `selectNewNode` is returning either random node(new node) or provided node(current node), which is confusing

### Add factory method (maybe)
- Add getTestnetInstance()
- Add getMainnetInstance()
which returns a new API client instance with each setting

src/api/apiMethod.js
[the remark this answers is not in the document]
Tosch110: I am totally in favour of verbosity here. The more descriptive we can get with those variable names, the better. 
If possible, I would also prefer `parameter` instead of `params`. 

I would be fine with `urlData` => `resolvedURLObject`.
`requestData` vs `requestContent` is the same for me.

src/api/apiResource.js
```diff
@@ -0,0 +1,64 @@
...
+ * Removal or modification of this copyright notice is prohibited.
+ *
+ */
+
+import * as popsicle from 'popsicle';
+
+export default class APIResource {
+	constructor(liskAPI) {
+		if (!liskAPI) {
+			throw Error('Require LiskAPI instance to be initialized.');
```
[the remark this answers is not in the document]
Tosch110: This is now part of https://github.com/LiskHQ/lisk-js/issues/588

src/api/liskApi.js
```diff
@@ -12,212 +12,179 @@
...
+		this.ssl = options.ssl !== false;
+		this.testnet = options.testnet === true;
```
Tosch110: Why usually use ` x !== false` before and `x === true` here as a check?

src/api/liskApi.js
```diff
@@ -12,212 +12,179 @@
...
+			current: this.node,
+			default: this.defaultNodes,
+			testnet: this.defaultTestnetNodes,
+		};
+	}
 
-		if (providedNethash) {
-			NetHash.nethash = providedNethash;
-			NetHash.version = '0.0.0a';
+	get currentNodes() {
```
Tosch110: Why is `currentNodes` only returning defaultNodes and not `current` ?

src/api/liskApi.js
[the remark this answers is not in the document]
Tosch110: I am in favor of having it customizable if possibe.

src/api/liskApi.js
```diff
@@ -12,212 +12,179 @@
...
 
-	broadcastTransactions(transactions) {
-		return privateApi.sendRequestPromise
-			.call(this, POST, 'transactions', transactions)
-			.then(result => result.body);
+	setTestnet(newTestnetValue) {
+		if (this.testnet !== newTestnetValue) {
+			const nonTestnetPort = this.ssl ? SSL_PORT : LIVE_PORT;
+
+			this.testnet = newTestnetValue;
```
Tosch110: afaik testnet variable is boolean

src/api/liskApi.js
```diff
@@ -12,212 +12,179 @@
...
+			return true;
+		}
+		return false;
 	}
 
-	broadcastSignatures(signatures) {
-		return privateApi.sendRequestPromise
-			.call(this, POST, 'signatures', { signatures })
-			.then(result => result.body);
+	banActiveNodeAndSelect() {
```
Tosch110: :+1:

toschdev (dismissed):

src/api/liskApi.js
shuse2: `ssl` default is true and `testnet` default is false

src/api/liskApi.js
shuse2: if you need all, there is `allNodes`, and if you need only `current` it's `node`.
This was originally `nodes` but it collides with the resource name `nodes`, therefore, it was changed to `currentNodes`

Tosch110: Okay. The name currentNodes to show defaultNodes is still confusing to me. How about currentDefaultNodes? :D

shuse2: yea I think so, too.
I think defaultNodes doesn't really make sense, it is kind of related to #588.
It shouldn't have multiple nodes stored.

src/api/apiMethod.js
willclarktech: @Tosch110 I don't understand that point about shorthand property assignment?

src/api/apiMethod.js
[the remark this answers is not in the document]
shuse2: params are often used in various projects, so i think it’s safe to use `params` much like `args`.
changed `urlData` => `resolvedURLObject`.

src/api/liskApi.js
Tosch110: I think it is a good idea!

toschdev (dismissed):

toschdev (dismissed):

toschdev (approved):

PR #513, opened 2017-12-23
Update broadcastTransaction function to match new API - Closes #483

### LINKED ISSUE #483: Update broadcast methods to match new API
The lisk-core public API will now accept an array of transactions. This should also ensure `broadcastSignedTransaction` and `broadcastSignatures` make proper use of `sendRequestPromise` and drop `Signed` from the former's name because the API accepts unsigned transactions in the case of multisignature accounts.

alepop:
- Rename `broadcastSignedTransaction` to `broadcastSignedTransactions`
- Update `requestParams`
- Update tests

close #483

### DIFF
--- a/src/api/liskApi.js
+++ b/src/api/liskApi.js
@@ -148,18 +148,18 @@ export default class LiskAPI {
 	}
 
 	/**
-	 * @method broadcastSignedTransaction
-	 * @param transaction
+	 * @method broadcastSignedTransactions
+	 * @param transactions
 	 * @param callback
 	 *
 	 * @return {Object}
 	 */
 
-	broadcastSignedTransaction(transaction, callback) {
+	broadcastSignedTransactions(transactions, callback) {
 		const request = {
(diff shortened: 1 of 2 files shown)


willclarktech:
Thanks for this PR, @alepop!
@Tosch110 What do you think about having both `broadcastSignedTransactions` and `broadcastSignedTransaction` (which just takes a single transaction, wraps it in an array and calls the former)?

Also, I believe this endpoint works for unsigned transactions too, so we may want to drop the `Signed` from the name?

Tosch110:
Yes, I think both are viable ideas. But probably beyond the scope of this issues and we can create new issues to implement this.

willclarktech:
Hi @alepop, good news and bad news:
**Good news:** your PR has helped us uncover some larger issues in the api module. Specifically that `broadcastSignedTransactions` and `broadcastSignatures` aren't calling `sendRequestPromise` with correct arguments.
**Bad news:** We'll have to close this PR as these issues should be dealt with at the same time.
Thanks for your help!

alepop:
np

PR #514, opened 2017-12-28
:seedling: Add passphrase validation functions - Closes #456

### LINKED ISSUE #456: Passphrase validation client side
Lisk-Nano right now makes use of some logic to trim whitespaces (front / end) for the mnemonic passphrases. Also checking for passphrase validation etc: https://github.com/LiskHQ/lisk-nano/blob/development/src/components/passphraseInput/index.js#L35-L71

We could have this logic in lisk-js to unify this among all the client using lisk-js

Tosch110:
Closes #456

### DIFF
--- a/src/passphrase/validation.js
+++ b/src/passphrase/validation.js
@@ -1,0 +1,60 @@
+/*
+ * Copyright © 2017 Lisk Foundation
+ *
+ * See the LICENSE file at the top-level directory of this distribution
+ * for licensing information.
+ *
+ * Unless otherwise agreed in a custom licensing agreement with the Lisk Foundation,
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ *
+ */
+import Mnemonic from 'bip39';
+
+export const countPassphraseWhitespaces = passphrase => {
+	const whitespaceMatches = passphrase.match(/\s/g);
+	return whitespaceMatches ? whitespaceMatches.length : 0;
+};
+
+export const countPassphraseWords = passphrase =>
+	passphrase.split(' ').filter(Boolean).length;
+
+export const countCapitalCharacters = passphrase => {
+	const capitalCharacterMatches = passphrase.match(/[A-Z]/g);
+	return capitalCharacterMatches ? capitalCharacterMatches.length : 0;
+};
+
+export const validatePassphrase = passphrase => {
+	const expectedWords = 12;
+	const expectedWhitespaces = 11;
+	const expectedCapitalLetters = 0;
+	const wordsInPassphrase = countPassphraseWords(passphrase);
+	const whiteSpacesInPassphrase = countPassphraseWhitespaces(passphrase);
+	const capitalCharacterInPassphrase = countCapitalCharacters(passphrase);
+
+	if (wordsInPassphrase !== expectedWords) {
+		throw new Error(
+			`Passphrase contains ${wordsInPassphrase} words instead of expected ${expectedWords}. Please check the passphrase.`,
+		);
+	}
+	if (whiteSpacesInPassphrase !== expectedWhitespaces) {
+		throw new Error(
+			`Passphrase contains ${whiteSpacesInPassphrase} whitespaces instead of expected ${expectedWhitespaces}. Please check the passphrase.`,
+		);
+	}
+	if (capitalCharacterInPassphrase !== expectedCapitalLetters) {
+		throw new Error(
+			`Passphrase contains ${capitalCharacterInPassphrase} capital character instead of expected ${expectedCapitalLetters}. Please check the passphrase.`,
+		);
+	}
+	if (!Mnemonic.validateMnemonic(passphrase)) {
+		throw new Error(
+			'Passphrase is not a valid mnemonic passphrase. Please check the passphrase.',
+		);
+	}
+
+	return true;
+};
--- a/src/passphrase/validation.js
+++ b/src/passphrase/validation.js
@@ -30,7 +30,7 @@ export const countUppercaseCharacters = passphrase => {
 export const validatePassphrase = (passphrase, wordlist) => {
 	const expectedWords = 12;
 	const expectedWhitespaces = 11;
-	const expectedUppercaseCharacter = 0;
+	const expectedUppercaseCharacterCount = 0;
 	const wordsInPassphrase = countPassphraseWords(passphrase);
 	const whiteSpacesInPassphrase = countPassphraseWhitespaces(passphrase);
 	const uppercaseCharacterInPassphrase = countUppercaseCharacters(passphrase);
@@ -41,7 +41,7 @@ export const validatePassphrase = (passphrase, wordlist) => {
 			code: 'INVALID_AMOUNT_OF_WORDS',
 			message: `Passphrase contains ${wordsInPassphrase} words instead of expected ${expectedWords}. Please check the passphrase.`,
 			expected: expectedWords,
-			has: wordsInPassphrase,
+			actual: wordsInPassphrase,
 		};
 		errors.push(passphraseWordError);
 	}
@@ -51,17 +51,17 @@ export const validatePassphrase = (passphrase, wordlist) => {
 			code: 'INVALID_AMOUNT_OF_WHITESPACES',
 			message: `Passphrase contains ${whiteSpacesInPassphrase} whitespaces instead of expected ${expectedWhitespaces}. Please check the passphrase.`,
 			expected: expectedWhitespaces,
-			has: whiteSpacesInPassphrase,
+			actual: whiteSpacesInPassphrase,
 		};
 		errors.push(whiteSpaceError);
 	}
 
-	if (uppercaseCharacterInPassphrase !== expectedUppercaseCharacter) {
+	if (uppercaseCharacterInPassphrase !== expectedUppercaseCharacterCount) {
 		const uppercaseCharacterError = {
 			code: 'INVALID_AMOUNT_OF_UPPERCASE_CHARACTER',
-			message: `Passphrase contains ${uppercaseCharacterInPassphrase} uppercase character instead of expected ${expectedUppercaseCharacter}. Please check the passphrase.`,
-			expected: expectedUppercaseCharacter,
-			has: uppercaseCharacterInPassphrase,
+			message: `Passphrase contains ${uppercaseCharacterInPassphrase} uppercase character instead of expected ${expectedUppercaseCharacterCount}. Please check the passphrase.`,
+			expected: expectedUppercaseCharacterCount,
+			actual: uppercaseCharacterInPassphrase,
 		};
 		errors.push(uppercaseCharacterError);
 	}
@@ -77,7 +77,7 @@ export const validatePassphrase = (passphrase, wordlist) => {
 			message:
 				'Passphrase is not a valid mnemonic passphrase. Please check the passphrase.',
 			expected: true,
-			has: false,
+			actual: false,
 		};
 		errors.push(validationError);
 	}
--- a/src/passphrase/validation.js
+++ b/src/passphrase/validation.js
@@ -22,7 +22,7 @@ export const countPassphraseWhitespaces = passphrase => {
 export const countPassphraseWords = passphrase =>
 	passphrase.split(' ').filter(Boolean).length;
 
-export const countCapitalCharacters = passphrase => {
+export const countUppercaseCharacters = passphrase => {
 	const capitalCharacterMatches = passphrase.match(/[A-Z]/g);
 	return capitalCharacterMatches ? capitalCharacterMatches.length : 0;
 };
@@ -30,10 +30,10 @@ export const countCapitalCharacters = passphrase => {
 export const validatePassphrase = passphrase => {
 	const expectedWords = 12;
 	const expectedWhitespaces = 11;
-	const expectedCapitalLetters = 0;
+	const expectedUppercaseCharacters = 0;
 	const wordsInPassphrase = countPassphraseWords(passphrase);
 	const whiteSpacesInPassphrase = countPassphraseWhitespaces(passphrase);
-	const capitalCharacterInPassphrase = countCapitalCharacters(passphrase);
+	const capitalCharacterInPassphrase = countUppercaseCharacters(passphrase);
 
 	if (wordsInPassphrase !== expectedWords) {
 		throw new Error(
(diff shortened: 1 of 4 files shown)


src/passphrase/validation.js
```diff
@@ -0,0 +1,60 @@
...
+
+export const countPassphraseWhitespaces = passphrase => {
+	const whitespaceMatches = passphrase.match(/\s/g);
+	return whitespaceMatches ? whitespaceMatches.length : 0;
+};
+
+export const countPassphraseWords = passphrase =>
+	passphrase.split(' ').filter(Boolean).length;
+
+export const countCapitalCharacters = passphrase => {
```
willclarktech: Prefer `Uppercase` to `Capital`

src/passphrase/validation.js
```diff
@@ -0,0 +1,60 @@
...
+	const capitalCharacterMatches = passphrase.match(/[A-Z]/g);
+	return capitalCharacterMatches ? capitalCharacterMatches.length : 0;
+};
+
+export const validatePassphrase = passphrase => {
+	const expectedWords = 12;
+	const expectedWhitespaces = 11;
+	const expectedCapitalLetters = 0;
```
willclarktech: `UppercaseCharacters`

src/passphrase/validation.js
```diff
@@ -0,0 +1,60 @@
...
+		throw new Error(
+			`Passphrase contains ${whiteSpacesInPassphrase} whitespaces instead of expected ${expectedWhitespaces}. Please check the passphrase.`,
+		);
+	}
+	if (capitalCharacterInPassphrase !== expectedCapitalLetters) {
+		throw new Error(
+			`Passphrase contains ${capitalCharacterInPassphrase} capital character instead of expected ${expectedCapitalLetters}. Please check the passphrase.`,
+		);
+	}
+	if (!Mnemonic.validateMnemonic(passphrase)) {
```
willclarktech: This check covers some of the cases above. We're adding them to have more specific messages, correct?

willclarktech: Does this cover all the language requirements we have?

test/passphrase/validation.js
```diff
@@ -0,0 +1,279 @@
...
+ */
+import {
+	countCapitalCharacters,
+	countPassphraseWhitespaces,
+	countPassphraseWords,
+	validatePassphrase,
+} from '../../src/passphrase/validation';
+
+describe('passphrase validation', () => {
+	describe('countPassphraseWhitespaces', () => {
```
willclarktech: Do we really need to export and test these simple helper functions separately to the main validation function? It seemed like they were extracted for readability rather than because they abstract away some serious logic.

test/passphrase/validation.js
```diff
@@ -0,0 +1,279 @@
...
+				countPassphraseWhitespaces(passphrase).should.be.equal(
+					expectedAmountOfWhitespaces,
+				);
+			});
+		});
+	});
+
+	describe('countPassphraseWords', () => {
+		describe('given a valid passphrase', () => {
+			const expextedAmountOfWords = 12;
```
willclarktech: Spelling

test/passphrase/validation.js
```diff
@@ -0,0 +1,279 @@
...
+		describe('given a valid passphrase', () => {
+			const passphrase =
+				'model actor shallow eight glue upper seat lobster reason label enlist bridge';
+
+			it('should return true', () => {
+				validatePassphrase(passphrase).should.be.true();
+			});
+		});
+
+		describe('given a passphrase with extra words', () => {
```
willclarktech: `too many`

test/passphrase/validation.js
```diff
@@ -0,0 +1,279 @@
...
+				'model actor shallow eight glue upper seat lobster reason label enlist bridge actor';
+			const errorMessage =
+				'Passphrase contains 13 words instead of expected 12. Please check the passphrase.';
+
+			it('should throw the error', () => {
+				validatePassphrase.bind(null, passphrase).should.throw(errorMessage);
+			});
+		});
+
+		describe('given a passphrase with less words', () => {
```
willclarktech: `too few`

test/passphrase/validation.js
```diff
@@ -0,0 +1,279 @@
...
+				'model actor shallow eight glue upper seat lobster reason label enlist';
+			const errorMessage =
+				'Passphrase contains 11 words instead of expected 12. Please check the passphrase.';
+
+			it('should throw the error', () => {
+				validatePassphrase.bind(null, passphrase).should.throw(errorMessage);
+			});
+		});
+
+		describe('given a passphrase with one more whitespace in the beginning', () => {
```
willclarktech: `extra` or `an extra`

test/passphrase/validation.js
```diff
@@ -0,0 +1,279 @@
...
+				' model actor shallow eight glue upper seat lobster reason label enlist bridge';
+			const errorMessage =
+				'Passphrase contains 12 whitespaces instead of expected 11. Please check the passphrase.';
+
+			it('should throw the error', () => {
+				validatePassphrase.bind(null, passphrase).should.throw(errorMessage);
+			});
+		});
+
+		describe('given a passphrase with one more whitespace in the end', () => {
```
willclarktech: As above

test/passphrase/validation.js
```diff
@@ -0,0 +1,279 @@
...
+				'model actor shallow eight glue upper seat lobster reason label enlist bridge ';
+			const errorMessage =
+				'Passphrase contains 12 whitespaces instead of expected 11. Please check the passphrase.';
+
+			it('should throw the error', () => {
+				validatePassphrase.bind(null, passphrase).should.throw(errorMessage);
+			});
+		});
+
+		describe('given a passphrase with more whitespaces in between', () => {
```
willclarktech: `extra whitespaces in the middle` or `in between words`

willclarktech:
In addition to the above review, after https://github.com/LiskHQ/lisk-js/pull/506, all `it` blocks should return.

src/passphrase/validation.js
Tosch110: Yes.

Tosch110: Good catch/question. Will go through it!

test/passphrase/validation.js
Tosch110: I thought better have more than less tests regarding this logic. Since it can be quite sensitive.

willclarktech: As per discussion, different clients may have different requirements for passphrase validation and these functions are exported for their convenience (and hence require unit tests).

src/passphrase/validation.js
```diff
@@ -0,0 +1,65 @@
...
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ *
+ */
+import Mnemonic from 'bip39';
+
+export const countPassphraseWhitespaces = passphrase => {
```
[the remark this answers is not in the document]
Tosch110: Yes, I will be adding more tests to cover other whitespace charakter

src/passphrase/validation.js
```diff
@@ -0,0 +1,65 @@
...
+	const whitespaceMatches = passphrase.match(/\s/g);
+	return whitespaceMatches ? whitespaceMatches.length : 0;
+};
+
+export const countPassphraseWords = passphrase =>
+	passphrase.split(' ').filter(Boolean).length;
+
+export const countUppercaseCharacters = passphrase => {
+	const uppercaseCharacterMatches = passphrase.match(/[A-Z]/g);
+	return uppercaseCharacterMatches ? uppercaseCharacterMatches.length : 0;
```
Reshetnyak: I'm not aware of your guidelines, but this function could be shorter
```javascript
// option 1
const uppercaseCharacterMatches = passphrase.match(/[A-Z]/g) || [];
return uppercaseCharacterMatches.length;
// option 2, just in one line. But not sure about readability.
export const countUppercaseCharacters = passphrase => (passphrase.match(/[A-Z]/g) || []).length;
```

The same could be applied to the function `countPassphraseWhitespaces `

willclarktech: Hi @Reshetnyak, thanks for the input, I actually almost suggested option 2 in my review (I always like a direct return from an arrow function 🙂 ). In this case I think readability/intelligibility is clearest with the longer option currently in place.

Tosch110:
@slaweet 

Here is a proposal for the error messages. I have done a lot that you requested but not all. Still, we can discuss these things before a merge.

So now the validate function returns instead of throwing errors. It has an error code, with a message including the details but also expected state vs. is state. 
I did not include at which places those (uppercase, extra whitespace) occur because I could not think of a good method to put this understandable into the error message.

test/passphrase/validation.js
```diff
@@ -229,12 +229,13 @@ describe('passphrase validation', () => {
 
 	describe('countUppercaseCharacters', () => {
 		describe('given a passphrase without uppercase character', () => {
+			const expectedAmountUppercaseCharacter = 0;
 			const passphrase =
 				'model actor shallow eight glue upper seat lobster reason label enlist bridge';
 
-			it('should return 0', () => {
+			it('should return the amount of uppercase character', () => {
```
willclarktech: `the number of uppercase characters`

test/passphrase/validation.js
```diff
@@ -263,103 +264,211 @@ describe('passphrase validation', () => {
 	});
 
 	describe('validatePassphrase', () => {
+		const passphraseIsValid = [];
```
willclarktech: This is a confusing name now.

test/passphrase/validation.js
```diff
@@ -263,103 +264,211 @@ describe('passphrase validation', () => {
...
+
 		describe('given a valid passphrase', () => {
 			const passphrase =
 				'model actor shallow eight glue upper seat lobster reason label enlist bridge';
 
-			it('should return true', () => {
-				return validatePassphrase(passphrase).should.be.true();
+			it('should return an array without the errors', () => {
```
willclarktech: `with no errors`: `without the errors` sounds like there are errors, they just shouldn't be returned.

test/passphrase/validation.js
```diff
@@ -263,103 +264,211 @@ describe('passphrase validation', () => {
...
 			const passphrase =
 				'model actor shallow eight glue upper seat lobster reason label enlist bridge actor';
-			const errorMessage =
-				'Passphrase contains 13 words instead of expected 12. Please check the passphrase.';
 
-			it('should throw the error', () => {
-				return validatePassphrase
-					.bind(null, passphrase)
-					.should.throw(errorMessage);
+			const passphraseTooManyWordsError = [
```
willclarktech: Should be plural. (Also below.)

test/passphrase/validation.js
```diff
@@ -368,22 +477,31 @@ describe('passphrase validation', () => {
 			const passphrase =
 				'model actor shallow eight glue upper seat lobster reason label enlist bridge';
 
-			it('should return true', () => {
-				return validatePassphrase(passphrase, wordlist).should.be.true();
+			it('should return the array without the error', () => {
```
willclarktech: As above

src/passphrase/validation.js
```diff
@@ -0,0 +1,86 @@
...
+			code: 'INVALID_MNEMONIC',
+			message:
+				'Passphrase is not a valid mnemonic passphrase. Please check the passphrase.',
+			expected: true,
+			actual: false,
+		};
+		errors.push(validationError);
+	}
+
+	return errors;
```
willclarktech: I still find it confusing that the return value for a success case is an empty array of errors. This seems really odd:
```
if (!validatePassphrase(passphrase).length) {
    // do something with a valid passphrase
}
```
Perhaps we need a different name, like `getPassphraseValidationErrors`?

test/passphrase/validation.js
```diff
@@ -0,0 +1,508 @@
...
+				return validatePassphrase(passphrase).should.be.eql(
+					passphraseTooManyWordsErrors,
+				);
+			});
+		});
+
+		describe('given a passphrase with too few words', () => {
+			const passphrase =
+				'model actor shallow eight glue upper seat lobster reason label enlist';
+			const passphraseTooFewWordsError = [
```
willclarktech: Plural

test/passphrase/validation.js
```diff
@@ -0,0 +1,508 @@
...
+				return validatePassphrase(passphrase).should.be.eql(
+					passphraseTooFewWordsError,
+				);
+			});
+		});
+
+		describe('given a passphrase with an extra whitespace in the beginning', () => {
+			const passphrase =
+				' model actor shallow eight glue upper seat lobster reason label enlist bridge';
+			const passphraseTooManyWhitespacesError = [
```
willclarktech: Plural

test/passphrase/validation.js
```diff
@@ -0,0 +1,508 @@
...
+				return validatePassphrase(passphrase).should.be.eql(
+					passphraseTooManyWhitespacesError,
+				);
+			});
+		});
+
+		describe('given a passphrase with an extra whitespace in the end', () => {
+			const passphrase =
+				'model actor shallow eight glue upper seat lobster reason label enlist bridge ';
+			const passphraseTooManyWhitespacesError = [
```
willclarktech: Plural

test/passphrase/validation.js
```diff
@@ -0,0 +1,508 @@
...
+				return validatePassphrase(passphrase).should.be.eql(
+					passphraseTooManyWhitespacesError,
+				);
+			});
+		});
+
+		describe('given a passphrase with too many whitespaces in between words', () => {
+			const passphrase =
+				'model actor shallow eight glue  upper seat  lobster reason label enlist bridge';
+			const passphraseTooManyWhitespacesError = [
```
willclarktech: Plural

test/passphrase/validation.js
```diff
@@ -0,0 +1,508 @@
...
+				return validatePassphrase(passphrase).should.be.eql(
+					passphraseTooManyWhitespacesError,
+				);
+			});
+		});
+
+		describe('given a passphrase with uppercase characters', () => {
+			const passphrase =
+				'modEl actor shallow eight glue upper sEat lobster reaSon label enlist bridge';
+			const passphraseWithUppercaseCharacterError = [
```
willclarktech: Plural

test/passphrase/validation.js
```diff
@@ -0,0 +1,508 @@
...
+				return validatePassphrase(passphrase).should.be.eql(
+					passphraseWithUppercaseCharacterError,
+				);
+			});
+		});
+
+		describe('given a passphrase that is an invalid mnemonic passphrase', () => {
+			const passphrase =
+				'model actor shallow eight glue upper seat lobster reason label engage bridge';
+			const passphraseInvalidMnemonic = [
```
willclarktech: `Error`

test/passphrase/validation.js
```diff
@@ -0,0 +1,508 @@
...
+					emptyErrorArray,
+				);
+			});
+		});
+
+		describe('given a passphrase that uses a different wordlist for the passphrase', () => {
+			const wordlist = Mnemonic.wordlists.spanish;
+			const passphrase =
+				'model actor shallow eight glue upper seat lobster reason label enlist bridge';
+			const passphraseInvalidMnemonic = [
```
willclarktech: `Error`

test/passphrase/validation.js
```diff
@@ -0,0 +1,589 @@
...
+				return getPassphraseValidationErrors(passphrase).should.be.eql(
+					passphraseTooFewWordsErrors,
+				);
+			});
+		});
+
+		describe('given a passphrase with an extra whitespace in the beginning', () => {
+			const passphrase =
+				' model actor shallow eight glue upper seat lobster reason label enlist bridge';
+			const passphraseTooManyWhitespacesErrorsss = [
```
willclarktech: Prefer English over Parseltongue.

PR #517, opened 2018-01-04
Add HOF for transaction creators - Closes #508

### LINKED ISSUE #508: Create a higher order function for transaction creator
All the transactions share some common logic, which is repeated in the several transaction creation functions.

It makes sense to uniform this logic in a higher order function which can be triggered by each transaction to generate the fields it needs.

willclarktech:
Closes #508 
~~This is a WIP: I thought I'd get some feedback on whether the scope of my function is correct before writing unit tests for it. (So far the existing transaction-creator unit tests have been sufficient to guarantee the refactoring does not alter transaction-creator functionality.)~~

### DIFF
--- a/src/transactions/0_transfer.js
+++ b/src/transactions/0_transfer.js
@@ -16,68 +16,51 @@
  * Transaction module provides functions for creating balance transfer transactions.
  * @class transaction
  */
-import cryptoModule from '../crypto';
 import { TRANSFER_FEE, DATA_FEE } from '../constants';
 import {
-	prepareTransaction,
 	getAddressAndPublicKeyFromRecipientData,
-	getTimeWithOffset,
+	wrapTransactionCreator,
 } from './utils';
+
+const createAsset = data => {
+	if (data && data.length > 0) {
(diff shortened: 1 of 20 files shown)


Tosch110:
Looking good. Just one thing I realised is that probably we do not need to have the `unsigned` flag when somebody creates a transaction without a passphrase. It only got so obvious with this HOF 👍

alepop:
I like this! Btw @willclarktech I've seen that you start to use a lot of functions that return another function. What about to add `curry` helper function, that allows writing something like this `const main = (fn1, fn2, args) => ...` instead of this `const main = fn1 => fn2 => args => ...` 
```
const curry = (fn, ...args) => 
  (fn.length <= args.length) ? 
    fn(...args) : 
    (...more) => curry(fn, ...args, ...more);
```
I know, you like FP ;D

willclarktech:
@alepop Haha, well with ES6 arrow functions I actually really like the `a => b => c => d` syntax, but feel free to open an issue and we can discuss the approach we want to take in general.

PR #521, opened 2018-01-11
🔥 remove yuidoc

### LINKED ISSUE #511: Remove yuidoc
We are not using yuidoc and it adds a lot of unnecessary boilerplate onto the files.

Slamper:
Closes #511


Slamper:
For me the tests failed and it refused to push without the second commit.

willclarktech:
@Slamper Which version of node are you using?

Slamper:
@willclarktech v9.2.1

willclarktech:
@Slamper Please use node v6. Those tests will break in later versions of node due to a difference in how `Buffer.from` treats invalid hex strings.

Sorry, this will be properly documented once we fully apply the standards set in https://github.com/LiskHQ/lisk-template

Slamper:
@willclarktech Ok thank you for the information.
I simply removed the test commit 👍

.gitignore
```diff
@@ -49,9 +49,6 @@ jspm_packages
 # IDE directories
 .idea
 
-# Docs
-yuidoc
```
willclarktech: 👌 Thanks for being thorough!

src/api/liskApi.js
```diff
@@ -12,43 +12,11 @@
...
- *         testnet: true,
- *         port: '7000',
- *         bannedNodes: [],
- *         nodes: [],
- *         nethash: ''
- *     };
- *
- *
- *     var LSK = lisk.api(options);
-*/
```
willclarktech: This isn't actually a yuidoc comment, but I'm happy with it being removed as we will have documentation on the documentation site. What do you think @Tosch110 ?

Tosch110: Yes, agree! It is actually there for yuidoc, it would show it as an example also in yuidoc.

willclarktech:
Thanks @Slamper!

PR #534, opened 2018-01-26
Redial stop on POST or PUT - Closes #533

### LINKED ISSUE #533: Stop redial when using POST / PUT request
**Current behavior**
In rare cases, when the node is available but does not respond, it was possible that the transaction got transmitted to the first node, while the library would create the transaction and look for a new node to submit it to, eventually both ending up with a transaction and therefore replayed the transaction to two different nodes with two different Id's

**Expected behavior**

The transaction should never be submitted twice. On POST / PUT request, a new node should not be discovered but the user should be informed that the request failed and would be adviced to try again with a new node.

**Additional**

Writing tests to confirm the described behavior.

Tosch110:
Closes #533


lib/api/liskApi.js
```diff
@@ -335,6 +335,17 @@ function handleTimestampIsInFutureFailures (requestType, options, result) {
 
 function handleSendRequestFailures (requestType, options, error) {
 	var that = this;
+	var type = this.checkRequest(requestType, options);
+
+	if(type === 'POST' || type === 'PUT') {
```
willclarktech: Prefer whitelisting `GET` rather than blacklisting.

lib/api/liskApi.js
```diff
@@ -335,6 +335,17 @@ function handleTimestampIsInFutureFailures (requestType, options, result) {
...
+		return Promise.resolve({
+			success: false,
+			error: error,
+			errorCode: 'PEER_TIMEOUT',
+			message: 'could not create http request. Please try again.'
```
willclarktech: `Could`

PR #546, opened 2018-02-01
Add lisk address validation function - Closes #531

### LINKED ISSUE #531: Add function to check Lisk address
To check address with `L` in the end - throw when it does not exist, check the amount of characters.

Tosch110:
Closes #531


test/transactions/utils/index.js
```diff
@@ -13,19 +13,35 @@
...
+	prependMinusToPublicKeys,
+	prependPlusToPublicKeys,
 	signTransaction,
 	multiSignTransaction,
 	verifyTransaction,
+	validateAddress,
+	validateKeysgroup,
+	validatePublicKey,
+	validatePublicKeys,
 } from '../../../src/transactions/utils';
```
willclarktech: Nice, but still missing `signRawTransaction` and `wrapTransactionCreator`.

src/transactions/utils/validation.js
```diff
@@ -47,3 +47,19 @@ export const validateKeysgroup = keysgroup => {
 	}
 	return validatePublicKeys(keysgroup);
 };
+
+// Specification: https://github.com/LiskHQ/lisk/blob/development/logic/account.js#L373
+export const validateAddress = address => {
+	if (address.length < 2 || address.length > 22) {
```
willclarktech: Not every address with length 22 is valid. A thorough check would see if it's in the valid range using bignum.

test/transactions/utils/validation.js
```diff
@@ -186,4 +187,44 @@ describe('public key validation', () => {
...
+	describe('#validateAddress', () => {
+		describe('Given a valid address', () => {
+			const address = '13133549779353512613L';
+
+			it('should return true if valid', () => {
+				return validateAddress(address).should.be.equal(true);
+			});
+		});
+
+		describe('Given a too short address', () => {
```
willclarktech: `Given an address that is too short`

test/transactions/utils/validation.js
```diff
@@ -186,4 +187,44 @@ describe('public key validation', () => {
...
+			const address = 'L';
+			const error =
+				'Address length does not match requirements. Expected between 2 and 22 characters.';
+
+			it('should throw if invalid', () => {
+				return validateAddress.bind(null, address).should.throw(error);
+			});
+		});
+
+		describe('Given a too long address', () => {
```
willclarktech: As above

test/transactions/utils/validation.js
```diff
@@ -186,4 +187,44 @@ describe('public key validation', () => {
 			});
 		});
 	});
+
+	describe('#validateAddress', () => {
+		describe('Given a valid address', () => {
+			const address = '13133549779353512613L';
+
+			it('should return true if valid', () => {
```
willclarktech: `if valid` is covered by the `describe` block so can be removed.

test/transactions/utils/validation.js
```diff
@@ -186,4 +187,44 @@ describe('public key validation', () => {
...
+				return validateAddress(address).should.be.equal(true);
+			});
+		});
+
+		describe('Given a too short address', () => {
+			const address = 'L';
+			const error =
+				'Address length does not match requirements. Expected between 2 and 22 characters.';
+
+			it('should throw if invalid', () => {
```
willclarktech: As above.

src/transactions/utils/validation.js
```diff
@@ -47,3 +47,19 @@ export const validateKeysgroup = keysgroup => {
 	}
 	return validatePublicKeys(keysgroup);
 };
+
+// Specification: https://github.com/LiskHQ/lisk/blob/development/logic/account.js#L373
```
willclarktech: `development` is not a stable reference

src/transactions/utils/validation.js
```diff
@@ -47,3 +47,19 @@ export const validateKeysgroup = keysgroup => {
 	}
 	return validatePublicKeys(keysgroup);
 };
+
+// Specification: https://github.com/LiskHQ/lisk/blob/1.0.0/logic/account.js#L373
```
willclarktech: `1.0.0` is an even less stable reference than `development`. If we're going to refer to a code line it should be via a commit. Preferably we'd have a documentation site though.

src/transactions/utils/validation.js
```diff
@@ -47,3 +47,19 @@ export const validateKeysgroup = keysgroup => {
...
+export const validateAddress = address => {
+	if (address.length < 2 || address.length > 22) {
```
willclarktech: This check still doesn't cover the correct numerical range.

test/transactions/utils/validation.js
```diff
@@ -186,4 +187,44 @@ describe('public key validation', () => {
 			});
 		});
 	});
+
+	describe('#validateAddress', () => {
+		describe('Given a valid address', () => {
+			const address = '13133549779353512613L';
+
+			it('should return true', () => {
+				return validateAddress(address).should.be.equal(true);
```
willclarktech: Prefer `.should.be.true()`

src/transactions/utils/validation.js
Tosch110: Okay, removing this entirely

src/transactions/utils/validation.js
```diff
@@ -61,5 +61,17 @@ export const validateAddress = address => {
 			'Address format does not match requirements. Expected "L" at the end.',
 		);
 	}
+
+	const addressAsBignum = bignum(address.slice(0, -1));
+	const biggestAddress = bignum
+		.fromBuffer(Buffer.from([255, 255, 255, 255, 255, 255, 255, 255]))
+		.plus(1);
```
willclarktech: Why `plus(1)`?

src/transactions/utils/validation.js
```diff
@@ -61,5 +61,17 @@ export const validateAddress = address => {
 			'Address format does not match requirements. Expected "L" at the end.',
 		);
 	}
+
+	const addressAsBignum = bignum(address.slice(0, -1));
+	const biggestAddress = bignum
+		.fromBuffer(Buffer.from([255, 255, 255, 255, 255, 255, 255, 255]))
```
willclarktech: How about `new Array(8).fill(255)` so we don't have to count?

src/transactions/utils/validation.js
```diff
@@ -61,5 +61,17 @@ export const validateAddress = address => {
...
+		.plus(1);
+
+	if (addressAsBignum.gt(biggestAddress)) {
+		throw new Error(
+			'Address length does not match requirements. Address out of maximum range.',
```
willclarktech: It's not really about length in this case.

src/transactions/utils/validation.js
Tosch110: Shouldnt the biggest address still be a possible address?

src/transactions/utils/validation.js
```diff
@@ -63,13 +63,11 @@ export const validateAddress = address => {
 	}
 
 	const addressAsBignum = bignum(address.slice(0, -1));
-	const biggestAddress = bignum
-		.fromBuffer(Buffer.from([255, 255, 255, 255, 255, 255, 255, 255]))
-		.plus(1);
+	const biggestAddress = bignum.fromBuffer(Buffer.from(new Array(8).fill(255)));
```
willclarktech: How about defining this in `constants`?

Tosch110: Yeah, I also had that thought. Lets do that then.

src/constants/index.js
```diff
@@ -34,3 +34,6 @@ export const SSL_PORT = '443';
 
 export const GET = 'GET';
 export const POST = 'POST';
+
+// biggesr address derived from bignum.fromBuffer(Buffer.from(new Array(8).fill(255)))
```
willclarktech: "Largest possible address. Derived from `bignum.fromBuffer(Buffer.from(new Array(8).fill(255)))`."

Issue #417, opened 2017-10-12
Implement strict type system


Tosch110:
Typechecking will make the implementation more secure.
e.g. https://flow.org

Aimed for Lisk-JS 1.1

----
Updates:
A decision has finally been made for TypeScript.
Aimed for Lisk-Elements 2.0

### Child Issues
- [x] Set up TypeScript for lisk-elements #781 
- [x] Migrate @liskhq/lisk-cryptography package to TypeScript #797 
- [x] Migrate @liskhq/lisk-passphrase package to TypeScript #798
- [x] Migrate @liskhq/lisk-api-client package to TypeScript #799
- [x] Migrate @liskhq/lisk-transactions package to TypeScript #800
- [x] Migrate @liskhq/lisk-client package to TypeScript #801
- [x] Migrate lisk-elements package to TypeScript #802

alepop:
I propose to consider Typescript instead of Flow

alepop:
My points to use **Typescript**:
- Typescript is a compiler with support for new Javascript features (`es6/7/8, async/await, decorators`)
- `React` support out of the box
- This allows you get rid of `babel`
- Easy setup (`webpack + ts-loader + tsconfig.json file`)
- You can move your project to Typescript incrementally with one field in config file

kazzkiq:
In addition to @alepop, I would like to take some points where TypeScript shines over Flow:

- **Typescript** is a superset of JavaScript, which means it not only provides type-checking, but it actually *adds* other features on top of it, without losing compatibility with JavaScript code. This means that features not seen in JavaScript such as Interfaces, Enums, Classes (out of the box), Decorators and others are all supported by default, but you only use them if you want to. It also have ES6/7 support out of the box, compiling to ES6 (and bellow) if needed.

- **Flow** is a great tool in what is was made for, that being the possibility of adding a tiny "typecheck" layer on top of your JavaScript code while leaving everything else up to you (afaik).

TypeScript on the other hand is at another level, it managed to build a terse language set around JavaScript itself, and extend JavaScript to the point it can compete with mature languages, while maintaining the dynamic features of the JS ecossystem.

The work Microsoft made in TypeScript was surprisingly good (not usual if you take into account their historical open-souce decisions), and by assigning personalities like Anders Hejlsberg to work on it only shows its commitment to the language.

[...]

PR #549, opened 2018-02-02
Added sample flow setup - Discussion For #417

shuse2:
This is a sample setup regards to Implement strict type system #417.
It shows flow can easily setup, and partially included and gradually complete the improvement.

-----

Set up flow for start using without affecting the current code too much.
What I did
- Install Flow and set basic config
- Move `lib` to `src`, and within src, `flow` tags are used (changed import path too)
- For build purpose, `index.js` is also moved to inside of `src`
- Changed gruntfile to use babel (might be good idea to get rid of grunt)

-----

Also, `npm` publish flow needs to be discussed.
If we should keep the `lib` in repo, or build in CI and directly publish.


shuse2:
Closing due to wrong branch root

PR #554, opened 2018-02-08
Get transaction bytes expand testsuite - Closes #542

### LINKED ISSUE #542: Expand testsuite for getTransactionBytes
Child Issue for https://github.com/LiskHQ/lisk-js/issues/502

Tosch110:
Closes #542


test/transactions/utils/signAndVerify.js
```diff
@@ -250,3 +250,1410 @@ describe('signAndVerify transaction utils', () => {
...
+
+		describe('#signTransaction', () => {
+			describe('given a passphrase and a second passphrase', () => {
+				const passphrase =
+					'wagon stock borrow episode laundry kitten salute link globe zero feed marble';
+				const secondPassphrase =
+					'trouble float modify long valve group ozone possible remove dirt bicycle riot';
+
+				describe('when tested on the first signature', () => {
+					it('should create the same signature', done => {
```
shuse2: This block seems to be synchronised. 
Do we need `done`?

test/transactions/utils/signAndVerify.js
```diff
@@ -250,3 +250,1410 @@ describe('signAndVerify transaction utils', () => {
...
+							delete rawTx.signature;
+							delete rawTx.signSignature;
+							signTransaction(rawTx, passphrase).should.be.equal(signature);
+						});
+						done();
+					});
+				});
+
+				describe('when tested on the second signature', () => {
+					it('should create the same signature', done => {
```
shuse2: Same as above

test/transactions/utils/signAndVerify.js
```diff
@@ -250,3 +250,1410 @@ describe('signAndVerify transaction utils', () => {
...
+					});
+				});
+			});
+		});
+
+		describe('#verifyTransaction', () => {
+			describe('when executed', () => {
+				const secondPublicKey =
+					'f9666bfed9ef2ff52a04408f22f2bfffaa81384c9433463697330224f10032a4';
+				it('should verify all the transactions', done => {
```
shuse2: same as above

src/transactions/utils/getTransactionBytes.js
```diff
@@ -150,43 +150,48 @@ export const checkTransaction = transaction => {
...
+	const {
+		type,
+		timestamp,
+		requesterPublicKey,
+		senderPublicKey,
+		recipientId,
+		amount,
+		signature,
+		signSignature,
+	} = transaction;
```
shuse2: Do we prefer this syntax?

willclarktech: I like it, but maybe it makes more sense for `checkTransaction(transaction)` to come before the destructuring?

test/transactions/utils/signAndVerify.js
```diff
@@ -250,3 +250,1410 @@ describe('signAndVerify transaction utils', () => {
 		});
 	});
 });
+
+describe('integration sign and verify', () => {
+	describe('given a set of transactions', () => {
+		let transactions;
+		beforeEach(() => {
+			transactions = [
```
willclarktech: Let's move these into a separate fixtures file (could even be JSON?).

test/transactions/utils/signAndVerify.js
```diff
@@ -250,3 +250,1410 @@ describe('signAndVerify transaction utils', () => {
 		});
 	});
 });
+
+describe('integration sign and verify', () => {
+	describe('given a set of transactions', () => {
+		let transactions;
```
willclarktech: Maybe rename these `validTransactions`?

test/transactions/utils/signAndVerify.js
willclarktech: A comment about how these were generated would be nice.

test/transactions/utils/signAndVerify.js
willclarktech: @shuse2 Yes, we're using this ESLint rule: https://github.com/lo1tuma/eslint-plugin-mocha/blob/master/docs/rules/no-synchronous-tests.md

The alternative is to return from the `forEach` call though. That would be my preference since we don't use callbacks anywhere else.

test/transactions/utils/signAndVerify.js
willclarktech: Response as above.

test/transactions/utils/signAndVerify.js
willclarktech: `should create the correct signature` or `should recreate the original signature`?

test/transactions/utils/signAndVerify.js
```diff
@@ -250,3 +250,1410 @@ describe('signAndVerify transaction utils', () => {
...
+				const passphrase =
+					'wagon stock borrow episode laundry kitten salute link globe zero feed marble';
+				const secondPassphrase =
+					'trouble float modify long valve group ozone possible remove dirt bicycle riot';
+
+				describe('when tested on the first signature', () => {
+					it('should create the same signature', done => {
+						transactions.forEach(transaction => {
+							const { signature } = transaction;
+							const rawTx = Object.assign({}, transaction);
```
willclarktech: This is fine for signatures, but we should be careful if we extend this test pattern to assets because the nested objects will have the same reference.

test/transactions/utils/signAndVerify.js
```diff
@@ -250,3 +250,1410 @@ describe('signAndVerify transaction utils', () => {
...
+		});
+
+		describe('#verifyTransaction', () => {
+			describe('when executed', () => {
+				const secondPublicKey =
+					'f9666bfed9ef2ff52a04408f22f2bfffaa81384c9433463697330224f10032a4';
+				it('should verify all the transactions', done => {
+					transactions.forEach(transaction => {
+						if (!transaction.signSignature) {
+							verifyTransaction(transaction).should.be.equal(true);
```
willclarktech: Prefer `should.be.true()` syntax.

Tosch110:
I have used the sign / verify function on purpose because it uses all the steps in order to test the whole fall through of the functions - therefore integration test. 

The sign function starts off by using getTransactionBytes, then getTransactionHash, then sign the transaction. Same for the verify function. The transactions have been created by the software before we made the changes (refactoring) and are recreated here. Therefore we can confirm the behaviour has not changed.
All transaction types are part of the tests and each type has multiple transactions.

test/transactions/utils/transactions.json
```diff
@@ -0,0 +1 @@
...4e639bda406f6ed8810d75a7b4066ad5e90ef3c2030927b917e8492db59ffad017ce51878b7217b27d3a506","id":"7976119586785833934"},{"type":6,"amount":"881374","fee":"10000000","recipientId":null,"senderPublicKey":"c094ebee7ec0c50ebee32918655e089f6e1a604b83bcaa760293c61e0f18ab6f","timestamp":54196080,"asset":{"inTransfer":{"dappId":"1000000"}},"signature":"f9facd4dd0b278e3fdca44e6b5527ff4d71f65fb05d1a993cc094f7a6e08118d96ad9103c7577c9efff41d3deec08ffb06346ab27aeff800d499785b53bea40d","signSignature":"ddbe2671cb80cf478b581772242e20d49b8f88445928f8c685ea06a83733a7c7abec31c53f52f1962126abc6e1e03b2c8c7b881f6a3d6808496eb500498b730b","id":"10748675444711410396"},{"type":7,"amount":"370999","fee":"10000000","recipientId":"1859190791819301L","senderPublicKey":"c094ebee7ec0c50ebee32918655e089f6e1a604b83bcaa760293c61e0f18ab6f","timestamp":54196080,"asset":{"outTransfer":{"dappId":"763580659","transactionId":"89998498"}},"signature":"8041323c405e23fc8ebaaf7ded1a61e1922c29643d6ab5cf536655fb69380d7e64a158198a21cd34faa06ed1777321f403421fe59fc1b4c40087c19ab47eda0c","signSignature":"177664c1e4afc88462f44f8af6192d36d435bf9457d2466e9ced8b52d44eb740be03232a6a9e0467762cf336bbd155fba04dd0a74cb7a6938ee976453bf6070b"}]
```
willclarktech: Let's prettify this so we can debug it more easily.

test/transactions/utils/signAndVerify.js
```diff
@@ -250,3 +252,65 @@ describe('signAndVerify transaction utils', () => {
...
+				it('should verify all the transactions', () => {
+					return validTransactions.forEach(transaction => {
+						if (!transaction.signSignature) {
+							return verifyTransaction(transaction).should.be.true();
+						}
+
+						return verifyTransaction(
+							transaction,
+							secondPublicKey,
+						).should.be.equal(true);
```
willclarktech: Prefer `should.be.true()` syntax.

.eslintignore
```diff
@@ -1,6 +1,5 @@
 coverage/
 dist-*/
 docs/
-browsertest/browsertest.js
-browsertest/browsertest.min.js
+browsertest/
```
willclarktech: This excludes setup files etc.?

Tosch110: Yes, but do we need it for 2 lines of code? The current setup did not let me pass CI

PR #558, opened 2018-02-09
Apply template - Closes #568,#569,#570,#571,#572,#573,#575

### LINKED ISSUE #552: Apply lisk-template
We should apply the `lisk-template` repository created by @LiskHQ/standards to this repository. This will align it with the newly defined standards, patterns and workflows that will be applied to all new and existing repositories from now on.

### Todo

- [x] Update .nvmrc (#568)
- [x] Update .npmrc file (#569)
- [x] Standardize .gitignore file (#570)
- [x] Standardize package.json field (#571)
- [x] Standardize npm scripts  (#572)
- [x] Update Jenkinsfiles to standards (#573)
- [x] Update/Standardize mocha/chai/sinon setup (e.g. sandbox) (#574)
- [x] Update/Standardize package manager (#575)
- [x] Be consistent about use of `should` vs `expect` (#576)
- [x] underscore_separated file/directory names (#577)
- [x] Update ESLint (#578)
- [x] Husky and lint-staged

### LINKED ISSUE #568: Update .nvmrc
Update .nvmrc to match standard

### LINKED ISSUE #569: Update .npmrc file
Update .npmrc file to match standard

### LINKED ISSUE #570: Standardize .gitignore file
Standardize .gitignore file

### LINKED ISSUE #571: Standardize package.json field
Standardize package.json field

### LINKED ISSUE #572: Standardize npm scripts
Standardize npm scripts

### LINKED ISSUE #573: Update Jenkinsfiles to standards
Update Jenkinsfiles to standards

### LINKED ISSUE #575: Update/Standardize package manager
Update/Standardize package manager

shuse2:
Closes #568
Closes #569 
Closes #570 
Closes #571 
Closes #572 
Closes #573 
Closes #575


package.json
```diff
@@ -2,34 +2,64 @@
...
+	"homepage": "https://github.com/LiskHQ/lisk-js#readme",
+	"repository": {
+		"type": "git",
+		"url": "git+https://github.com/LiskHQ/lisk-js.git"
+	},
+	"bugs": {
+		"url": "https://github.com/LiskHQ/lisk-js/issues"
+	},
+	"engines": {
+		"node": "=6.12.3",
```
AndrewBarba: I have a feeling this would break yarn and maybe npm for any user trying to install this on a newer version of node or npm. Lisk-js should not be required to run on this exact version of node. `>=6` I think is more appropriate

package.json
```diff
@@ -2,34 +2,64 @@
...
+		"npm": "=3.10.10"
```
AndrewBarba: Same comment as above

shuse2:
- [X] Update .nvmrc
- [X] Update .npmrc file
- [X] Standardize .gitignore file
- [X] Standardize package.json field
- [X] Standardize npm scripts
- [X] Update Jenkinsfiles to standards
- [] Update/Standardize mocha/chai/sinon setup (e.g. sandbox)
- [] Update/Standardize package manager
- [] Be consistent about use of should vs expect
- [] underscore_separated file/directory names
- [X] Update ESLint
- [] Husky and lint-staged

package.json
shuse2: Currently `lisk` core only supports particular version of node.
That is why it is freeze at `6.12.3`

package.json
shuse2: Same comment as above

package.json
AndrewBarba: But this is not lisk core... the js sdk should be able to run on any version of node or the browser that supports the necessary syntax

AndrewBarba: @vitaly-t

.eslintrc.json
```diff
@@ -1,5 +1,6 @@
 {
 	"extends": ["lisk-base"],
+	"plugins": ["mocha"],
```
willclarktech: This should be in `test/.eslintrc.json`.

package.json
willclarktech: Breaks on yarn `v1.3.2` but works fine on npm `v3.10.10` (which ships with Node.js `v6.12.3`). @Tosch110 thoughts?

.gitignore
```diff
@@ -58,3 +59,7 @@ dist-node
 browsertest/package.json
 browsertest/browsertest.js
 browsertest/browsertest.min.js
+
+# lock files (required version is npm below 3)
```
willclarktech: Do we really want to add these? We want to be discouraging people from creating them in the first place.

package.json
```diff
@@ -2,34 +2,62 @@
 	"name": "lisk-js",
 	"version": "0.4.5",
 	"description": "JavaScript library for sending Lisk transactions from the client or server",
-	"main": "./dist-node/index.js",
+	"author": "Lisk Foundation <admin@lisk.io>, lightcurve GmbH <admin@lightcurve.io>",
+	"license": "GPL-3.0",
+	"keywords": [
+		"lisk",
+		"blockchain"
```
willclarktech: We might want to extend this for LiskJS specifically.

Jenkinsfile
```diff
@@ -50,9 +58,17 @@ pipeline {
...
 		success {
 			deleteDir()
+			githubNotify context: 'continuous-integration/jenkins/lisk-js', description: 'The build passed.', status: 'SUCCESS'
 		}
 		failure {
 			archiveArtifacts allowEmptyArchive: true, artifacts: 'cypress/screenshots/'
+			githubNotify context: 'continuous-integration/jenkins/lisk-js', description: 'The build failed.', status: 'FAILURE'
+		}
+		aborted {
+			githubNotify context: 'continuous-integration/jenkins/lisk-template', description: 'The build was aborted.', status: 'ERROR'
```
willclarktech: `lisk-js` not `lisk-template`

package.json
```diff
@@ -2,34 +2,64 @@
...
 		"start": "./bin/start.sh",
-		"test": "nyc mocha test --recursive",
-		"test:watch": "chokidar \"./(src|test)/**/*.js\" -c \"mocha test --recursive\" --initial",
+		"clean": "rm -rf ./node_modules ./dist-browser/lisk-js.js ./dist-browser/lisk-js.min.js ./dist-node",
+		"clean:browsertest": "rm -r browsertest/src browsertest/test",
+		"format": "prettier --write \"browsertest/{setup,runTests}.js\" \"{cypress/integration,src,test}/**/*.js\"",
+		"lint": "eslint .",
+		"lint:fix": "npm run lint -- --fix",
+		"test": "NODE_ENV=test nyc mocha test",
+		"test:watch": "npm test -- --watch",
```
willclarktech: We were using chokidar because js-nacl broke mocha's watch functionality. This might have been fixed along the way with version upgrades, but we need to confirm. Does this still work ok?

package.json
```diff
@@ -2,34 +2,64 @@
...
-		"eslint": "grunt eslint",
-		"prettier": "prettier --write \"browsertest/{setup,runTests}.js\" \"{cypress/integration,src,test}/**/*.js\"",
+		"cover": "if [ -z $JENKINS_HOME ]; then npm run cover:local; else npm run cover:ci; fi",
+		"cover:base": "NODE_ENV=test nyc report",
+		"cover:local": "npm run cover:base -- --reporter=html --reporter=text",
+		"cover:ci": "npm run cover:base -- --reporter=text-lcov | coveralls -v",
 		"serve:browsertest": "http-server browsertest",
-		"jenkins": "grunt jenkins --verbose",
-		"build": "grunt build",
+		"prebuild:node": "rm -r dist-node/* || mkdir dist-node | echo",
```
willclarktech: Should there also be a `prebuild:browser` script? I think it got lost from the `Gruntfile.js`.

package.json
```diff
@@ -2,34 +2,64 @@
...
+		"prebuild:browsertest": "cp package.json browsertest/",
+		"build": "npm run build:node && npm run build:browser",
+		"build:browsertest":
+			"npm run prebuild:browsertest && npm run babel:browsertest && npm run browserify:browsertest && npm run uglify:browsertest && npm run clean:browsertest",
+		"build:node": "npm run prebuild:node && npm run babel",
+		"build:browser": "npm run browserify && npm run uglify",
```
willclarktech: The problem with this is as a standalone script is it assumes the node build has already happened.

package.json
```diff
@@ -2,34 +2,64 @@
...
+		"cover:local": "npm run cover:base -- --reporter=html --reporter=text",
+		"cover:ci": "npm run cover:base -- --reporter=text-lcov | coveralls -v",
 		"serve:browsertest": "http-server browsertest",
-		"jenkins": "grunt jenkins --verbose",
-		"build": "grunt build",
+		"prebuild:node": "rm -r dist-node/* || mkdir dist-node | echo",
+		"prebuild:browsertest": "cp package.json browsertest/",
+		"build": "npm run build:node && npm run build:browser",
+		"build:browsertest":
+			"npm run prebuild:browsertest && npm run babel:browsertest && npm run browserify:browsertest && npm run uglify:browsertest && npm run clean:browsertest",
```
willclarktech: `prebuild:browsertest` runs automatically before this script is run so does not need to be specified here.

package.json
```diff
@@ -2,34 +2,64 @@
...
+		"npm": "=3.10.10"
+	},
+	"main": "dist-node/index.js",
 	"scripts": {
 		"prestart": "./bin/prestart.sh",
 		"start": "./bin/start.sh",
-		"test": "nyc mocha test --recursive",
-		"test:watch": "chokidar \"./(src|test)/**/*.js\" -c \"mocha test --recursive\" --initial",
+		"clean": "rm -rf ./node_modules ./dist-browser/lisk-js.js ./dist-browser/lisk-js.min.js ./dist-node",
+		"clean:browsertest": "rm -r browsertest/src browsertest/test",
```
willclarktech: Would this make sense as a `postbuild:browsertest` script?

test/mocha.opts
```diff
@@ -1,3 +1,4 @@
---timeout 10000
 --require ./test/setup.js
---compilers js:babel-register
```
willclarktech: We might still need this in case we want to run mocha on the source code without nyc.

package.json
shuse2: As i reverted this should be reverted too.
As long as `js-nacl` only applied once to the global, mocha watch works fine.
I will contact add that change here

.gitignore
shuse2: if user accidentally use npm >= 5, it will automatically adds lock file.
Therefore, I thought it will be good to added to ignorefile until we can support higher version

willclarktech: One thought is whether having to fight with git to keep lockfiles around might encourage people to use the package manager we support to avoid that struggle?

package.json
shuse2: `build` command can be standalone, but `browserify` and `unglify` expects to be build before that, and each of them is not very useful as standalone.

.gitignore
shuse2: wouldn't it increase the PR discussion every time someone commits the lock file?

package.json
willclarktech: Ah great, thank you!

.gitignore
willclarktech: Probably. Let's add the here then, but text should be more like `Supported package manager is npm 3`.

package.json
willclarktech: Actually, now I see the alternative, I'm not sure it's better. :/

package.json
```diff
@@ -2,34 +2,69 @@
...
+		"build:browser": "npm run babel && npm run browserify && npm run uglify",
 		"build:check": "node -e \"require('./dist-node')\"",
-		"build:browsertest": "grunt build-browsertest",
-		"precommit": "lint-staged && npm run eslint",
-		"prepush": "npm run eslint && npm test",
-		"prepublishOnly": "npm run prepush && npm run build",
+		"babel": "babel src -d dist-node",
+		"babel:browsertest":
+			"babel src -d ./browsertest/src && BABEL_ENV=browsertest babel test --ignore test/transactions/dapp.js -d ./browsertest/test",
+		"postbuild:browsertest": "rm -r browsertest/src browsertest/test",
```
willclarktech: Would be great to have a slightly more logical/consistent order for these scripts.

package.json
AndrewBarba: Node LTS is v8. This is just incorrect usage of the engines dict, needs to be fixed. No sdk like this would be limited to such a specific version of node AND package manager. His reasoning was that Lisk core uses v6 but I have a swift sdk for Lisk (matches this api) so I’m only supposed to allow people to use my swift sdk if they are on node v6? It just makes no sense, no one will be able to install this with these requirements. Heck electron is on a much later version of node, nano will probably fail to install this as a dependency

shuse2: Yes I understand the point. However, due to the fact that, there was change in Buffer behaviour, 
lisk-js also supports only v6 now.

I created the follow up issue #561 for this.

willclarktech: @AndrewBarba I'm afraid I don't really understand your points about a swift SDK, maybe there's some context I'm missing? Regarding the current proposal for these fields...

From the npm docs:
> You can specify the version of node that your stuff works on:
{ "engines" : { "node" : ">=0.10.3 <0.12" } }

We aren't currently testing multiple versions of node with LiskJS, which is why we're planning to put only one version of node in there. I'm definitely open to the idea that we should allow all node 6 versions. However, there's at least one breaking change in node 8 which breaks LiskJS functionality. So according to the npm documentation—as far as I can see—we definitely shouldn't put a range in engines which allowed the use of node 8.

Do you have any resources which advocate a different way of using the engines field?

yarn is not our recommended package manager, but it does offer the `--ignore-engines` option if you want to ignore our recommendation as well as the fact that we only test LiskJS with one version of node and that's the only version we currently support.

AndrewBarba: The original answer to my concern was that lisk core uses Node v6. My point with the swift sdk was that an SDK using the core API to spec has nothing to do with the software versions running on core. Thats the whole point of an api, i call according to spec and I dont need to know what the core machine is using. But now the answer has changed, that apparently lisk js breaks on Node v8 because of a buffer issue. That is completely different (and valid) reason to use the engines options, but the original reason was incorrect usage of the engines options. I currently use lisk-js 0.5.1 on Node v8 extensively, and also on AWS lambda which uses Node 6.10. I bet a lot of others do this as well. So if tests legitimately do not pass on Node v8, then fine, limit to a range of Node 6. But if you leave it as is, very few people are going to be able to install and use this as they currently are

AndrewBarba: Additionally, this is the main dependency in Lisk Nano which is built on Electron, which uses Node v8. So how could you not support Node v8?

test/.eslintrc.json
```diff
@@ -8,7 +9,8 @@
 		"sinon": true
 	},
 	"rules": {
-		"arrow-body-style": "off"
+		"arrow-body-style": "off",
+		"no-unused-expressions": "off"
```
Tosch110: why turn this off?

docs/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,16 @@
...
+
+### Review checklist
+
+* The PR solves #INSERT_ISSUE_NUMBER
+* All new code is covered with unit tests
+* All new code was formatted with Prettier
+* Linting passes
+* Tests pass
+* Commit messages follow the
+	[commit guidelines](CONTRIBUTING.md#git-commit-messages)
```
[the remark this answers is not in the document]
shuse2: It seems like it's working by looking at the `view` on this github

package.json
```diff
@@ -2,35 +2,68 @@
...
+		"node": "=6.12.3",
+		"npm": "=3.10.10"
+	},
+	"main": "dist-node/index.js",
 	"scripts": {
 		"prestart": "./bin/prestart.sh",
 		"start": "./bin/start.sh",
-		"test": "nyc mocha test --recursive",
-		"test:watch": "chokidar \"./(src|test)/**/*.js\" -c \"mocha test --recursive\" --initial",
+		"clean": "rm -rf ./node_modules ./dist-browser/lisk-js.js ./dist-browser/lisk-js.min.js ./dist-node",
```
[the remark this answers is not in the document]
shuse2: this was not meant to be included in the commit.
we had another command to clean up before building, but that became another script, and this got left.

However I think it's useful to have it because when changing the branch, it doesnt rsync the node_module folder.
Therefore, if you change the dependencies (especially delete), you should test it with clean install, thats why I prefer to have command to clean every non-managed stuff.

Jenkinsfile
```diff
@@ -17,27 +17,35 @@ pipeline {
 	stages {
 		stage('Install dependencies') {
 			steps {
-				sh '''
-				cp -r ~/cache/${CHANGE_TARGET:-${BRANCH_NAME:-development}}/node_modules ./ || true
```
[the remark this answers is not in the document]
shuse2: Following up the template mainly.

vitaly-t:
> range of node versions (6.3 - 8 at least) asap. This will be done before the release.

Are you sure? Last I spoke to Oliver - 1 day ago, Node 8.x was out of the question till after the 1.0.0 release.

By the way, I was the one asking to move toward Node.js 8.x :wink:

willclarktech:
@vitaly-t It sounds like you're talking about the version range for which we support people running Lisk Core. Lisk Elements, as a library intended for use in other projects has different requirements here.

docs/PULL_REQUEST_TEMPLATE.md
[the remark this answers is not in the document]
willclarktech: GitHub has some magic here. ✨

Jenkinsfile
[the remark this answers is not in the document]
Tosch110: I think we had it for a reason @fchavant - can you have a look please? :)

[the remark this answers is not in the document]
shuse2: I think main reason is the `cypress` dependency, which takes quite long time.
However, I seems like not being a problem with current `1.0.0`, also test should be as stateless as possible.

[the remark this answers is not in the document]
Tosch110: Okay, that would be quite cool actually. Yes, I agree to your point.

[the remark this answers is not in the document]
fchavant: @Tosch110 I seem to remember it was to prevent downloading cypress every single time, as @shuse2 pointed out.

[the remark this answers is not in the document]
shuse2: @fchavant 
Should we keep the cache for the jenkins work load?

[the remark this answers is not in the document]
fchavant: @shuse2 from your comment I assume cypress was gone from 1.0.0 -- in that case we do not need any cache.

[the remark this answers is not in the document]
willclarktech: Cypress is still around.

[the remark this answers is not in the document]
shuse2: @fchavant Ah, sorry. What I meant was not being problem with this *current* branch.
Install dependency step finishes in 1.5 min

Tosch110:
As discussed, let us keep the package-lock.json (as we support node 5+) to provide consistency and as recommended by npm. But let us gitignore yarn.lock because even though we like yarn we do not support it.

Jenkinsfile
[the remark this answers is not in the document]
Tosch110: I think probably 1 minute of the 1.5 minutes is alone cypress. Since we do not care so much about cypress version, maintaining the cache here would make sense to me.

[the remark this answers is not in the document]
shuse2: If it is not critical to have cache here, I prefer not having the cache here.
For example, the version of cypress was affecting this PR, which would not be able to be detected easily if it was cached.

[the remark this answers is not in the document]
Tosch110: Oh. I was not aware this was causing problems already. In that case not cached is fine with me :+1:

PR #562, opened 2018-02-14
Add multiple node.js version support - Closes #561

### LINKED ISSUE #561: Support multiple versions of node.js
### Description
Currently, lisk-js only supports fixed node.js version of v6.12.3.
It would be great if it supports all major v6, and even v6 or higher.

### Affected versions

1.0.0

### Changes
- Change all `Buffer.from` to use crypto function.
- `Buffer.from` should ignore invalid hex strings.

shuse2:
Closes #561 

### Description
- Changed `hexToBuffer` behavior to mimic node version >= 8 in all versions.
- Changed `Buffer.from` to use `hexToBuffer`

### Note
Run test against
* v6.12.3
* v7.10.1
* v8.9.4
* v9.5.0

### Changes
- [x] Update all direct usage of `Buffer.from(hex, 'hex')`
- [x] `Buffer.from(hex, 'hex')` should throw Type error if first argument is not string
- [x] `Buffer.from(hex, 'hex')` should ignore invalid hex string

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
 	"description": "JavaScript library for sending Lisk transactions from the client or server",
 	"main": "./dist-node/index.js",
 	"engines": {
-		"node": ">=6 <=9",
+		"node": ">=6.3 <=9",
 		"npm": ">=3 <=5"
 	},
 	"scripts": {
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
 	"description": "JavaScript library for sending Lisk transactions from the client or server",
 	"main": "./dist-node/index.js",
 	"engines": {
-		"node": ">=6.3 <=9",
+		"node": ">=6.3 <=9.5",
 		"npm": ">=3 <=5"
 	},
 	"scripts": {
--- a/package.json
+++ b/package.json
@@ -3,6 +3,10 @@
 	"version": "0.4.5",
 	"description": "JavaScript library for sending Lisk transactions from the client or server",
 	"main": "./dist-node/index.js",
+	"engines": {
+		"node": ">=6 <=9",
+		"npm": "=>3 <=5"
+	},
 	"scripts": {
 		"prestart": "./bin/prestart.sh",
 		"start": "./bin/start.sh",
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
 	"main": "./dist-node/index.js",
 	"engines": {
 		"node": ">=6 <=9",
-		"npm": "=>3 <=5"
+		"npm": ">=3 <=5"
 	},
 	"scripts": {
 		"prestart": "./bin/prestart.sh",
--- a/src/cryptography/convert.js
+++ b/src/cryptography/convert.js
@@ -27,7 +27,7 @@ export const bufferToHex = buffer => naclInstance.to_hex(buffer);
 const hexRegex = /^[0-9a-f]+/i;
 export const hexToBuffer = hex => {
 	if (typeof hex !== 'string') {
-		throw new TypeError('Argument must be a string');
+		throw new TypeError('Argument must be a string.');
 	}
 	const matchedHex = (hex.match(hexRegex) || [])[0];
 	if (!matchedHex) {
--- a/src/cryptography/convert.js
+++ b/src/cryptography/convert.js
@@ -24,17 +24,17 @@ export const bufferToBigNumberString = bigNumberBuffer =>
 
 export const bufferToHex = buffer => naclInstance.to_hex(buffer);
 
-const hexRegex = /^([0-9]|[a-f])+/gi;
+const hexRegex = /^[0-9a-f]+/i;
 export const hexToBuffer = hex => {
 	if (typeof hex !== 'string') {
-		throw new TypeError('argument must be string');
+		throw new TypeError('Argument must be a string');
 	}
-	const matchedHex = hex.match(hexRegex) || [];
-	if (matchedHex.length === 0) {
+	const matchedHex = (hex.match(hexRegex) || [])[0];
+	if (!matchedHex) {
 		return Buffer.alloc(0);
 	}
-	const evenLength = Math.floor(matchedHex[0].length / 2) * 2;
-	return Buffer.from(matchedHex[0].slice(0, evenLength), 'hex');
+	const evenLength = Math.floor(matchedHex.length / 2) * 2;
+	return Buffer.from(matchedHex.slice(0, evenLength), 'hex');
 };
 
 export const getFirstEightBytesReversed = publicKeyBytes =>
--- a/src/cryptography/convert.js
+++ b/src/cryptography/convert.js
@@ -24,17 +24,17 @@ export const bufferToBigNumberString = bigNumberBuffer =>
 
 export const bufferToHex = buffer => naclInstance.to_hex(buffer);
 
-const hexRegex = /([0-9]|[a-f])/gim;
+const hexRegex = /^([0-9]|[a-f])+/gi;
 export const hexToBuffer = hex => {
 	if (typeof hex !== 'string') {
-		return Buffer.alloc(0);
+		throw new TypeError('argument must be string');
 	}
 	const matchedHex = hex.match(hexRegex) || [];
 	if (matchedHex.length === 0) {
 		return Buffer.alloc(0);
 	}
-	const evenLength = Math.floor(matchedHex.length / 2) * 2;
-	return Buffer.from(matchedHex.slice(0, evenLength).join(''), 'hex');
+	const evenLength = Math.floor(matchedHex[0].length / 2) * 2;
+	return Buffer.from(matchedHex[0].slice(0, evenLength), 'hex');
 };
 
 export const getFirstEightBytesReversed = publicKeyBytes =>
--- a/src/cryptography/convert.js
+++ b/src/cryptography/convert.js
@@ -24,7 +24,18 @@ export const bufferToBigNumberString = bigNumberBuffer =>
 
 export const bufferToHex = buffer => naclInstance.to_hex(buffer);
 
-export const hexToBuffer = hex => Buffer.from(hex, 'hex');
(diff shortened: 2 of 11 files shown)


package.json
```diff
@@ -3,6 +3,10 @@
 	"version": "0.4.5",
 	"description": "JavaScript library for sending Lisk transactions from the client or server",
 	"main": "./dist-node/index.js",
+	"engines": {
+		"node": ">=6 <=9",
+		"npm": "=>3 <=5"
```
AndrewBarba: Small type `=>3` should be `>=3`

src/crypto/convert.js
```diff
@@ -24,7 +24,18 @@ export const bufferToBigNumberString = bigNumberBuffer =>
 
 export const bufferToHex = buffer => naclInstance.to_hex(buffer);
 
-export const hexToBuffer = hex => Buffer.from(hex, 'hex');
+const hexRegex = /([0-9]|[a-f])/gim;
```
willclarktech: Do we really want multiline?

willclarktech: This regex gives odd behaviour.

```
> Buffer.from('abhcd', 'hex')
// Node 6: throws 'Invalid hex string'
<Buffer ab> // Node 8
> hexToBuffer('abhcd')
<Buffer abcd>
```

test/crypto/convert.js
```diff
@@ -65,6 +65,22 @@ describe('convert', () => {
...
+			return buffer.should.be.eql(Buffer.alloc(0));
+		});
+		it('should create partial Buffer from partially non hex string', () => {
+			const buffer = hexToBuffer('Abxzzzz');
+			return buffer.should.be.eql(Buffer.from('Ab', 'hex'));
+		});
+		it('should create even numbered Buffer from odd number hex string', () => {
+			const buffer = hexToBuffer('c3a5c3a4c3b6a');
+			return buffer.should.be.eql(Buffer.from('c3a5c3a4c3b6', 'hex'));
+		});
```
willclarktech: We need a test case where there's a non-hex character in the middle of hex characters.

test/crypto/convert.js
```diff
@@ -65,6 +65,22 @@ describe('convert', () => {
 			const buffer = hexToBuffer(defaultHex);
 			return buffer.should.be.eql(defaultBuffer);
 		});
+		it('should create empty Buffer from non-string', () => {
```
willclarktech: Should it? Node 6 and 8 both throw errors if you use a number for example.

shuse2: Will be changed to throw type error if it's not string

src/crypto/convert.js
shuse2: will be changed to take only first part =)

shuse2:
I think the `Buffer.from` used in the test is the controlled variable, which shouldn't throw error, so not really necessary to change there

shuse2:
Until node `v6.3.0`, createCipheriv with "aes-256-gcm" did not allow IV to be longer than 12 bytes.
but changed in 
[https://github.com/nodejs/node/pull/6376](https://github.com/nodejs/node/pull/6376)
In our code, we use `16 bytes`, so changed engine to start from `6.3`

src/crypto/convert.js
```diff
@@ -24,7 +24,18 @@ export const bufferToBigNumberString = bigNumberBuffer =>
 
 export const bufferToHex = buffer => naclInstance.to_hex(buffer);
 
-export const hexToBuffer = hex => Buffer.from(hex, 'hex');
+const hexRegex = /^([0-9]|[a-f])+/gi;
```
willclarktech: Let's make this regex as simple as possible. Does the following work?
```
const hexRegex = /^[0-9a-f]+/i;
```

package.json
```diff
@@ -3,6 +3,10 @@
 	"version": "0.4.5",
 	"description": "JavaScript library for sending Lisk transactions from the client or server",
 	"main": "./dist-node/index.js",
+	"engines": {
+		"node": ">=6.3 <=9",
```
willclarktech: We should put an upper bound on the latest version of 9 that we've tested.

src/crypto/convert.js
```diff
@@ -24,7 +24,18 @@ export const bufferToBigNumberString = bigNumberBuffer =>
...
+export const hexToBuffer = hex => {
+	if (typeof hex !== 'string') {
+		throw new TypeError('argument must be string');
```
willclarktech: `Argument must be a string`

src/crypto/convert.js
```diff
@@ -24,7 +24,18 @@ export const bufferToBigNumberString = bigNumberBuffer =>
...
+	}
+	const matchedHex = hex.match(hexRegex) || [];
+	if (matchedHex.length === 0) {
```
willclarktech: Why not just this?
```
const matchedHex = hex.match(hexRegex);
if (!matchedHex) {
    return Buffer.alloc(0);
}

**Comment by willclarktech** [comment]
Whoops, sorry just realised one of my `Buffer.from` finds in `src` was for the `hexToBuffer` function which obviously needs to remain.

File: test/crypto/convert.js [hunk 9]
```diff
@@ -65,6 +65,32 @@ describe('convert', () => {
...
+			return buffer.should.be.eql(Buffer.from('Ab', 'hex'));
+		});
+		it('should create partial Buffer with only first valid hex string', () => {
+			const buffer = hexToBuffer('Abxzzab');
+			return buffer.should.be.eql(Buffer.from('Ab', 'hex'));
+		});
+		it('should create even numbered Buffer from odd number hex string', () => {
+			const buffer = hexToBuffer('c3a5c3a4c3b6a');
+			return buffer.should.be.eql(Buffer.from('c3a5c3a4c3b6', 'hex'));
+		});
```
willclarktech: Would also be good to have a test case with a valid odd-numbered hex string followed by non-hex. Eg. `123xxx` and/or `123x`.

src/crypto/convert.js
shuse2: i think global key is not needed but `^([0-9]|[a-f])` this is more readable isn't it?

src/crypto/convert.js
shuse2: In the following code, it uses first element like `matchedHex[0]`, so I think it would be better to check length here already

shuse2:
> I get an installation failure with npm v5.0.0.
this was solved by install without cache.

src/crypto/convert.js
willclarktech: As discussed, a capturing group which never gets used is confusing.

test/cryptography/convert.js
```diff
@@ -65,6 +65,36 @@ describe('convert', () => {
 			const buffer = hexToBuffer(defaultHex);
 			return buffer.should.be.eql(defaultBuffer);
 		});
+		it('should throw TypeError with number', () => {
+			return (() => hexToBuffer(123)).should.throw(
```
willclarktech: Prefer `hexToBuffer.bind(null, 123).should.throw` syntax.

test/cryptography/convert.js
```diff
@@ -65,6 +65,36 @@ describe('convert', () => {
 			const buffer = hexToBuffer(defaultHex);
 			return buffer.should.be.eql(defaultBuffer);
 		});
+		it('should throw TypeError with number', () => {
```
willclarktech: Prefer an empty line between `it`s.

test/cryptography/convert.js
```diff
@@ -65,6 +65,36 @@ describe('convert', () => {
...
+			return (() => hexToBuffer(123)).should.throw(
+				new TypeError('argument must be string'),
```
willclarktech: This test gives false positives: the message is wrong here.

src/cryptography/convert.js
```diff
@@ -24,7 +24,18 @@ export const bufferToBigNumberString = bigNumberBuffer =>
 
 export const bufferToHex = buffer => naclInstance.to_hex(buffer);
 
-export const hexToBuffer = hex => Buffer.from(hex, 'hex');
+const hexRegex = /^[0-9a-f]+/i;
```
Tosch110: 👍

src/cryptography/convert.js
```diff
@@ -24,7 +24,18 @@ export const bufferToBigNumberString = bigNumberBuffer =>
...
+export const hexToBuffer = hex => {
+	if (typeof hex !== 'string') {
+		throw new TypeError('Argument must be a string.');
+	}
+	const matchedHex = (hex.match(hexRegex) || [])[0];
+	if (!matchedHex) {
+		return Buffer.alloc(0);
+	}
+	const evenLength = Math.floor(matchedHex.length / 2) * 2;
+	return Buffer.from(matchedHex.slice(0, evenLength), 'hex');
```
Tosch110: Do you think it is better to return a sliced element here rather than throwing an error in case it is not even?

Tosch110: Same applies to non-hex strings

shuse2: I think main reason not to throw an error for non-even input is because node 8 is not behaving like that.

shuse2: Also, for the non-hex string, node 8 ignores the non hex string, and create the buffer from whatever it can from the front.
```
> process.version
'v8.9.4'
> Buffer.from('abcde', 'hex');
<Buffer ab cd>
> Buffer.from('abcdzz123', 'hex');
<Buffer ab cd>
```

willclarktech: But this is our opportunity to make the behaviour we actually want.

shuse2: https://github.com/nodejs/node/pull/12012#issuecomment-289107822
I think main thing is that parallel behavior with `base64` and coherent eror handlings of various things it can happen

Tosch110: Actually the sentiment seems to be like "you can do validation on your side if you want" on the node side of things: https://github.com/nodejs/node/issues/8569

And in my opinion I would prefer to have an error thrown.

PR #565, opened 2018-02-19
Update API definition to match v1.0.0 - Closes #482

### LINKED ISSUE #482: Update API to 1.0.0 finalised version
Since we refactored the API, there have been some changes to the specification that we should reflect in LiskJS.

shuse2:
Closes #482 

~Merge after #512~

### DIFF
--- a/src/api/resources/delegates.js
+++ b/src/api/resources/delegates.js
@@ -33,5 +33,16 @@ export default class DelegatesResource extends APIResource {
 				offset: 101,
 			},
 		}).bind(this);
+
+		this.getForgers = apiMethod({
+			method: GET,
+			path: '/forgers',
+		}).bind(this);
+
+		this.getForgingStatus = apiMethod({
+			method: GET,
+			path: '/{address}/forging_stats',
+			urlParams: ['address'],
+		}).bind(this);
 	}
 }
--- a/src/api/resources/delegates.js
+++ b/src/api/resources/delegates.js
@@ -39,7 +39,7 @@ export default class DelegatesResource extends APIResource {
 			path: '/forgers',
 		}).bind(this);
 
-		this.getForgingStatus = apiMethod({
+		this.getForgingStats = apiMethod({
 			method: GET,
 			path: '/{address}/forging_stats',
 			urlParams: ['address'],
--- a/src/api/resources/delegates.js
+++ b/src/api/resources/delegates.js
@@ -29,7 +29,7 @@ export default class DelegatesResource extends APIResource {
 		this.getStandby = apiMethod({
 			method: GET,
 			defaultData: {
-				orderBy: 'rate:asc',
+				sort: 'rate:asc',
 				offset: 101,
 			},
 		}).bind(this);
--- a/src/api/resources/index.js
+++ b/src/api/resources/index.js
@@ -16,7 +16,7 @@ export { default as AccountsResource } from './accounts';
 export { default as BlocksResource } from './blocks';
 export { default as DappsResource } from './dapps';
 export { default as DelegatesResource } from './delegates';
-export { default as NodesResource } from './nodes';
+export { default as NodeResource } from './node';
 export { default as PeersResource } from './peers';
 export { default as SignaturesResource } from './signatures';
 export { default as TransactionsResource } from './transactions';
--- a/src/api/resources/index.js
+++ b/src/api/resources/index.js
@@ -12,11 +12,12 @@
  * Removal or modification of this copyright notice is prohibited.
  *
  */
-
 export { default as AccountsResource } from './accounts';
 export { default as BlocksResource } from './blocks';
 export { default as DappsResource } from './dapps';
 export { default as DelegatesResource } from './delegates';
+export { default as NodesResource } from './nodes';
+export { default as PeersResource } from './peers';
 export { default as SignaturesResource } from './signatures';
 export { default as TransactionsResource } from './transactions';
 export { default as VotersResource } from './voters';
--- a/src/api/resources/nodes.js
+++ b/src/api/resources/nodes.js
@@ -13,30 +13,39 @@
  *
  */
 
-import { GET } from 'constants';
-import APIResource from '../api_resource';
-import apiMethod from '../api_method';
+import { GET, PUT } from 'constants';
+import apiMethod from '../apiMethod';
+import APIResource from '../apiResource';
 
-export default class AccountsResource extends APIResource {
+export default class NodesResource extends APIResource {
 	constructor(liskAPI) {
 		super(liskAPI);
+		this.path = '/nodes';
 
-		this.path = '/accounts';
-
-		this.get = apiMethod({
+		this.getConstants = apiMethod({
 			method: GET,
+			path: '/constants',
 		}).bind(this);
 
-		this.getMultisignatureGroup = apiMethod({
+		this.getStatus = apiMethod({
 			method: GET,
-			path: '/{address}/multisignature_groups',
-			urlParams: ['address'],
+			path: '/status',
 		}).bind(this);
 
-		this.getMultisignatureMembership = apiMethod({
+		this.getForgingStatus = apiMethod({
 			method: GET,
-			path: '/{address}/multisignature_memberships',
-			urlParams: ['address'],
+			path: '/status/forging',
+		}).bind(this);
+
+		this.updateForgingStatus = apiMethod({
+			method: PUT,
+			path: '/status/forging',
+		}).bind(this);
+
+		this.getTransactions = apiMethod({
+			method: GET,
+			path: '/transactions/{state}',
+			urlParams: ['state'],
 		}).bind(this);
 	}
 }
--- a/src/api/resources/transactions.js
+++ b/src/api/resources/transactions.js
@@ -29,10 +29,5 @@ export default class TransactionsResource extends APIResource {
 		this.create = apiMethod({
 			method: POST,
 		}).bind(this);
-
-		this.getUnsignedMultisignature = apiMethod({
-			method: GET,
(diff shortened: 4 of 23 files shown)


src/api/liskApi.js
```diff
@@ -56,6 +56,8 @@ export default class LiskAPI {
 		this.blocks = new resources.BlocksResource(this);
 		this.dapps = new resources.DappsResource(this);
 		this.delegates = new resources.DelegatesResource(this);
+		this.nodes = new resources.NodesResource(this);
```
Tosch110: On the core API this is singular `node` - we probably should stick to this

src/api/resources/nodes.js
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ *
+ */
+
+import { GET, PUT } from 'constants';
+import apiMethod from '../apiMethod';
+import APIResource from '../apiResource';
+
+export default class NodesResource extends APIResource {
```
Tosch110: Let us also use `NodeResource`

src/api/resources/nodes.js
```diff
@@ -0,0 +1,51 @@
...
+	constructor(liskAPI) {
+		super(liskAPI);
+		this.path = '/nodes';
```
Tosch110: path is `/node`

test/api/resources/nodes.js
```diff
@@ -0,0 +1,70 @@
...
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ *
+ */
+
+import APIResource from 'api/apiResource';
+import NodeResource from 'api/resources/nodes';
```
Tosch110: same

test/api/resources/nodes.js
```diff
@@ -0,0 +1,70 @@
...
+
+describe('NodesResource', () => {
+	const defaultBasePath = 'http://localhost:1234';
+	const path = '/nodes';
```
Tosch110: same

src/api/resources/delegates.js
```diff
@@ -33,5 +33,16 @@ export default class DelegatesResource extends APIResource {
 				offset: 101,
```
willclarktech: API spec says `sort` instead of `orderBy`, but that might be the default on the API anyway?

src/api/resources/accounts.js
```diff
@@ -26,5 +26,17 @@ export default class AccountsResource extends APIResource {
 		this.get = apiMethod({
 			method: GET,
```
willclarktech: This is the default. Do we want to be explicit here?

test/api/resources/peers.js
```diff
@@ -0,0 +1,62 @@
...
+			return (() => new PeerResource()).should.throw(
+				'Require LiskAPI instance to be initialized.',
+			);
+		});
+
+		it('should be instance of APIResource', () => {
+			return resource.should.be.instanceOf(APIResource);
+		});
+
+		it('should have correcrt full path', () => {
```
willclarktech: Typo: `correct`

test/api/resources/nodes.js
```diff
@@ -0,0 +1,70 @@
...
+			return (() => new NodeResource()).should.throw(
+				'Require LiskAPI instance to be initialized.',
+			);
+		});
+
+		it('should be instance of APIResource', () => {
+			return resource.should.be.instanceOf(APIResource);
+		});
+
+		it('should have correcrt full path', () => {
```
willclarktech: Typo: `correct`

test/api/resources/nodes.js
```diff
@@ -0,0 +1,70 @@
...
+ */
+
+import APIResource from 'api/apiResource';
+import NodeResource from 'api/resources/nodes';
+
+describe('NodesResource', () => {
+	const defaultBasePath = 'http://localhost:1234';
+	const path = '/nodes';
+
+	let LiskAPI;
```
willclarktech: Why begin with a capital?

test/api/resources/nodes.js
```diff
@@ -0,0 +1,70 @@
...
+		});
+
+		it('should set resource path', () => {
+			return resource.path.should.equal(path);
+		});
+
+		it('should have methods', () => {
+			return resource.should.have.keys(
+				'liskAPI',
+				'path',
```
willclarktech: These two aren't methods, and we already checked them. Also prefer `.should.have.all.keys` explicit syntax.

We aren't checking if these keys correspond to functions or some other type.

Also applies to other test files.

willclarktech:
Branch name has wrong issue number.

src/api/resources/accounts.js
shuse2: I think we should be explicit as much as possible

test/api/resources/delegates.js
```diff
@@ -53,15 +53,20 @@ describe('DelegatesResource', () => {
...
-				'getStandby',
-				'getForgers',
-				'getForgingStatus',
-			);
+		it('should have a "get" function', () => {
+			return resource.should.have.property('get').which.is.a('function');
+		});
+
+		it('should have a "getStandby" function', () => {
+			return resource.should.have.property('get').which.is.a('function');
```
willclarktech: These names need to match the tests.

test/api/resources/node.js
```diff
@@ -45,24 +45,40 @@ describe('NodeResource', () => {
...
+
+		it('should have a "getForgingStatus" function', () => {
+			return resource.should.have
+				.property('getForgingStatus')
+				.which.is.a('function');
+		});
+
+		it('should have a "updateForgingStatus" function', () => {
+			return resource.should.have
+				.property('getForgingStatus')
```
willclarktech: This needs to be updated.

src/api/resources/delegates.js
```diff
@@ -29,9 +29,20 @@ export default class DelegatesResource extends APIResource {
...
 			},
 		}).bind(this);
+
+		this.getForgers = apiMethod({
+			method: GET,
+			path: '/forgers',
+		}).bind(this);
+
+		this.getForgingStatus = apiMethod({
```
willclarktech: Typo: `getForgingStats`

src/api/resources/index.js
```diff
@@ -12,11 +12,12 @@
  * Removal or modification of this copyright notice is prohibited.
  *
  */
-
 export { default as AccountsResource } from './accounts';
```
willclarktech: We could use a test for the exports from this file.

src/api/resources/delegates.js
willclarktech: Although since this is confusing, maybe we want to call it `getForgingStatistics`?

test/api/resources/node.js
```diff
@@ -0,0 +1,84 @@
...
+			return resource.should.have
+				.property('getConstants')
+				.which.is.a('function');
+		});
+
+		it('should have a "getStatus" function', () => {
+			return resource.should.have.property('getStatus').which.is.a('function');
+		});
+
+		it('should have a "getTransactions" function', () => {
```
willclarktech: Prefer matching the order defined in the source file.

src/api/resources/transactions.js
```diff
@@ -29,10 +29,5 @@ export default class TransactionsResource extends APIResource {
 		this.create = apiMethod({
```
willclarktech: Should this be `create` or `broadcast`? Same applies to signatures. @Tosch110

test/api/resources/node.js
```diff
@@ -0,0 +1,84 @@
...
+
+	beforeEach(() => {
+		liskAPI = {
+			headers: {},
+			nodeFullURL: defaultBasePath,
+			hasAvailableNodes: () => {},
+			randomizeNodes: () => {},
+			banActiveNodeAndSelect: () => {},
+		};
+		resource = new NodeResource(liskAPI);
```
willclarktech: Needs to be async now. (Also applies to other tests.)

src/api/resources/transactions.js
Tosch110: good point! When it only sends (broadcasts) the transaction and signature I find `broadcast` better. It does not create anything.

shuse2: I think `API` wise it's attempting to create.
`broadcast` itself is done by the nodes

test/api/resources/node.js
shuse2: Yes! i guess depends on which PR goes in first =0

src/api/resources/delegates.js
shuse2: I see I got confused but wouldn't it be better to match API definition? `getForgingStats`

src/api/resources/transactions.js
Tosch110: The confusion already starts. I think this is `broadcast` rather, we provide the already created transaction/signature and `broadcast` it to the node with this API.

src/api/resources/delegates.js
willclarktech: Probably it would be better to update the API definition.

willclarktech: https://github.com/LiskHQ/lisk/issues/1713

src/constants/index.js
```diff
@@ -34,6 +34,7 @@ export const SSL_PORT = '443';
 
 export const GET = 'GET';
 export const POST = 'POST';
+export const PUT = 'PUT';
```
Tosch110: Could you add a test case for this constant?
In https://github.com/LiskHQ/lisk-js/pull/599 I have added all the other missing constants.

PR #566, opened 2018-02-19
Update cypress version to 2.0.2 - Closes #563

### LINKED ISSUE #563: Update cypress dependency
We're still using a 0.x version, but v1.x is out.

shuse2:
Closes #563


willclarktech:
@shuse2 Can you correct the spelling in the commit message?

cypress/plugins/index.js
```diff
@@ -0,0 +1,18 @@
...
+// ***********************************************************
+
+// This function is called when a project is opened or re-opened (e.g. due to
+// the project's config changing)
+
+module.exports = () => {
+	// module.exports = (on, config) => {
+	// `on` is used to hook into various events Cypress emits
+	// `config` is the resolved Cypress config
+};
```
willclarktech: Do we need this file?

shuse2: It is auto generated, so I think it would be better to keep it.

willclarktech: Why not add it to `.gitignore` and add it to vcs when necessary (if ever)?

shuse2: sounds good to me 👍

PR #584, opened 2018-03-06
Check amount in getTransactionBytes - Closes #543

### LINKED ISSUE #543: Confirm use of e.g. Buffer.alloc in getTransactionBytes
Child Issue of https://github.com/LiskHQ/lisk-js/issues/502

willclarktech:
### What was the problem?

We allowed all amounts as inputs to GTB.

### How did I fix it?

Added checks for amount range. We might want to extend this to other properties too?

### Review checklist

* The PR solves #543 
* Linting passes
* Tests pass
* Commit messages follow the

### DIFF
--- a/src/transactions/utils/getTransactionBytes.js
+++ b/src/transactions/utils/getTransactionBytes.js
@@ -14,6 +14,7 @@
  */
 import bignum from 'browserify-bignum';
 import cryptography from 'cryptography';
+import { MAX_TRANSACTION_AMOUNT } from 'constants';
 
 export const isValidValue = value => ![undefined, false, NaN].includes(value);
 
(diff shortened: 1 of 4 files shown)


src/transactions/utils/getTransactionBytes.js
```diff
@@ -14,6 +14,7 @@
  */
 import bignum from 'browserify-bignum';
 import cryptography from 'cryptography';
+import { MAX_TRANSACTION_AMOUNT } from '../../constants';
```
shuse2: Prefer using alias
`import { MAX_TRANSACTION_AMOUNT } from 'constants';`

src/transactions/utils/getTransactionBytes.js
```diff
@@ -179,7 +180,16 @@ const getTransactionBytes = transaction => {
 			)
 		: Buffer.alloc(BYTESIZES.RECIPIENT_ID);
 
-	const transactionAmount = bignum(amount).toBuffer({
+	const amountBigNum = bignum(amount);
+	if (amountBigNum.lt(0)) {
```
shuse2: is `0` allowed considering the fee?

Tosch110: Amount !== fee - but in general that is a good question.

Only type 0, 6 and 7 transactions have a configurable amount, else it will be default to 0.

shuse2:
LGTM

PR #596, opened 2018-03-09
Update liskAPI constructor and added factory methods - Closes #588

### LINKED ISSUE #588: Refactor constructor of LiskAPI
### Description

Limit the constructor options to what actually needed.

#### Change name
`LiskAPI` => `HttpClient`

#### Limit constructor options
nodes: [string] (required)
nethash: string (required)
node: string default select from nodes
bannedNodes: [string] default []
verion: string default 1.0.0
minVersion: string default 1.0.0
randomizeNode: boolean default true

#### Add factory method
- Add getLiskMainnetHttpClient()
- Add getLiskTestnetHttpClient()
which returns a new HTTP client instance with each settings.

#### Separate headers
- `version`, `minVersion` and `nethash` is the only thing should be changed for headers
#### ssl/testnets and default settings
- `ssl` and `testnet` should be indicated by `nodes` (e.g. ['https://testnet.io:443`, 'http://testnet.io:7001'])
- it is a library, so it should not depends on `config.json`, it should be default constants.

### Versions

v1.0.0 branch

shuse2:
### What was the problem?
In liskAPI, there was a lot of properties, and some of them had sideeffects when it's changed.

### How did I fix it?
Refactor LiskAPI constructor and properties to be used in more generic ways.

### Review checklist 

* The PR solves #588
* Linting passes
* Tests pass
* Commit messages follow the


src/api_client/api_client.js
```diff
@@ -0,0 +1,112 @@
...
+	Object.assign({}, commonHeaders, {
+		nethash,
+		version,
+		minVersion,
+	});
+
+export default class APIClient {
+	constructor(nodes, nethash, providedOptions = {}) {
+		if (!Array.isArray(nodes) || nodes.length <= 0) {
+			throw Error('Require nodes to be initialized.');
```
willclarktech: Prefer `new` syntax. Also message could be improved. Maybe `APIClient requires nodes for initialization.`?

src/api_client/api_client.js
```diff
@@ -0,0 +1,112 @@
...
+		}
+
+		if (typeof nethash !== 'string' || nethash === '') {
+			throw Error('Require nethash to be initialized.');
```
willclarktech: As above.

src/api_client/api_client.js
```diff
@@ -0,0 +1,112 @@
...
+	hasAvailableNodes() {
+		return this.nodes.some(node => !this.isBanned(node));
+	}
+
+	isBanned(node) {
+		return this.bannedNodes.includes(node);
+	}
+}
+
+export const getMainnetClient = options =>
```
willclarktech: How about `createMainnetAPIClient`?

src/api_client/api_client.js
```diff
@@ -0,0 +1,112 @@
...
+	new APIClient(MAINNET_NODES, MAINNET_NETHASH, options);
+
+export const getTestnetClient = options =>
```
willclarktech: As above.

src/api_client/index.js
```diff
@@ -13,6 +13,7 @@
  *
  */
 
-import liskAPI from './liskApi';
+import APIClient from './api_client';
 
-export default liskAPI;
+export default APIClient;
+export { getMainnetClient, getTestnetClient } from './api_client';
```
willclarktech: Prefer the following syntax:
```
export {
    default as APIClient,
    getMainnetClient,
    getTestnetClient,
} from './api_client';
```

src/api_client/package.json
```diff
@@ -0,0 +1,3 @@
+{
+	"name": "apiClient"
```
willclarktech: Is this the capitalization style we want?

src/api_client/apiResource.js
```diff
@@ -16,20 +16,20 @@
 import * as popsicle from 'popsicle';
 
 export default class APIResource {
-	constructor(liskAPI) {
-		if (!liskAPI) {
-			throw Error('Require LiskAPI instance to be initialized.');
+	constructor(apiClient) {
+		if (!apiClient) {
+			throw Error('Require APIClient instance to be initialized.');
```
willclarktech: As below.

src/constants/index.js
```diff
@@ -42,5 +42,16 @@ export const MAX_TRANSACTION_AMOUNT = '18446744073709551615';
 
 export const TESTNET_NETHASH =
 	'da3ed6a45429278bac2666961289ca17ad86595d33b31037615d4b8e8f158bba';
+export const TESTNET_NODES = ['testnet.lisk.io'];
```
willclarktech: Should we fill this out with more?

src/index.js
```diff
@@ -26,4 +26,11 @@ naclFactory.instantiate(nacl => {
 	naclInstance = nacl;
 });
 
-export default { cryptography, transaction, API, time, passphrase };
+export default {
+	APIClient,
+	apiUtils: { getMainnetClient, getTestnetClient },
```
willclarktech: We should be consistent with the pattern used in transactions for utils.

test/api_client/api_client.js
```diff
@@ -0,0 +1,330 @@
...
+		});
+
+		it('should throw an error if no second argument is passed to constructor', () => {
+			return (() => new APIClient(defaultNodes)).should.throw(
+				Error,
+				'Require nethash to be initialized.',
+			);
+		});
+
+		it('should throw an error if second argument is passed to constructor is not string', () => {
```
willclarktech: `should throw an error if second argument passed to constructor is not string`. Also applies to other tests.

test/api_client/api_client.js
```diff
@@ -0,0 +1,330 @@
...
+		});
+
+		it('should throw an error if second argument is passed to constructor is empty string', () => {
+			return (() => new APIClient(defaultNodes, '')).should.throw(
+				Error,
+				'Require nethash to be initialized.',
+			);
+		});
+
+		it('should throw an error if second argument is passed to constructor is empty string', () => {
```
willclarktech: Duplicates previous test.

test/api_client/api_client.js
```diff
@@ -0,0 +1,330 @@
...
+			it('should set with passed nethash, with default options', () => {
+				return apiClient.should.have
+					.property('headers')
+					.and.eql(defaultHeaders);
+			});
+
+			it('should set custom headers with supplied options', () => {
+				apiClient = new APIClient(defaultNodes, testnetHash, {
+					version: '0.5.0',
+					minVersion: '>=0.1.0',
```
willclarktech: Would perhaps be clearer to extract these values from `customHeaders` here.

test/api_client/api_client.js
```diff
@@ -0,0 +1,330 @@
...
+				});
+				return apiClient.should.have.property('headers').and.eql(customHeaders);
+			});
+		});
+
+		describe('nodes', () => {
+			it('should set with random node with initialized setup if no node is specified by options', () => {
```
willclarktech: This description doesn't match what's tested.

src/api_client/api_client.js
```diff
@@ -0,0 +1,112 @@
...
+		this.voters = new resources.VotersResource(this);
+		this.votes = new resources.VotesResource(this);
+	}
+
+	getNewNode() {
+		const nodes = this.nodes.filter(node => !this.isBanned(node));
+
+		if (nodes.length === 0) {
+			throw new Error(
+				'Cannot get random node: all relevant nodes have been banned.',
```
willclarktech: `Cannot get new node: all nodes have been banned.`

src/api_client/api_client.js
```diff
@@ -0,0 +1,112 @@
...
+			);
+		}
+
+		const randomIndex = Math.floor(Math.random() * nodes.length);
+		return nodes[randomIndex];
+	}
+
+	banActiveNode() {
```
willclarktech: Do we want a method for banning a passed node directly? Then `banActiveNode` could make use of that method.

src/api_client/api_client.js
```diff
@@ -0,0 +1,112 @@
...
+		if (!this.isBanned(this.currentNode)) {
+			this.bannedNodes.push(this.currentNode);
+			return true;
+		}
+		return false;
+	}
+
+	banActiveNodeAndSelect() {
```
willclarktech: Does `Select` match our terminology now?

test/api_client/api_client.js
```diff
@@ -0,0 +1,329 @@
...
+				apiClient = new APIClient(defaultNodes, mainnetHash, { bannedNodes });
+				return apiClient.should.have
+					.property('bannedNodes')
+					.be.eql(bannedNodes);
+			});
+		});
+
+		describe('currentNode', () => {
+			it('should set with random node with initialized setup if no node is specified by options', () => {
+				return apiClient.should.have.property('currentNode').and.not.empty;
```
willclarktech: `.and.not.be.empty` or `.which.is.not.empty`

test/api_client/api_client.js
```diff
@@ -0,0 +1,329 @@
...
+			return apiClient.isBanned(currentNode).should.be.true;
+		});
+
+		it('should not duplicate a banned node', () => {
+			const bannedNodes = [currentNode];
+			apiClient.bannedNodes = bannedNodes;
+			apiClient.banActiveNode();
+
+			return apiClient.bannedNodes.should.be.eql(bannedNodes);
+		});
```
willclarktech: Do we want tests for return values here? We do use them elsewhere now.

test/api_client/resources/delegates.js
```diff
@@ -55,7 +55,12 @@ describe('DelegatesResource', () => {
 		});
 
 		it('should have methods', () => {
-			return resource.should.have.keys('liskAPI', 'path', 'get', 'getStandby');
+			return resource.should.have.keys(
```
willclarktech: This will be sorted after rebasing, right?

shuse2: yes!

src/api_client/index.js
shuse2: I think this would have different result.
`APIClient` is not exported as default anymore

src/api_client/api_client.js
shuse2: I think so since it's `selecting` into `currentNode`

src/api_client/index.js
willclarktech: Ah yeah, fair point.

test/api_client/api_client.js
shuse2: Already resolved

src/index.js
shuse2: do u mean changing `apiUtils` to some other name?

willclarktech: I mean in `transactions` we have a `.utils` property for accessing transactions utils.

src/index.js
```diff
@@ -26,4 +26,10 @@ naclFactory.instantiate(nacl => {
 	naclInstance = nacl;
 });
 
-export default { cryptography, transaction, API, time, passphrase };
+export default {
+	APIClient,
+	cryptography,
+	passphrase,
+	transaction,
+	time,
```
Tosch110: `time` should already be part of `transaction.utils` - what we need here is exporting the `lisk-constants` module.

PR #597, opened 2018-03-09
Update constants to lisk-constants - Closes #589

### LINKED ISSUE #589: Refactor constants package
### Description

constants for specific package should be defined in each package.

e.g. `GET`, `POST` etc

### Versions

v1.0.0 branch

shuse2:
### What was the problem?
- Package specific constants were exposed to global
- Package `constants` had conflict with node build-in package name

### How did I fix it?
- change package name to `lisk-constants`
- move package specific constants to `constants.js`

### Review checklist 

* The PR solves #589
* Linting passes
* Tests pass
* Commit messages follow the

### DIFF
--- a/src/lisk-constants/index.js
+++ b/src/lisk-constants/index.js
@@ -28,14 +28,6 @@ export const EPOCH_TIME = new Date(Date.UTC(2016, 4, 24, 17, 0, 0, 0));
 export const EPOCH_TIME_MILLISECONDS = EPOCH_TIME.getTime();
 export const EPOCH_TIME_SECONDS = Math.floor(EPOCH_TIME.getTime() / 1000);
 
-export const LIVE_PORT = '8000';
-export const TEST_PORT = '7000';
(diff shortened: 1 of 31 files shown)


src/lisk-constants/index.js
```diff
@@ -28,13 +28,6 @@ export const EPOCH_TIME = new Date(Date.UTC(2016, 4, 24, 17, 0, 0, 0));
 export const EPOCH_TIME_MILLISECONDS = EPOCH_TIME.getTime();
```
willclarktech: Maybe fees are specific to the `transactions` module? They are part of the protocol though. Thoughts @Tosch110 ?

Tosch110: I think it makes sense especially considering the near planned future, to put it into the `transactions` module. I can see that these constants get replaced by functions. They will be gathering data about transaction fees according to last transactions accepted by the network. Those functions will probably be part of the `transactions` module, too.

PR #601, opened 2018-03-12
Update retry policy - Closes #587

### LINKED ISSUE #587: Update API retry policy
### Description
Currently, on API retry
- if `randomizeNode` is set to true, if will retry until success.
- it doesn't check the status code when it retries.

### Expected behavior
It should trigger retry if it is `error` or status code is not `200` and not `4xx`.
Moreover, 
If randomize is true, try until all the nodes are banned.
If randomize is false, try 3 times.

### Versions
v1.0.0 branch

shuse2:
### What was the problem?
If randomizeNode was not set to true, it will retry forever to the same node.

### How did I fix it?
limit the number of try to 3, if randomizeNode is set to false.

### How to test it?
`npm test`

### Review checklist

* The PR solves #587 
* Merge after #596 
* Linting passes
* Tests pass
* Commit messages follow the

### DIFF
--- a/src/api_client/api_resource.js
+++ b/src/api_client/api_resource.js
@@ -14,7 +14,7 @@
  */
 
 import * as popsicle from 'popsicle';
-import { MAX_RETRY_COUNT } from './constants';
+import { API_RECONNECT_MAX_RETRY_COUNT } from './constants';
 
 export default class APIResource {
 	constructor(apiClient) {
@@ -52,7 +52,7 @@ export default class APIResource {
 			return new Promise(resolve => setTimeout(resolve, 1000)).then(() => {
 				if (this.apiClient.randomizeNodes) {
 					this.apiClient.banActiveNodeAndSelect();
-				} else if (retryCount > MAX_RETRY_COUNT) {
+				} else if (retryCount > API_RECONNECT_MAX_RETRY_COUNT) {
 					throw error;
 				}
 				return this.request(req, true, retryCount + 1);
--- a/src/api_client/api_resource.js
+++ b/src/api_client/api_resource.js
@@ -14,6 +14,7 @@
  */
 
 import * as popsicle from 'popsicle';
+import { MAX_RETRY_COUNT } from './constants';
 
 export default class APIResource {
 	constructor(apiClient) {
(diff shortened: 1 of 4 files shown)


src/api_client/constants.js
```diff
@@ -15,3 +15,4 @@
 export const GET = 'GET';
 export const POST = 'POST';
 export const PUT = 'PUT';
+export const MAX_RETRY_COUNT = 3;
```
Tosch110: I think it makes sense to indicate what the retry count is for - how about `API_RECONNECT_MAX_RETRY_COUNT` ?

test/api_client/api_resource.js
```diff
@@ -89,8 +88,10 @@ describe('API resource module', () => {
 	describe('#request', () => {
 		let popsicleStub;
 		let handleRetryStub;
+		let defaultError;
 
 		beforeEach(() => {
+			defaultError = new Error('network error');
```
Tosch110: I know this is in the tests, a little more verbose error still makes sense to me.
Something like`could not connect to a node` error

test/api_client/api_resource.js
```diff
@@ -128,7 +129,9 @@ describe('API resource module', () => {
 
 	describe('#handleRetry', () => {
 		let requestStub;
+		let defaultError;
 		beforeEach(() => {
+			defaultError = new Error('network error');
```
Tosch110: same as above

src/api_client/api_resource.js
```diff
@@ -34,31 +35,29 @@ export default class APIResource {
...
-				return this.request(req, true);
+				return this.request(req, true, retryCount + 1);
 			});
 		}
-		return Promise.resolve({
-			success: false,
-			error,
-			message: 'Could not create an HTTP request to any known nodes.',
-		});
+		throw error;
```
willclarktech: Prefer making this `async` or rejecting a promise.

src/api_client/api_resource.js
```diff
@@ -34,31 +35,29 @@ export default class APIResource {
 		return `${this.apiClient.currentNode}/api${this.path}`;
 	}
 
-	request(req, retry) {
+	request(req, retry, retryCount = 1) {
```
willclarktech: Default value for `retry` should be `true`?

test/api_client/api_resource.js
```diff
@@ -161,14 +164,21 @@ describe('API resource module', () => {
...
+				const req = resource.handleRetry(defaultError, defaultRequest, 1);
 				clock.tick(1000);
 				return req.then(res => {
 					expect(apiClient.banActiveNodeAndSelect).not.to.be.called;
 					expect(requestStub).to.be.calledWith(defaultRequest, true);
 					return expect(res).to.be.eql(sendRequestResult.body);
 				});
 			});
+
+			it('should throw an error when randomizeNodes is false and if it exceed max retry count', () => {
```
willclarktech: `and the maximum retry count has been reached`

src/api_client/api_resource.js
shuse2: I think it should be false, we probably should decide which API should retry

PR #611, opened 2018-03-16
Add check for buffer length passed to cryptography toAddress - Closes #607

### LINKED ISSUE #607: Validation in crypto toAddress
### Description

cryptography.toAddress function right now accepts every input. It should probably be restricted to 8-byte input only, since this is being used for Lisk always. 

**Expected behavior:**

cryptography.toAddress should be clear on what it takes as parameter, so users know what to use in this place and do not get confused

**Actual behavior:** 

cryptography.toAddress take any buffer input and creates a bignum string

### Versions

1.0.0

Tosch110:
### What was the problem?

cryptography.toAddress did not have any input validation

### How did I fix it?

Adding validation for buffer !== Buffer.from(buffer).slice(0, 8)

### How to test it?

`npm test`

### Review checklist

* The PR solves #607
* Linting passes
* Tests pass
* Commit messages follow the

### DIFF
--- a/src/cryptography/convert.js
+++ b/src/cryptography/convert.js
@@ -44,7 +44,17 @@ export const getFirstEightBytesReversed = publicKeyBytes =>
 		.slice(0, 8)
 		.reverse();
 
-export const toAddress = buffer => `${bufferToBigNumberString(buffer)}L`;
+export const toAddress = buffer => {
+	if (
+		!Buffer.from(buffer)
+			.slice(0, 8)
+			.equals(buffer)
+	)
+		throw new Error(
+			'The buffer for Lisk addresses must not have more than 8 bytes',
+		);
+	return `${bufferToBigNumberString(buffer)}L`;
+};
 
 export const getAddressFromPublicKey = publicKey => {
 	const publicKeyHash = hash(publicKey, 'hex');
--- a/test/cryptography/convert.js
+++ b/test/cryptography/convert.js
@@ -141,6 +141,15 @@ describe('convert', () => {
 			const address = toAddress(bufferInit);
 			return expect(address).to.be.eql(defaultAddressFromBuffer);
 		});
+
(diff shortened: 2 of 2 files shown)


src/cryptography/convert.js
```diff
@@ -44,7 +44,12 @@ export const getFirstEightBytesReversed = publicKeyBytes =>
 		.slice(0, 8)
 		.reverse();
 
-export const toAddress = buffer => `${bufferToBigNumberString(buffer)}L`;
+export const toAddress = buffer => {
+	const firstEightBytes = Buffer.from(buffer).slice(0, 8);
+	if (firstEightBytes.length < buffer.length)
```
willclarktech: Why not just check the length directly?
```
if (Buffer.from(buffer).length > 8)
```
or even
```
if (buffer.length > 8)
```

Tosch110: I thought having the Buffer.from(buffer).slice(0, 8) in there will make sure that the input is a proper Buffer.

src/cryptography/convert.js
```diff
@@ -44,7 +44,17 @@ export const getFirstEightBytesReversed = publicKeyBytes =>
 		.slice(0, 8)
 		.reverse();
 
-export const toAddress = buffer => `${bufferToBigNumberString(buffer)}L`;
+export const toAddress = buffer => {
+	if (
+		!Buffer.from(buffer)
+			.slice(0, 8)
+			.equals(buffer)
```
willclarktech: This will break if input is an `ArrayBuffer`. Do we want that?

Tosch110: Probably yes. When you input an ArrayBuffer here, this probably means you have not used getFirstEightBytesReversed before. It is likely that people end up with a wrong address. So I would say keep it strict.

PR #615, opened 2018-03-19
Add recipientId field in register delegate transaction

### LINKED ISSUE #614: Delegate registration transaction doesn't pass validation on the node side.
### Description
Delegate registration transaction doesn't pass validation on the node side.
When you create a delegate registration transaction with `lisk-js 1.0.0` it creates transaction object with `recipientId` field equal to `null`. This transaction doesn't pass the validation on the `lisk 1.0.0` node side when you pass it through the api. And return the error: 
```
{
	"message": "Validation errors",
	"errors": [{
		"code": "INVALID_REQUEST_PARAMETER",
		"message": "Invalid parameter (transaction): Value failed JSON Schema validation",
		"in": "body",
		"name": "transaction",
		"errors": [{
			"code": "INVALID_TYPE",
			"message": "Expected type string but found type null",
			"path": ["recipientId"]
		}, {
			"code": "INVALID_FORMAT",
			"message": "Object didn't pass validation for format address: null",
			"path": ["recipientId"]
		}]
	}]
}
```

### Steps to Reproduce

1. Create a transaction for delegate registration with `lisk-js 1.0.0`
2. Pass it to the node `lisk 1.0.0` through the `api` (api/transactions)

**Expected behavior:** 
A delegate registration transaction will pass validation step

**Actual behavior:**
It causes a validation error

**Reproduces how often:**
Constatly

### Versions
`lisk-js 1.0.0`
`lisk 1.0.0`

alepop:
close #614

### DIFF
--- a/src/transactions/2_register_delegate.js
+++ b/src/transactions/2_register_delegate.js
@@ -12,17 +12,25 @@
  * Removal or modification of this copyright notice is prohibited.
  *
  */
+import cryptography from 'cryptography';
 import { DELEGATE_FEE } from './constants';
 import { wrapTransactionCreator } from './utils';
 
-const registerDelegate = ({ username }) => ({
-	type: 2,
-	fee: DELEGATE_FEE.toString(),
-	asset: {
-		delegate: {
-			username,
+const registerDelegate = ({ passphrase, username }) => {
+	const recipientId = passphrase
+		? cryptography.getAddressAndPublicKeyFromPassphrase(passphrase).address
+		: null;
+
+	return {
+		type: 2,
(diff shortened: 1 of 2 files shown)


shuse2:
As I mentioned on https://github.com/LiskHQ/lisk-js/issues/614#issuecomment-374635618, 
rather than fixing it everywhere, I think it should be set as default on `wrap_transaction_creator`.
What do you guys think? @Tosch110 @willclarktech

Tosch110:
Thanks for your help alepop!

This has been solved by https://github.com/LiskHQ/lisk-js/pull/620

PR #617, opened 2018-03-20
Fix buffer allocation for dapp asset - Closes #616

### LINKED ISSUE #616: Invalid id on transaction type 5
### Description

Last version of LiskJS produces an invalid ID for DApp registrations transaction type 5. The problem was discovered while implementing the migration in https://github.com/LiskHQ/lisk/issues/1629.
After some research, we realised LiskJS has removed a necessary part of the code when getting bytes from DApp asset. Exactly related to these lines: https://github.com/LiskHQ/lisk/blob/ab613eb7da7952169a1457cd1274c588612152be/logic/dapp.js#L287

The problem is been overlooked because type and category used in the transaction type 5 fixture is both 0 which produces a wrong output. When we use a random integer not being 0 tests fail.
https://github.com/LiskHQ/lisk-js/blob/72ea39e25c14b30cbb39bebf5742b2fd680e16f1/fixtures/transactions.json#L99

**Expected behavior:**
LiskJS replicates the exact behaviour from Core.

**Actual behavior:**
DApp registrations with type or category different than 0, produce invalid ID which cannot be validated by Core.

**Reproduces how often:**
Always.

### Versions
`1.0.0`

### Additional Information
This issue is blocking Core migration https://github.com/LiskHQ/lisk/issues/1629

shuse2:
### What was the problem?
Dapp transactionId did not match to `lisk-core`.
When we create the buffer for the asset field for Dapp, we were doing
```
const typeBuffer = Buffer.alloc(4, type);
const categoryBuffer = Buffer.alloc(4, category);
```
This would fill each byte with the second argument of the `alloc`
```
Buffer.alloc(4, 3)
<Buffer 03 03 03 03>
```
However what `lisk-core` was expecting was
```
<Buffer 03 00 00 00>
```
which is done in [here](https://github.com/LiskHQ/lisk/blob/ab613eb7da7952169a1457cd1274c588612152be/logic/dapp.js#L287-L290)

### How did I fix it?
Instead of allocating to each bytes,
```
const categoryBuffer = Buffer.alloc(4);
categoryBuffer.writeIntLE(category, 0);
```
### How to test it?
Create transaction and send it to lisk-core

### Review checklist

* The PR solves #616 
* Linting passes
* Tests pass
* Commit messages follow the

### DIFF
--- a/src/transactions/utils/get_transaction_bytes.js
+++ b/src/transactions/utils/get_transaction_bytes.js
@@ -83,8 +83,10 @@ export const getAssetDataForCreateDappTransaction = ({ dapp }) => {
 	const { name, description, tags, link, icon, type, category } = dapp;
 	const nameBuffer = Buffer.from(name, 'utf8');
 	const linkBuffer = Buffer.from(link, 'utf8');
-	const typeBuffer = Buffer.alloc(4, type);
-	const categoryBuffer = Buffer.alloc(4, category);
+	const typeBuffer = Buffer.alloc(4);
+	typeBuffer.writeIntLE(type, 0);
+	const categoryBuffer = Buffer.alloc(4);
+	categoryBuffer.writeIntLE(category, 0);
 
 	const descriptionBuffer = description
 		? Buffer.from(description, 'utf8')


diego-G:
I also suggest to change the fixture to have different type and category than 0 or to extend the test coverage somehow to address that situation.

Tosch110:
Thanks Diego, we have already an intensive testsuites for all kind of transactions: 
https://github.com/LiskHQ/lisk-js/tree/development/fixtures

It is possible that this has been problematic for a longer time though. Good to have a fix now!

PR #622, opened 2018-03-21
Change APIClient to throw error when API returns non 2xx - Closes #621

### LINKED ISSUE #621: APIClient should check status code and throw error if it's not 200
### Description
popsicle does not handle status code, and for users there is no good way to check if the API was success or not. 

**Expected behavior:
If API status is not 2xx, it should throw error

**Actual behavior:
All the API response are resolved in promise

shuse2:
### What was the problem?
When APIClient gets non-2xx status code, it does not throw an error.
If we return the body as the way currently work, there is no good way to check if it is an error or not from user.

### How did I fix it?
Check if the status code is greater than 300 (>= 300), then if it's 300 then throw an error.
If API supplies error message, throw an error with the message.
If not, return with unknown error.

### How to test it?
`npm test` and try to use it with Lisk core 1.0.0

### Review checklist

* The PR solves #621
* Linting passes
* Tests pass
* Commit messages follow the

### DIFF
--- a/src/api_client/api_resource.js
+++ b/src/api_client/api_resource.js
@@ -42,9 +42,11 @@ export default class APIResource {
 			.then(res => {
 				if (res.status >= 300) {
 					if (res.body && res.body.message) {
-						throw new Error(res.body.message);
+						throw new Error(`Status ${res.status} : ${res.body.message}`);
 					}
-					throw new Error('An unknown error has occurred.');
+					throw new Error(
+						`Status ${res.status} : An unknown error has occurred.`,
+					);
 				}
 				return res.body;
 			});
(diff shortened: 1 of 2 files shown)


src/api_client/api_resource.js
```diff
@@ -39,7 +39,15 @@ export default class APIResource {
 		const request = popsicle
 			.request(req)
 			.use(popsicle.plugins.parse(['json', 'urlencoded']))
-			.then(res => res.body);
+			.then(res => {
+				if (res.status >= 300) {
+					if (res.body && res.body.message) {
+						throw new Error(res.body.message);
+					}
+					throw new Error('An unknown error has occurred.');
```
Tosch110: How about `An unknown error in the API response has occured.` Maybe also provide the statusCode so that users have some insight?

shuse2: I think the phrase `An unknown error has occurred.` is common, and since the user is doing the request function, he should know it is an API response error.
but i think it's good to have status code

PR #627, opened 2018-03-26
Make `signData` function atomic - Closes #626

### LINKED ISSUE #626: Make signData function more atomic
### Description
I propose to move `getPrivateAndPublicKeyBytesFromPassphrase` out from `signData` function, and pass `privateKey` dircetrly to the `signData` function instead of `passphrase`. Like in `verifyData` which expects `publicKey` instead of passphrase. 

**Expected behavior:**
`signData` expect `data` and `privateKey ` as parameters

**Actual behavior:**
`signData` expect `data` and `passphrase ` as parameters

**Reproduces how often:**
Always

### Versions
`lisk-js 1.0.0`

alepop:
### What was the problem?
`signData` function expect` passphrase` instead of `privateKey`
### How did I fix it?
get `privateKey` in the previous function
### How to test it?
run test

### Review checklist

* The PR solves #626 
* Linting passes
* Tests pass


alepop:
Updated code based on https://github.com/LiskHQ/lisk-js/issues/626#issuecomment-378592539 @willclarktech

src/cryptography/sign.js
```diff
@@ -172,12 +172,19 @@ export const signAndPrintMessage = (message, passphrase, secondPassphrase) => {
 	return printSignedMessage(signedMessage);
 };
 
-export const signData = (data, passphrase) => {
+export const signDataWithPassphrase = (data, passphrase) => {
 	const { privateKey } = getPrivateAndPublicKeyBytesFromPassphrase(passphrase);
 	const signature = naclInstance.crypto_sign_detached(data, privateKey);
```
webmaster128: `signDataWithPassphrase` should use `signDataWithPrivateKey` as it' implementation to avoid duplicating all the logic that comes after the `privateKey` is known, e.g.

```
export const signDataWithPassphrase = (data, passphrase) => {
 	const { privateKey } = getPrivateAndPublicKeyBytesFromPassphrase(passphrase);
        return signDataWithPrivateKey(data, privateKey);
}
```

src/cryptography/sign.js
```diff
@@ -172,12 +172,19 @@ export const signAndPrintMessage = (message, passphrase, secondPassphrase) => {
...
 	return bufferToHex(signature);
 };
 
+export const signDataWithPrivateKey = (data, privateKey) => {
+	const signature = naclInstance.crypto_sign_detached(data, privateKey);
+	return bufferToHex(signature);
+};
+
+export const signData = signDataWithPassphrase;
```
webmaster128: I think this `signData` adds more confusion than it is helping. You never know if it was the password or the key version and you have multiple methods doing the same thing. Especially with no strict types in place, it is too easy to mix up methods and input types.

Isn't 1.0.0 a great chance to break the existing function and add to the migration guide:

* signData was renamed to signDataWithPassphrase

src/cryptography/sign.js
```diff
@@ -172,12 +172,19 @@ export const signAndPrintMessage = (message, passphrase, secondPassphrase) => {
...
 };
 
-export const signData = (data, passphrase) => {
+export const signDataWithPassphrase = (data, passphrase) => {
 	const { privateKey } = getPrivateAndPublicKeyBytesFromPassphrase(passphrase);
 	const signature = naclInstance.crypto_sign_detached(data, privateKey);
 	return bufferToHex(signature);
 };
 
+export const signDataWithPrivateKey = (data, privateKey) => {
```
webmaster128: For symmetry reasons, I think `privateKey` should have the same type `publicKey` in `verifyData`, i.e. a hex string.

src/cryptography/sign.js
alepop: @webmaster128 yes, make sense.

src/cryptography/sign.js
alepop: It is up to core developer, but I agree with your decision. What do you think @willclarktech, @Tosch110?

src/cryptography/sign.js
alepop: @webmaster128  thanks 👍

alepop: @webmaster128 in this case if `signDataWithPassphrase` use `signDataWithPrivateKey ` for data signing you need to convert `privateKey` to hex in `signDataWithPassphrase` function and convert it back to `Buffer` in the `signDataWithPrivateKey` because `crypto_sign_detached` expect Buffer in arguments not a hex. I think it a redundant logic.

webmaster128: Buffer is probably the better type for signing as well as verifying. So the question is how much symmetry is valued over backward compatibility. Anyway I don't have a strong opinion here

test/cryptography/sign.js
```diff
@@ -301,6 +302,19 @@ ${defaultSecondSignature}
 		});
 	});
```
willclarktech: Could you duplicate the `signData` tests for `signDataWithPassphrase`? We need to check the exported function is working as intended.

alepop: @willclarktech done

test/cryptography/sign.js
```diff
@@ -301,6 +303,35 @@ ${defaultSecondSignature}
...
 
+	describe('#ssignDataWithPassphrase', () => {
```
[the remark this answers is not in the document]
alepop: fixed

Issue #640, opened 2018-04-11
Offline signing multisig related transactions


diego-G:
### Description

After `0.9.13`, Core has the feature of signing a multisig related transaction in an offline way. I open the discussion to decide if it is in interesting feature to provide on Lisk-JS/Lisk-elements. 

**Actual behavior:**

Core accepts multisig related transactions signed offline. 

### Versions

`^0.9.13`

### Additional Information

PR https://github.com/LiskHQ/lisk/pull/1848 assures the feature works properly.

PR #641, opened 2018-04-11
Add a feature to sign transaction for multi-signature 

shuse2:
### What was the problem?
In Lisk core, we have `/signature` endpoint, but there was no direct way to sign a transaction.

### How did I fix it?
Add a `signMultisignature` function.

### How to test it?
```
elements.transaction.signMultisignature(transaction, passphrase)
```
### Review checklist

* The PR solves #640 
* Linting passes
* Tests pass
* Commit messages follow the

### DIFF
--- a/src/transactions/index.js
+++ b/src/transactions/index.js
@@ -18,12 +18,7 @@ import registerDelegate from './2_register_delegate';
 import castVotes from './3_cast_votes';
 import registerMultisignature from './4_register_multisignature_account';
 import createDapp from './5_create_dapp';
-<<<<<<< HEAD
-=======
-import transferIntoDapp from './6_transfer_into_dapp';
-import transferOutOfDapp from './7_transfer_out_of_dapp';
 import createSignatureObject from './create_signature_object';
->>>>>>> :recycle: Change to have separate file and fix test
 import * as utils from './utils';
 
 export default {
--- a/src/transactions/index.js
+++ b/src/transactions/index.js
@@ -16,7 +16,9 @@ import transfer from './0_transfer';
 import registerSecondPassphrase from './1_register_second_passphrase';
 import registerDelegate from './2_register_delegate';
 import castVotes from './3_cast_votes';
-import registerMultisignature from './4_register_multisignature_account';
+import registerMultisignature, {
+	signMultisignature,
+} from './4_register_multisignature_account';
 import createDapp from './5_create_dapp';
 import * as utils from './utils';
 
@@ -26,6 +28,7 @@ export default {
 	registerDelegate,
 	castVotes,
 	registerMultisignature,
+	signMultisignature,
 	createDapp,
 	utils,
 };
--- a/src/transactions/index.js
+++ b/src/transactions/index.js
@@ -16,10 +16,14 @@ import transfer from './0_transfer';
 import registerSecondPassphrase from './1_register_second_passphrase';
 import registerDelegate from './2_register_delegate';
 import castVotes from './3_cast_votes';
-import registerMultisignature, {
-	signMultisignature,
-} from './4_register_multisignature_account';
+import registerMultisignature from './4_register_multisignature_account';
 import createDapp from './5_create_dapp';
+<<<<<<< HEAD
+=======
+import transferIntoDapp from './6_transfer_into_dapp';
+import transferOutOfDapp from './7_transfer_out_of_dapp';
+import createSignatureObject from './create_signature_object';
+>>>>>>> :recycle: Change to have separate file and fix test
 import * as utils from './utils';
 
 export default {
@@ -28,7 +32,7 @@ export default {
 	registerDelegate,
 	castVotes,
 	registerMultisignature,
-	signMultisignature,
+	createSignatureObject,
 	createDapp,
 	utils,
 };
--- a/test/transactions/4_register_multisignature_account.js
+++ b/test/transactions/4_register_multisignature_account.js
@@ -13,461 +13,357 @@
  *
  */
 import cryptography from 'cryptography';
-import registerMultisignatureAccount, {
-	signMultisignature,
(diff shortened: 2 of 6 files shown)


src/transactions/index.js
```diff
@@ -16,7 +16,9 @@ import transfer from './0_transfer';
 import registerSecondPassphrase from './1_register_second_passphrase';
 import registerDelegate from './2_register_delegate';
 import castVotes from './3_cast_votes';
-import registerMultisignature from './4_register_multisignature_account';
+import registerMultisignature, {
+	signMultisignature,
```
willclarktech: Does this not belong more in `utils` alongside signing transactions normally? `4_register_multisignature_account` exports a transaction creator.

test/transactions/index.js
```diff
@@ -46,6 +46,12 @@ describe('expect(transaction)s', () => {
 				.and.be.a('function');
 		});
 
+		it('to have the sign multisignature expect(transaction) function', () => {
```
willclarktech: I think the text went a bit strange here (also the following existing test).

test/transactions/4_register_multisignature_account.js
```diff
@@ -13,357 +13,461 @@
...
+				timestamp: 59353522,
+				type: 0,
+				fee: '10000000',
+				recipientPublicKey: null,
+				asset: {},
+				signature:
+					'b84b95087c381ad25b5701096e2d9366ffd04037dcc941cd0747bfb0cf93111834a6c662f149018be4587e6fc4c9f5ba47aa5bbbd3dd836988f153aa8258e604',
+				id: '0000000000000000000',
+			};
+			const passphrase = '';
```
willclarktech: I think a non-empty passphrase could make it clearer what's invalid about this situation. Preferably we would create a valid transaction for a plausible passphrase, and then modify it in some minimal way, e.g. change the last couple of bytes of the signature.

test/transactions/4_register_multisignature_account.js
```diff
@@ -13,357 +13,461 @@
...
+			it('should throw an Error', () => {
+				return expect(
+					signMultisignature.bind(null, transaction, passphrase),
+				).to.throw();
```
willclarktech: Can we expect a particular message?

test/transactions/4_register_multisignature_account.js
```diff
@@ -13,357 +13,461 @@
...
+				signature:
+					'b84b95087c381ad25b5701096e2d9366ffd04037dcc941cd0747bfb0cf93111834a6c662f149018be4587e6fc4c9f5ba47aa5bbbd3dd836988f153aa8258e604',
+				id: '3694188453012384790',
+			};
+
+			it('should throw an Error if passphrase is number', () => {
+				const passphrase = 1;
+				return expect(
+					signMultisignature.bind(null, transaction, passphrase),
+				).to.throw();
```
willclarktech: As above: message?

test/transactions/4_register_multisignature_account.js
```diff
@@ -13,357 +13,461 @@
...
+				return expect(signature.transactionId).to.equal(transaction.id);
 			});
 
-			it('should not have the signature', () => {
-				return expect(registerMultisignatureTransaction).not.to.have.property(
-					'signature',
-				);
+			it('should have the corresponding public key with the passphrase', () => {
+				return expect(signature.publicKey).to.equal(account.publicKey);
 			});
```
willclarktech: We should have an expectation about the `signature` property too.

test/transactions/4_register_multisignature_account.js
```diff
@@ -13,357 +13,461 @@
...
 
-			it('should not have the id', () => {
-				return expect(registerMultisignatureTransaction).not.to.have.property(
-					'id',
-				);
+			it('should have all the keys for signature type', () => {
```
willclarktech: This case should be redundant once we have test cases for each property.

src/transactions/index.js
shuse2: I think having in the top layer make more sense rather than having in `utils` because imo this creating signature object is the same layer as creating transaction.

shuse2: I think it might even better to have it on separate file

willclarktech: Agree with top level, but definitely a separate file I would say.

test/transactions/index.js
```diff
@@ -16,52 +16,58 @@ import transaction from 'transactions';
 
 describe('expect(transaction)s', () => {
```
willclarktech: Thanks for fixing, but this should also probably just be `transactions`.

test/transactions/create_signature_object.js
```diff
@@ -0,0 +1,100 @@
...
+				'Unsupported data format. Currently only Buffers or `hex` and `utf8` strings are supported.',
+			);
+		});
+	});
+
+	describe('when valid transaction and passphrase is used', () => {
+		let signatureObject;
+		beforeEach(() => {
+			signatureObject = createSignatureObject(transaction, account.passphrase);
+			return Promise.resolve(true);
```
willclarktech: Elsewhere we just `return Promise.resolve()`. Do you have a preference for including `true`?

test/transactions/create_signature_object.js
```diff
@@ -0,0 +1,100 @@
...
+			return expect(signatureObject.transactionId).to.equal(transaction.id);
+		});
+
+		it('should have the corresponding public key with the passphrase', () => {
+			return expect(signatureObject.publicKey).to.equal(account.publicKey);
+		});
+
+		it('should have non-empty hex string signature', () => {
+			expect(signatureObject.signature).not.to.be.empty;
+			return expect(signatureObject.signature).to.be.hexString;
```
willclarktech: Why not make an expectation against a specific signature string?

PR #647, opened 2018-04-17
Change nethash to optional - Closes #646

### LINKED ISSUE #646: APIClient does not require detailed information
### Description

APIClient used to require a lot of information, such as
- nethash
- version
- minVersion

However, none of those header information are required in the 1.0.0 version

### Steps to Reproduce
Call any 1.x.x API endpoint through browser.

### Versions
1.0.0

shuse2:
### What was the problem?
Although nethash was not required from `core` API, it was required in the APIClient.

### How did I fix it?
Remove `nethash` from required constructor input, and read it from optional field.

### How to test it?
```
const client = elements.APIClient(['http://localhost:4000']);
client.node.getConstants().then(console.log);
```

### Review checklist

* The PR solves #646
* Linting passes
* Tests pass
* Commit messages follow the

### DIFF
--- a/src/api_client/api_client.js
+++ b/src/api_client/api_client.js
@@ -40,6 +40,7 @@ const defaultOptions = {
 
 const commonHeaders = {
 	'Content-Type': 'application/json',
+	os: 'lisk-js-api',
 };
 
 export default class APIClient {
@@ -61,21 +62,21 @@ export default class APIClient {
 	static createMainnetAPIClient(options) {
 		return new APIClient(
 			MAINNET_NODES,
-			Object.assign({}, options, { nethash: MAINNET_NETHASH }),
+			Object.assign({}, { nethash: MAINNET_NETHASH }, options),
 		);
 	}
 
 	static createTestnetAPIClient(options) {
 		return new APIClient(
 			TESTNET_NODES,
-			Object.assign({}, options, { nethash: TESTNET_NETHASH }),
+			Object.assign({}, { nethash: TESTNET_NETHASH }, options),
 		);
 	}
 
 	static createBetanetAPIClient(options) {
 		return new APIClient(
 			BETANET_NODES,
-			Object.assign({}, options, { nethash: BETANET_NETHASH }),
+			Object.assign({}, { nethash: BETANET_NETHASH }, options),
 		);
 	}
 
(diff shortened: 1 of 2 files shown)


src/api_client/api_client.js
```diff
@@ -69,29 +59,36 @@ export default class APIClient {
 	}
 
 	static createMainnetAPIClient(options) {
-		return new APIClient(MAINNET_NODES, MAINNET_NETHASH, options);
+		return new APIClient(
+			MAINNET_NODES,
+			Object.assign({}, options, { nethash: MAINNET_NETHASH }),
```
willclarktech: Is this definitely the right order? Could there be a use case where somebody wants to use this with a custom nethash? (In which case `Object.assign({}, { nethash: MAINNET_NETHASH }, options)` would be better.)

src/api_client/api_client.js
```diff
@@ -69,29 +59,36 @@ export default class APIClient {
...
-		if (typeof nethash !== 'string' || nethash === '') {
-			throw new Error('APIClient requires nethash for initialization.');
-		}
-
 		const options = Object.assign({}, defaultOptions, providedOptions);
 
-		this.headers = getHeaders(nethash, options.version, options.minVersion);
+		this.headers = options.nethash
+			? Object.assign({}, commonHeaders, { nethash: options.nethash })
+			: commonHeaders;
```
willclarktech: We should have a fresh object here in case people override `myClientInstance.headers` and don't expect it to affect other clients.

src/api_client/api_client.js
```diff
@@ -69,29 +59,36 @@ export default class APIClient {
...
+		);
 	}
 
-	initialize(nodes, nethash, providedOptions = {}) {
+	initialize(nodes, providedOptions = {}) {
 		if (!Array.isArray(nodes) || nodes.length <= 0) {
 			throw new Error('APIClient requires nodes for initialization.');
 		}
 
-		if (typeof nethash !== 'string' || nethash === '') {
```
willclarktech: We might still want a minimal check like this even if it's provided in options. E.g. somebody passes an array...

test/api_client/api_client.js
```diff
@@ -38,18 +38,11 @@ describe('APIClient module', () => {
 	];
 	const defaultHeaders = {
 		'Content-Type': 'application/json',
-		nethash: mainnetHash,
-		os: 'lisk-js-api',
```
willclarktech: This might be useful for analytics.

slaweet: Yes, I was thinking about analytics recently and this might be very useful.

src/api_client/api_client.js
```diff
@@ -84,11 +85,17 @@ export default class APIClient {
 			throw new Error('APIClient requires nodes for initialization.');
 		}
 
+		if (typeof providedOptions !== 'object' || Array.isArray(providedOptions)) {
+			throw new Error(
+				'APIClient requires the second parameter to be an object.',
```
willclarktech: We should make clear that it's optional. E.g. `APIClient takes an optional object as the second parameter.`?

Issue #636, opened 2018-04-10
Expose transactions constants


willclarktech:
**Expected behavior:** Users should have access to transactions constants.

**Actual behavior:** They aren't exported from the transactions module.

### Versions

1.0.0

PR #650, opened 2018-04-19
Refactor constants location and export path - Closes 636

shuse2:
### What was the problem?
transaction constant was not exposed, and constants that are used only in APIClient was in general constants

### How did I fix it?
Expose constants from transaction
Expose APIClient constants

### How to test it?
```
elements.transaction.constants;
elements.APICleint.constants;
```

### Review checklist

* The PR solves #636 
* Linting passes
* Tests pass
* Commit messages follow the

### DIFF
--- a/src/api_client/constants.js
+++ b/src/api_client/constants.js
@@ -16,3 +16,25 @@ export const GET = 'GET';
 export const POST = 'POST';
 export const PUT = 'PUT';
 export const API_RECONNECT_MAX_RETRY_COUNT = 3;
+
+export const BETANET_NETHASH =
+	'ef3844327d1fd0fc5785291806150c937797bdb34a748c9cd932b7e859e9ca0c';
+export const BETANET_NODES = [
+	'http://94.237.42.109:5000',
+	'http://83.136.252.99:5000',
+];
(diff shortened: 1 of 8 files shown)


src/api_client/constants.js
```diff
@@ -16,3 +16,25 @@ export const GET = 'GET';
 export const POST = 'POST';
 export const PUT = 'PUT';
 export const API_RECONNECT_MAX_RETRY_COUNT = 3;
+
+export const BETANET_NETHASH =
```
willclarktech: Do nethashes definitely belong in `api_client`? There's an argument to be made that they're part of the protocol (or at least that they're general enough to warrant being placed in the global constants module).

test/api_client/api_client.js
```diff
@@ -141,6 +141,12 @@ describe('APIClient module', () => {
 		});
 	});
 
+	describe('#constants', () => {
+		it('should return APIClient instance', () => {
```
willclarktech: `should expose API constants`

src/api_client/constants.js
shuse2: yea true, but is there any other place that can potentially use the nethash?

PR #653, opened 2018-04-24
Add User-Agent header to API requests - Closes #651

### LINKED ISSUE #651: Change os header in api
**Expected behavior:** We should use HTTP headers properly.

**Actual behavior:** We send `lisk-js-api` as an `os` header so the server can spot our requests. This isn't very semantic.

willclarktech:
### What was the problem?

We set an `os` header for some reason.

### How did I fix it?

Set `User-Agent` instead.

### How to test it?

Send requests to a node and check the headers

### Review checklist

* The PR solves #651 
* Linting passes
* Tests pass
* Commit messages follow the


src/api_client/api_client.js
```diff
@@ -38,7 +38,7 @@ const defaultOptions = {
 
 const commonHeaders = {
 	'Content-Type': 'application/json',
-	os: 'lisk-js-api',
+	'User-Agent': 'lisk-js/1.0',
```
nazarhussain: @willclarktech This change is good but should't we change it to **LiskElements**? Or that will be done in one transition? 

Also one change I could suggest here. The standard `User-Agent` sting as following pattern: 

```
[ProductName]/[version] ([system and browser information]) [platform] ([platform details]) [extensions].
```

So here I believe; 

```
ProductName: LiskElements
Version: 1.0
System Information: If we can add we should 
Platform: We can add this information easily 
```

One last thing around all `UserAgents` I see every one uses `CamelCase` for Product Name, should not we follow it as well? e.g. Mozilla, AppleWebKit, Googlebot

If you think its not a browser, even then custom agents follow the same pattern.e.g. 

```
Googlebot/2.1 (+http://www.google.com/bot.html)
```

slaweet: I would consider getting the version number from `package.json` (crop to only major and minor version) to keep it up-to-date.

shuse2: I don't think we should add `lisk-js` version number.
Lisk-js APIClient is just a library and not actual client imo

nazarhussain: @shuse2 That APIClient is based on an engine/library/module `lisk-js`which have a specific version number. And for debugging and analytics its a valuable information.

willclarktech: Let's arrange a meeting to discuss this.

src/api_client/api_client.js
```diff
@@ -37,8 +38,25 @@ const defaultOptions = {
...
+		process.env.LC_ALL ||
+		process.env.LC_MESSAGES ||
+		process.env.LANG ||
+		process.env.LANGUAGE;
+	const systemInformation = `${os.platform()} ${os.release()}; ${os.arch()}${
+		locale ? `; ${locale}` : ''
+	}`;
+	return `${name}/${version} (${engine}) ${liskJSInformation} ${
+		systemInformation
+	}`;
```
nazarhussain: There should be a `,` between lisk and system information.

willclarktech: Do you have any examples of this as a standard? It doesn't appear anywhere here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent

PR #656, opened 2018-04-24
Update authenticated passphrase encryption - Closes #649

willclarktech:
### What was the problem?

Passphrase encryption needed to be updated to align with core.

### How did I fix it?

I updated it.

### How to test it?

`npm t`, or try encrypting a passphrase, storing it in the `config.json` of a node and start the node.

### Review checklist

* The PR solves #649 
* All new code is covered with unit tests
* All new code was formatted with Prettier
* Linting passes
* Tests pass
* Commit messages follow the
	[commit guidelines](CONTRIBUTING.md#git-commit-messages)
* Documentation has been added/updated

PR #658, opened 2018-04-25
Change GCM iv to default

karek314:
GCM iv by default should be 12 bytes. Additionally it appears that's inconsistency between core and lisk-js. I've seen 12bytes iv's in core test configs.


willclarktech:
Will be addressed as part of https://github.com/LiskHQ/lisk-js/pull/656

karek314:
My mistake then, i've missed this PR.

Issue #418, opened 2017-10-12
Make Lisk Elements a monorepo


Tosch110:
Create a monorepo using lerna, resulting in several npm packages that can be installed seperately.

Aimed for Lisk-JS 1.1

alepop:
@Tosch110 `lisk-js` will be split into a bunch of packages? why use monorepo for a small library?

willclarktech:
@alepop 
The reasoning is that the functionality in the various source modules (`api`, `constants`, `crypto`, `transactions`) are pretty distinct (or should be). Splitting them into separate packages will allow projects (e.g. going forward sidechain applications) to be more selective about which parts they want. We'll probably still provide a package which bundles them all together in case somebody needs everything.

Why a monorepo? It simplifies various things:
- less restructuring required for the migration
- synchronising release schedules
- common dependency management
- easier to see cross-dependency relationships
- easier to share common resources such as (upcoming) flow types

alepop:
@willclarktech I understood your point about using only necessary parts of `liks-js.` agree with it 👍 I had tried to use `lerna` in my project. Seems simple and straightforward.

willclarktech:
Yeah, I'm aware of at least one circular dependency we'll have to deal with, but the hope is that the process will lead to us placing code in more logical locations because of it. I'd be interested to hear more about the issues that arose with types?

willclarktech:
Also errors maybe?

willclarktech:
Remaining issues will be opened separately. E.g. Jenkins.

PR #666, opened 2018-05-07
Convert to multi-package monorepo

willclarktech:
~~Note: This is based off `1.0.0` but is only open now for discussion.~~

### What was the problem?

We had only one package encompassing all Elements functionality.

Still outstanding as of 9/5:
- Browser tests
- Jenkins/CI
- Coverage integration (how to gather several reports into one) and coveralls
- Independent versioning
- Proper individual READMEs

### How did I fix it?

Converted it to a multi-package monorepo.

Some info:
- `.babelrc.tmpl` in root directory provides a default babel config. So far all the packages just symlink to this. `.babelrc` does support an `extends` property (see https://github.com/babel/babel/issues/2383) but it's undocumented and certainly not exactly the same functionality as e.g. `.eslintrc`'s `extends`. Any package requiring a different config can at worst use a distinct file.
- `.editorconfig` is only in the root directory as it shouldn't vary by package.
- `.eslintignore.tmpl` in root directory is a default `.eslintignore` file for symlink convenience.
- `.eslintrc.json` extends our config along with some temporary overrides before they're published as part of our config. It might make more sense to move these overrides into the individual packages rather than extending a root config as in the current implementation.
- `.eslintrc.test.json` contains common test config for eslint.
- `.gitignore` in the root directory covers all the normal cases. The `lisk-elements` package has a special structure so it introduces a `.gitignore` of its own.
- `.lintstagedrc.json` is to do with committing, so stays in the root (along with the `precommit` and `prepush` husky hooks). Hence the dev dependencies in the root package.
- `.npmignore.tmpl` in root directory provides a default `.npmignore` file for symlink convenience.
- `.npmrc` in root directory the same (except it's an actual `rc` file, not just a template: the root has dependencies of its own).
- `.nvmrc`, like `.editorconfig` affects the entire repo.
- `.nycrc` should probably be a `.tmpl` (I can change it next week).
- `.prettierrc.json` is used in the root itself, and for symlinking in individual packages.
- `Jenkinsfile`s are untouched but will remain in the root.
- `lerna.json` belongs in the root,
- `LICENSE` in the root has been copied into the packages (`npm pack` won't follow symlinks).
- `mocha.opts.tmpl` provides a default set of mocha options for symlink convenience.
- `README.md` has not been updated, but placeholder files have been put into each package.

### How to test it?

In root directory `npm t`

### Review checklist

* The PR solves #418 
* Linting passes
* Tests pass
* Commit messages follow the


shuse2:
One note for this,
in order to run test with dependencies, we need to do `npm run build` first.

shuse2:
Also, lerna should be set to independent mode, which is not default

Issue #673, opened 2018-05-22
Add prefix to signed message


alepop:
### Description
Lisk signed message don't have prefix bytes in the header. This may cause next type of errors:
1) It can sign whatever you give to message signing function, even a transaction digest.
2) It can verify a signature from a separate coin with the same curve.

### Steps to Reproduce

1. Sign message with another ed25519 curve implementation
2. Verify it with `lisk-elements`

**Expected behavior:**

Message signing adds specific for Lisk prefix bytes to the signed message.

**Actual behavior:** 

No prefix

**Reproduces how often:** 

Constantly

### Versions

`1.0.0`

### Additional Information
From Ethereum repos:

> By adding a prefix to the message makes the calculated signature recognisable as an Ethereum
> specific signature. This prevents misuse where a malicious DApp can sign arbitrary data (e.g. transaction) and use the signature to impersonate the victim.

willclarktech:
@alepop Thanks for this, we'll arrange a meeting to establish the best way forward here as this has implications for front end applications/backward compatibility.

alepop:
@willclarktech also, all most popular coins use a hash algorithm for signed messages before sign it. 
The whole hash looks similar to this before passing to sign function:
```
h = hash_algorithm_function 
result = h(varint(len(prefix))) + h(prefix) + h(varint(len(message))) + h(message)
```
Then `result` passed to the sign function. What about implement something like this with sha256?

willclarktech:
## Update from meeting today

This is something we want to address and we will include a solution in the release for 1.0.0 since we have breaking changes in the elements function then already. Users have a responsibility to check what they are signing, but in the interests of usability this can be an important barrier to a certain class of user manipulation attempts.

Since signatures are usually intended for single-use situations, and this is a security-related issue, we aren't planning to provide backwards compatibility.

### Issues to investigate

1. Whether we should be hashing the message before signing and why?
1. Whether there's an advantage to including the lengths of the prefix and/or message itself?
1. How this relates to https://github.com/LiskHQ/lisk/issues/1610 because poor coordination between these two issues could reintroduce replay attacks (e.g. if we opt for a prefix-based solution to both and the prefixes are related to each other in the wrong ways).

IkerAlus:
Some conclusions after studying the issue:

### Regarding 1: 
[...]

PR #674, opened 2018-05-22
Add prefix bytes to signed message - Closes#673

alepop:
### What was the problem?
Lisk signed message doesn't have a header prefix

### How did I fix it?
Calculate message signature with this formula:
~~sign (Lisk Signed Message:\n + message length + message)~~
```

digest = sha256(sha256(
  varint(len(prefix)) + prefix + varint(len(message)) + message
))

sign(dugest)
```

### How to test it?
run tests

### Review checklist

* The PR solves #673 
* Linting passes
* Tests pass
* Commit messages follow the

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,8 +69,9 @@
     "bip39": "=2.4.0",
     "browserify-bignum": "=1.3.0-2",
     "ed2curve": "=0.2.1",
-    "js-nacl": "LiskHQ/js-nacl#6dc1417",
-    "popsicle": "=9.1.0"
+	"js-nacl": "LiskHQ/js-nacl#6dc1417",
+	"popsicle": "=9.1.0",
+	"varuint-bitcoin": "=1.1.0"
   },
   "devDependencies": {
     "babel-cli": "=6.24.1",
--- a/package.json
+++ b/package.json
@@ -69,9 +69,9 @@
     "bip39": "=2.4.0",
     "browserify-bignum": "=1.3.0-2",
     "ed2curve": "=0.2.1",
-	"js-nacl": "LiskHQ/js-nacl#6dc1417",
-	"popsicle": "=9.1.0",
-	"varuint-bitcoin": "=1.1.0"
+    "js-nacl": "LiskHQ/js-nacl#6dc1417",
+    "popsicle": "=9.1.0",
+    "varuint-bitcoin": "=1.1.0"
   },
   "devDependencies": {
     "babel-cli": "=6.24.1",
--- a/src/cryptography/sign.js
+++ b/src/cryptography/sign.js
@@ -12,6 +12,8 @@
  * Removal or modification of this copyright notice is prohibited.
  *
  */
+import { encode as encodeVarInt } from 'varuint-bitcoin';
+import hash from './hash';
 import { hexToBuffer, bufferToHex } from './convert';
 import { getPrivateAndPublicKeyBytesFromPassphrase } from './keys';
 
@@ -24,8 +26,25 @@ const signatureHeader = createHeader('SIGNATURE');
 const secondSignatureHeader = createHeader('SECOND SIGNATURE');
 const signatureFooter = createHeader('END LISK SIGNED MESSAGE');
 
-export const signMessageWithPassphrase = (message, passphrase) => {
+const SIGNED_MESSAGE_PREFIX = Buffer.from('Lisk Signed Message:\n', 'utf8');
+const SIGNED_MESSAGE_PREFIX_LENGTH = encodeVarInt(SIGNED_MESSAGE_PREFIX.length);
+
+export const digestMessage = message => {
 	const msgBytes = Buffer.from(message, 'utf8');
+	const msgLenBytes = encodeVarInt(message.length);
+	const dataBytes = Buffer.concat(
+		[
+			SIGNED_MESSAGE_PREFIX_LENGTH,
+			SIGNED_MESSAGE_PREFIX,
+			msgLenBytes,
+			msgBytes,
+		],
+	);
+	return hash(hash(dataBytes));
+};
+
+export const signMessageWithPassphrase = (message, passphrase) => {
+	const msgBytes = digestMessage(message);
 	const { privateKey, publicKey } = getPrivateAndPublicKeyBytesFromPassphrase(
 		passphrase,
 	);
@@ -43,7 +62,7 @@ export const verifyMessageWithPublicKey = ({
 	signature,
 	publicKey,
 }) => {
-	const msgBytes = Buffer.from(message, 'utf8');
+	const msgBytes = digestMessage(message);
 	const signatureBytes = hexToBuffer(signature);
 	const publicKeyBytes = hexToBuffer(publicKey);
 
@@ -67,7 +86,7 @@ export const signMessageWithTwoPassphrases = (
 	passphrase,
 	secondPassphrase,
 ) => {
-	const msgBytes = Buffer.from(message, 'utf8');
+	const msgBytes = digestMessage(message);
 	const keypairBytes = getPrivateAndPublicKeyBytesFromPassphrase(passphrase);
 	const secondKeypairBytes = getPrivateAndPublicKeyBytesFromPassphrase(
 		secondPassphrase,
@@ -98,7 +117,7 @@ export const verifyMessageWithTwoPublicKeys = ({
 	publicKey,
 	secondPublicKey,
 }) => {
-	const messageBytes = Buffer.from(message, 'utf8');
+	const messageBytes = digestMessage(message);
 	const signatureBytes = hexToBuffer(signature);
 	const secondSignatureBytes = hexToBuffer(secondSignature);
 	const publicKeyBytes = hexToBuffer(publicKey);
--- a/src/cryptography/sign.js
+++ b/src/cryptography/sign.js
@@ -13,6 +13,7 @@
  *
  */
 import { encode as encodeVarInt } from 'varuint-bitcoin';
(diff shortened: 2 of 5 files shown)


src/cryptography/sign.js
```diff
@@ -24,8 +24,16 @@ const signatureHeader = createHeader('SIGNATURE');
 const secondSignatureHeader = createHeader('SECOND SIGNATURE');
 const signatureFooter = createHeader('END LISK SIGNED MESSAGE');
 
-export const signMessageWithPassphrase = (message, passphrase) => {
+const SIGNED_MESSAGE_HEADER = Buffer.from('Lisk Signed Message:\n', 'utf8');
+
+export const messageDigest = message => {
 	const msgBytes = Buffer.from(message, 'utf8');
+	const msgLenBytes = Buffer.from(String(message.length), 'utf8');
```
willclarktech: This should be a set number of bytes or delimited in some way.

willclarktech:
Suspended until the strategy for this issue has been clarified.

src/cryptography/sign.js
alepop: Not necessary, but other cryptos use this way. And instead of `String(message.length)` they use `varint`. Bitcoin use something similar to this https://github.com/LiskHQ/lisk-elements/issues/673#issuecomment-391617664

willclarktech: Do you know why they use `varint`? It seems like it would make it possible for people to construct messages that are different but achieve the same `varint(length) + message`.

alepop: >A VarInt (variable integer) is a field used in transaction data to indicate the
>number of upcoming fields, or the length of an upcoming field.

http://learnmeabitcoin.com/glossary/varint

alepop: @willclarktech should I update this code with `sha256` and `varint` for message signing, or I need to wait until Iker make research?

willclarktech: @alepop As mentioned [here](https://github.com/LiskHQ/lisk-elements/issues/673#issuecomment-391671006) Iker has a few issues to look into before we can move ahead with a solution.

alepop:
Update code based on the @IkerAlus proposal. Now message digest is calculated this way:
```
sha256(sha256(prefix + message))
```

src/cryptography/sign.js
```diff
@@ -24,8 +25,35 @@ const signatureHeader = createHeader('SIGNATURE');
...
+		buf.writeUInt16LE(n, 1);
+	} else {
+		buf = Buffer.alloc(1 + 4);
+		buf.writeUInt8(254, 0);
+		buf.writeUInt32LE(n, 1);
+	}
+	return buf;
+};
+
+export const messageDigest = message => {
```
willclarktech: Prefer a function name beginning with a verb. E.g. `digestMessage` or `createMessageDigest`.

src/cryptography/sign.js
```diff
@@ -24,8 +25,35 @@ const signatureHeader = createHeader('SIGNATURE');
 const secondSignatureHeader = createHeader('SECOND SIGNATURE');
 const signatureFooter = createHeader('END LISK SIGNED MESSAGE');
 
-export const signMessageWithPassphrase = (message, passphrase) => {
+const SIGNED_MESSAGE_HEADER = Buffer.from('Lisk Signed Message:\n', 'utf8');
+
+export const writeVarInt = n => {
```
willclarktech: `write` is a confusing term here. Probably `encode` is more accurate?

test/cryptography/sign.js
```diff
@@ -116,6 +122,25 @@ ${defaultSecondSignature}
 			});
 	});
 
+	describe('#writeVarIt', () => {
```
willclarktech: Typo

src/cryptography/sign.js
```diff
@@ -24,8 +25,35 @@ const signatureHeader = createHeader('SIGNATURE');
...
+	let buf;
+	if (n < 253) {
```
willclarktech: To the extent we write our own function here (but see comment below), I think it's much easier to understand what's going on here if we specify this in hex. E.g. `if (n < 0xfd) { ... }`.

src/cryptography/sign.js
willclarktech: This is one of the few occasions where I think a comment would be really helpful. How about linking here? https://en.bitcoin.it/wiki/Protocol_documentation#Variable_length_integer

test/cryptography/sign.js
```diff
@@ -116,6 +122,25 @@ ${defaultSecondSignature}
...
+		it('should calculate correct value for n < 253', () => {
```
willclarktech: As above re hex representations vs magic numbers.

test/cryptography/sign.js
```diff
@@ -116,6 +122,25 @@ ${defaultSecondSignature}
...
+		it('should calculate correct value for n > 65536', () => {
+			return expect(writeVarInt(65537)).to.be.eql(Buffer.from([254, 1, 0, 1, 0]));
+		});
+	});
+
+	describe('#messageDigest', () => {
+		it('should create message digest', () => {
+			const msgBytes = messageDigest(defaultMessage);
+			return expect(msgBytes).to.be.eql(defaultMessageDigest);
+		});
```
willclarktech: What about an empty message or one that's too long? Should it handle buffers?

test/cryptography/sign.js
```diff
@@ -116,6 +122,25 @@ ${defaultSecondSignature}
...
+	describe('#writeVarIt', () => {
+		it('should calculate correct value for n < 253', () => {
+			return expect(writeVarInt(5)).to.be.eql(Buffer.from([5]));
+		});
+		it('should calculate correct value for n < 65536', () => {
+			return expect(writeVarInt(300)).to.be.eql(Buffer.from([253, 44, 1]));
+		});
+		it('should calculate correct value for n > 65536', () => {
+			return expect(writeVarInt(65537)).to.be.eql(Buffer.from([254, 1, 0, 1, 0]));
+		});
```
willclarktech: We should have test cases for `0`, a negative number, a number out of range (too large), a float, `undefined`(?), a buffer (should it handle buffers?)

src/cryptography/sign.js
```diff
@@ -24,8 +25,35 @@ const signatureHeader = createHeader('SIGNATURE');
 const secondSignatureHeader = createHeader('SECOND SIGNATURE');
 const signatureFooter = createHeader('END LISK SIGNED MESSAGE');
 
-export const signMessageWithPassphrase = (message, passphrase) => {
+const SIGNED_MESSAGE_HEADER = Buffer.from('Lisk Signed Message:\n', 'utf8');
```
willclarktech: Here we call this `HEADER` but below we call it `prefix`. We should be consistent. Since we already have a "Header" in the presentation function, I have a preference for `SIGNED_MESSAGE_PREFIX`.

test/cryptography/sign.js
```diff
@@ -116,6 +122,25 @@ ${defaultSecondSignature}
...
+	});
+
+	describe('#messageDigest', () => {
+		it('should create message digest', () => {
+			const msgBytes = messageDigest(defaultMessage);
+			return expect(msgBytes).to.be.eql(defaultMessageDigest);
+		});
+	});
```
willclarktech: Ideally we would have test cases for each length range.

test/cryptography/sign.js
alepop: I'm ok with `0`, but don't agree with negative, `float` and `undefined` because of this function accepts only length of messages which passed from the parent function. Or you see other cases? I think this is overtesting.

test/cryptography/sign.js
alepop: Which length range?

test/cryptography/sign.js
alepop: I'll add type check for `string` type. Also, I think that message length checker should be not in this function. And which message should be considered as a too long message? (what length)

willclarktech: One where the length can't be encoded as a varint. I'm OK with that check being delegated to the varint function though.

test/cryptography/sign.js
willclarktech: Fair enough.

test/cryptography/sign.js
willclarktech: 0–252, 253–65536, ...

package.json
```diff
@@ -70,7 +70,8 @@
     "browserify-bignum": "=1.3.0-2",
     "ed2curve": "=0.2.1",
     "js-nacl": "LiskHQ/js-nacl#6dc1417",
-    "popsicle": "=9.1.0"
+	"popsicle": "=9.1.0",
```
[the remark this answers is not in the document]
alepop: it is weird because indentation is fine in my package.json

src/cryptography/sign.js
```diff
@@ -24,8 +26,25 @@ const signatureHeader = createHeader('SIGNATURE');
 const secondSignatureHeader = createHeader('SECOND SIGNATURE');
 const signatureFooter = createHeader('END LISK SIGNED MESSAGE');
 
-export const signMessageWithPassphrase = (message, passphrase) => {
+const SIGNED_MESSAGE_PREFIX = Buffer.from('Lisk Signed Message:\n', 'utf8');
```
shuse2: If `'Lisk Signed Message:\n'` is an constant, shouldn't we put this in constant file?
Also, do we need line break?

alepop: All implementation for all other coins use the line break, so I think Lisk should use the same way too

alepop:
@shuse2 pr updated

willclarktech:
Thanks @alepop!

Issue #676, opened 2018-05-24
Add validate transaction method


shuse2:
### Description

Currently we have verify transaction, whether signature is correct or not, but we don't have a function to check if the schema itself is valid (without signature)

tobiaslins:
@willclarktech 
Want to throw something in here.
What about using ajv or joi for validation? 
Will be much simpler then what you did on #678 

https://github.com/hapijs/joi
https://github.com/epoberezkin/ajv

Not sure if it makes sense, but what do you think?

willclarktech:
We'll revisit this for 1.1.0 and come up with a comprehensive solution for schema checks of all useful sorts. Thanks for the resources @tobiaslins, we'll check them out when we come back to this.

willclarktech:
There's already enough in 1.1.0, we'll schedule this for 1.2.0.

willclarktech:
Closed in #734

PR #678, opened 2018-05-31
Add function to validate transactions

willclarktech:
### What was the problem?

We could verify transactions via verifying the signature, but not that they had the right shape. This made it difficult if you e.g. wanted to validate an input was a valid unsigned transaction before asking the user for their passphrase to sign it.

### How did I fix it?

Added a function to validate the schema of a transaction.

This is a WIP - submitting a PR now to get feedback on the general approach as I'm not 100% that it's the right one to take. An alternative would be to use something like z-schema, or to wait until we implement a type system in #417 .

### How to test it?

Create some valid/invalid transactions and try validating them.

### Review checklist

* The PR solves #676 
* Linting passes
* Tests pass
* Commit messages follow the


src/transactions/utils/validation.js
```diff
@@ -28,6 +44,18 @@ export const validatePublicKey = publicKey => {
...
+export const validateSignature = signature => {
+	const signatureBuffer = hexToBuffer(signature);
+	if (signatureBuffer.length !== 64) {
+		throw new Error(
+			`Signature ${
+				signature
+			} length differs from the expected 64 bytes for a public key.`,
+		);
+	}
+	return true;
```
shuse2: This return value is not parallel.
If we `return true`, i think it's better to have `return false`

src/transactions/utils/validation.js
```diff
@@ -71,3 +99,103 @@ export const validateAddress = address => {
...
+	if (![0, 1, 2, 3, 4, 5].includes(type)) {
+		throw new Error('Transaction type must be an integer between 0 and 5');
+	}
+	if (
+		typeof amount !== 'string' ||
+		!amount.match(/^[0-9]+$/) ||
+		bignum(amount).cmp(bignum(MAX_TRANSACTION_AMOUNT)) > 0
+	) {
+		throw new Error(
+			'Transaction amount must be a string integer between 0 and 18446744073709551615',
```
shuse2: It would be better to use the constant `MAX_TRANSACTION_AMOUNT`

src/transactions/utils/validation.js
```diff
@@ -71,3 +99,103 @@ export const validateAddress = address => {
...
+		timestamp > MAX_TIMESTAMP
+	) {
+		throw new Error('Transaction must include a valid timestamp');
+	}
+	try {
+		validateSignature(signature);
+	} catch (error) {
+		throw new Error('Transaction must include a valid signature');
+	}
+	if (![null, undefined].includes(signSignature)) {
```
shuse2: Maybe we should decide if we should allow `signSignature != null`

src/transactions/utils/validation.js
```diff
@@ -71,3 +99,103 @@ export const validateAddress = address => {
...
+	}
+	if (
+		typeof id !== 'string' ||
+		!id.match(/^[0-9]+$/) ||
+		bignum(id).cmp(bignum(MAX_TRANSACTION_ID)) > 0
+	) {
+		throw new Error('Transaction must include a valid id');
+	}
+	if (fee !== FEES[type].toString()) {
+		throw new Error('Type 0 transactions must have a fee of 0.1 LSK');
```
shuse2: I think this is not only Type 0.

src/transactions/utils/validation.js
```diff
@@ -71,3 +99,103 @@ export const validateAddress = address => {
...
+		try {
+			validatePublicKey(recipientPublicKey);
+		} catch (error) {
+			throw new Error('Transaction has an invalid recipientPublicKey');
+		}
+	}
+	if (typeof asset !== 'object' || asset === null || Array.isArray(asset)) {
+		throw new Error('Transaction must include an asset object');
+	}
+	if (typeof asset.data !== 'undefined' && (typeof asset.data !== 'string' || asset.data.length > MAX_TRANSACTION_DATA_BYTES)) {
```
shuse2: It would be better to check the asset contents depending on the transaction type

src/transactions/utils/validation.js
```diff
@@ -71,3 +99,103 @@ export const validateAddress = address => {
...
+		!id.match(/^[0-9]+$/) ||
+		bignum(id).cmp(bignum(MAX_TRANSACTION_ID)) > 0
+	) {
+		throw new Error('Transaction must include a valid id');
+	}
+	if (fee !== FEES[type].toString()) {
+		throw new Error('Type 0 transactions must have a fee of 0.1 LSK');
+	}
+	try {
+		validateAddress(recipientId);
```
shuse2: depending on type, it doesn't require `recipientId`

src/transactions/utils/validation.js
```diff
@@ -71,3 +99,103 @@ export const validateAddress = address => {
...
+	} catch (error) {
+		throw new Error('Transaction must include a valid recipientId');
+	}
+	if (![null, undefined].includes(recipientPublicKey)) {
```
shuse2: we should check if we should include recipientPublicKey or not depending on transaction type

src/transactions/utils/validation.js
```diff
@@ -71,3 +99,103 @@ export const validateAddress = address => {
...
+		fee,
+		recipientId,
+		recipientPublicKey,
+		asset,
+	} = transaction;
+	if (![0, 1, 2, 3, 4, 5].includes(type)) {
+		throw new Error('Transaction type must be an integer between 0 and 5');
+	}
+	if (
+		typeof amount !== 'string' ||
```
shuse2: I think amount should be greater than 0 only if it's type 0

shuse2 (changes requested):
Target branch is `development`

willclarktech:
Closing for now. We'll revisit this for v1.1.0.

Issue #670, opened 2018-05-21
Replace API client to enable usage from react-native


shuse2:
### Description

Current popsicle seems not to support `react-native`.
We should replace with some library that supports `node.js`, `browsers` and `react-native`.

Possible library
- https://github.com/lquixada/cross-fetch
- https://github.com/axios/axios

tobiaslins:
Hi @shuse2 ,
How was this decision made? 
Why cross-fetch > axios?
What's about frisbee https://github.com/niftylettuce/frisbee ?

THanks :)

shuse2:
@tobiaslins 
@reyraa was trying out axios on RN, and it wasn't working well.
Also, there was no official support for RN by axios, so we decided to use cross-fetch.

tobiaslins:
https://facebook.github.io/react-native/docs/network.html#using-other-networking-libraries

axios and frisbee are in their offical documentation!

I also have 2 RN apps running for a while and using axios there since months

But it's ok for now, just wanted to know why

shuse2:
Ah yes! thank you for the information.
I remember now why I used axios for my RN apps before =p

Im having a problem to write test using `cross-fetch`, so I might re-consider.

shuse2:
so I had struggle with the test with `cross-fetch`, so I changed with `axios` =)

PR #679, opened 2018-06-01
Change API Client library to axios

shuse2:
### What was the problem?
APIClient was not able to be used in React Native.

### How did I fix it?
Change internal API Client library to `axios`

### How to test it?
`npm t`
use it in RN environment

### Review checklist

* The PR solves #670 
* Linting passes
* Tests pass
* Commit messages follow the


shuse2:
Currently @reyraa is checking if it works on RN

tobiaslins:
Please rename the PR to ....  *axios* :)

test/api_client/api_resource.js
```diff
@@ -37,8 +36,8 @@ describe('API resource module', () => {
 	};
 
 	const sendRequestResult = {
-		status: 200,
-		body: { success: true, sendRequest: true },
+		data: [],
```
willclarktech: `data.data`? Why not define the whole result here instead of partly here and partly in the stub `resolves` call?

test/api_client/api_resource.js
```diff
@@ -87,41 +86,45 @@ describe('API resource module', () => {
...
-				expect(popsicleStub).to.be.calledOnce;
+				expect(requestStub).to.be.calledOnce;
+				expect(requestStub).to.be.calledWithExactly(defaultRequest);
 				expect(handleRetryStub).not.to.be.called;
-				return expect(res).to.eql(sendRequestResult.body);
+				return expect(res).to.eql(sendRequestResult);
 			});
 		});
 
 		it('should make a request to API without calling retry when it successes', () => {
```
willclarktech: `when it succeeds`

test/api_client/api_resource.js
```diff
@@ -87,41 +86,45 @@ describe('API resource module', () => {
...
 			return resource.request(defaultRequest, true).then(res => {
-				expect(popsicleStub).to.be.calledOnce;
+				expect(requestStub).to.be.calledOnce;
+				expect(requestStub).to.be.calledWithExactly(defaultRequest);
+				expect(requestStub).to.be.calledOnce;
```
willclarktech: Duplicate expectation

test/api_client/api_resource.js
```diff
@@ -133,14 +136,11 @@ describe('API resource module', () => {
...
 				const statusCode = 300;
-				popsicleStub.returns({
-					use: () =>
-						Promise.resolve({
-							status: statusCode,
-							body: {
-								message: serverErrorMessage,
-							},
-						}),
+				requestStub.rejects({
```
willclarktech: Does axios really treat 300s as errors? Seems strange.

src/api_client/api_resource.js
```diff
@@ -37,19 +37,23 @@ export default class APIResource {
...
 					throw new Error(
-						`Status ${res.status} : An unknown error has occurred.`,
+						`Status ${error.response.status} : An unknown error has occurred.`,
 					);
 				}
-				return res.body;
+				throw error;
 			});
 
 		if (retry) {
```
willclarktech: Unrelated to this PR, but do we want to return on the line below?

test/api_client/api_resource.js
shuse2: it is configurable, but yes
```
  validateStatus: function (status) {
    return status >= 200 && status < 300; // default
  },
````

src/api_client/api_resource.js
shuse2: I think it's the same result.
if the `retry` is on, we want to attach the function to retry

test/api_client/api_resource.js
shuse2: first `data` is the response by `axios`, so thats why I wanted to separate them.
`sendRequestResult` represents the actual response from the servere

src/api_client/api_resource.js
willclarktech: The issue I'm referring to is not whether the retry is carried out, but what the return value of `request` should be.

Issue #685, opened 2018-06-10
Inconsistency in encryptAES256GCMWithPassword


5an1ty:
```
const encryptAES256GCMWithPassword = (plainText, password, iterations) => {
	const iv = crypto.randomBytes(12);
	const salt = crypto.randomBytes(16);
	const key = getKeyFromPassword(password, salt, iterations);

	const cipher = crypto.createCipheriv('aes-256-gcm', key, iv);
	const firstBlock = cipher.update(plainText, 'utf8');
	const encrypted = Buffer.concat([firstBlock, cipher.final()]);
	const tag = cipher.getAuthTag();

	return {
		iterations: iterations || null,
		cipherText: encrypted.toString('hex'),
		iv: iv.toString('hex'),
		salt: salt.toString('hex'),
		tag: tag.toString('hex'),
		version: ENCRYPTION_VERSION,
	};
};
```

```
const getKeyFromPassword = (password, salt, iterations) =>
	crypto.pbkdf2Sync(
		password,
		salt,
		iterations || PBKDF2_ITERATIONS,
		PBKDF2_KEYLEN,
		PBKDF2_HASH_FUNCTION,
	);
```

getKeyFromPassword seems to have a sane default for iterations in PBKDF2_ITERATIONS but encryptAES256GCMWithPassword defaults iterations to null

willclarktech:
If `iterations` is not explicitly passed to `encryptAES256GCMWithPassword`, it gets the value `undefined`, which is passed to `getKeyFromPassword`, so the default of `PBKDF2_ITERATIONS` is used. Have I misunderstood your issue?

5an1ty:
Yes, getKeyFromPassword works as intended, but encryptAES256GCMWithPassword return value will contain null for the iterations property.

```
return {
		iterations: iterations || null,
```

willclarktech:
Gotcha, thanks for the report.

willclarktech:
Ah I see what you mean. When we stringify that, `null` means the iterations are left out entirely: https://github.com/LiskHQ/lisk-elements/blob/development/src/cryptography/convert.js#L73-L87

willclarktech:
@5an1ty OK to close?

5an1ty:
Ok to close, thank you for clarifying.

willclarktech:
[...]

PR #692, opened 2018-06-18
Output default iterations when encrypting passphrases

willclarktech:
### What was the problem?

We didn't output the default number of iterations when encrypting passphrases.

### How did I fix it?

Better to be explicit.

### How to test it?

Encrypt a passphrase without specifying the number of iterations.

### Review checklist

* The PR solves #685 
* Linting passes
* Tests pass
* Commit messages follow the

### DIFF
--- a/src/cryptography/convert.js
+++ b/src/cryptography/convert.js
@@ -86,6 +86,17 @@ export const stringifyEncryptedPassphrase = encryptedPassphrase => {
 	return querystring.stringify(objectToStringify);
 };
 
+const parseIterations = iterationsString => {
+	const iterations =
+		iterationsString === undefined ? undefined : parseInt(iterationsString, 10);
+
+	if (Number.isNaN(iterations)) {
+		throw new Error('Could not parse iterations.');
+	}
+
+	return iterations;
+};
+
 export const parseEncryptedPassphrase = encryptedPassphrase => {
 	if (typeof encryptedPassphrase !== 'string') {
 		throw new Error('Encrypted passphrase to parse must be a string.');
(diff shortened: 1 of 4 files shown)


test/cryptography/convert.js
```diff
@@ -260,7 +260,7 @@ describe('convert', () => {
 			const stringifiedEncryptedPassphrase =
 				'salt=e8c7dae4c893e458e0ebb8bff9a36d84&cipherText=c0fab123d83c386ffacef9a171b6e0e0e9d913e58b7972df8e5ef358afbc65f99c9a2b6fe7716f708166ed72f59f007d2f96a91f48f0428dd51d7c9962e0c6a5fc27ca0722038f1f2cf16333&iv=1a2206e426c714091b7e48f6&tag=3a9d9f9f9a92c9a58296b8df64820c15&version=1';
 			const encryptedPassphrase = {
-				iterations: null,
+				iterations: undefined,
```
shuse2: What should happen when iterations was non-number, which results in NaN?
ie) `iteration=` or `iteration=null`

src/cryptography/convert.js
```diff
@@ -91,14 +91,10 @@ export const parseEncryptedPassphrase = encryptedPassphrase => {
...
-	const { salt, cipherText, iv, tag, version } = keyValuePairs;
-	const iterations =
-		keyValuePairs.iterations !== undefined
-			? parseInt(keyValuePairs.iterations, 10)
-			: null;
+	const { iterations, salt, cipherText, iv, tag, version } = keyValuePairs;
 
 	return {
-		iterations,
+		iterations: iterations === undefined ? undefined : parseInt(iterations, 10),
```
shuse2: What is the reason of strictly checking only `undefined`?

willclarktech: If it's present it should be a numerical string (therefore `parseInt`able).

PR #694, opened 2018-06-19
Use shared jenkins library

abelboldu:
### What was the problem?

Jenkinsfile was using functions shared across projects that can be imported now from https://github.com/LiskHQ/lisk-jenkins

### How did I fix it?

Remove functions and add import

### How to test it?

Tested in Jenkins itself


willclarktech:
Thanks @abelboldu!

PR #709, opened 2018-07-10
Add validation to decrypt passphrase - Closes #688

### LINKED ISSUE #688: parseEncryptedPassphrase and decryptPassphraseWithPassword combination should return proper error
### Description

`lisk.default.cryptography.parseEncryptedPassphrase` when used in combination of `lisk.default.cryptography.decryptPassphraseWithPassword` is not returning valid error in case of no salt provided

### Steps to Reproduce

```
const encryptedPassphrase = 'iterations=1&cipherText=73f5827fcd8eeab475abff71476cbce3b1ecacdeac55a738bb2f0a676d8e543bb92c91e1c1e3ddb6cef07a503f034dc7718e39657218d5a955859c5524be06de5954a5875b4c7b1cd11835e3477f1d04&iv=aac6a3b77c0594552bd9c932&tag=86231fb20e7b263264ca68b3585967ca&version=1'
try {
   lisk.default.cryptography.decryptPassphraseWithPassword(
       lisk.default.cryptography.parseEncryptedPassphrase(encryptedPassphrase));
} catch(error) {
   console.log(error)
}
```
**Expected behavior:** 
I expect error `No salt provided`

**Actual behavior:** 
I expect error `Argument must be a string`. This makes it confusing for user to understand the error, which element is missing or not provided in encrypted passphrase. 

**Reproduces how often:**
Always

### Versions
`1.0.0-beta.4`

willclarktech:
Closes #688

### What was the problem?

We had no validation when decrypting passphrases, with unhelpful error messages.

### How did I fix it?

Added validation and helpful error messages.

### How to test it?

Try to decrypt a passphrase with some missing parameters.

### Review checklist

* The PR solves #688 
* Linting passes
* Tests pass
* Commit messages follow the

### DIFF
--- a/packages/lisk-cryptography/src/convert.js
+++ b/packages/lisk-cryptography/src/convert.js
@@ -26,16 +26,18 @@ export const bufferToBigNumberString = bigNumberBuffer =>
 export const bufferToHex = buffer => Buffer.from(buffer).toString('hex');
 
 const hexRegex = /^[0-9a-f]+/i;
-export const hexToBuffer = hex => {
+export const hexToBuffer = (hex, argumentName = 'Argument') => {
 	if (typeof hex !== 'string') {
-		throw new TypeError('Argument must be a string.');
+		throw new TypeError(`${argumentName} must be a string.`);
 	}
 	const matchedHex = (hex.match(hexRegex) || [])[0];
 	if (!matchedHex || matchedHex.length !== hex.length) {
-		throw new TypeError('Argument must be a valid hex string.');
+		throw new TypeError(`${argumentName} must be a valid hex string.`);
 	}
 	if (matchedHex.length % 2 !== 0) {
-		throw new TypeError('Argument must have a valid length of hex string.');
+		throw new TypeError(
+			`${argumentName} must have a valid length of hex string.`,
+		);
 	}
 	return Buffer.from(matchedHex, 'hex');
 };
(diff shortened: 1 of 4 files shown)


src/cryptography/encrypt.js
```diff
@@ -133,10 +133,33 @@ const getTagBuffer = tag => {
...
+	const requiredProperties = [
+		[cipherText, 'Cipher text'],
+		[iv, 'IV'],
+		[salt, 'Salt'],
+		[tag, 'Tag'],
+	];
+
+	requiredProperties.forEach(([property, propertyName]) => {
+		try {
+			hexToBuffer(property);
```
shuse2: Wouldn't it weird to have partial validation named `validateEncryptedPassphrase`.
For example, `getTagBuffer` still validates if the length is 16, which is still used in encrypted passphrase, but not validated here.

willclarktech: Yes. How would you suggest doing it instead?

shuse2: oops sorry, forgot to put the suggestion.
Since we have method for each one with separate logic, maybe we should put the validation logic with meaningful message on each method?

src/cryptography/encrypt.js
```diff
@@ -133,10 +133,33 @@ const getTagBuffer = tag => {
...
-	{ iterations = PBKDF2_ITERATIONS, cipherText, iv, salt, tag },
-	password,
-) => {
+const validateEncryptedPassphrase = ({ cipherText, iv, salt, tag }) => {
+	const requiredProperties = [
+		[cipherText, 'Cipher text'],
+		[iv, 'IV'],
+		[salt, 'Salt'],
+		[tag, 'Tag'],
+	];
```
yatki: I would prefer to use Map instead of nested array. It'd save some memory since it doesn't have a prototype. see: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map `Iterating Maps with for..of` section

PR #712, opened 2018-07-11
Disable user-agent header when client options were not provided - Closes #696

### LINKED ISSUE #696: Refused to set unsafe header “User-Agent” in axios
### Description

I used `lisk-elements` in `lisk-hub`. Everything works as expected but there is an error in the browser console. I think xhr lib in `axois` causes this error.

**Expected behavior:** there must be no error in browser console because of `lisk-elements` 

**Actual behavior:** it keeps showing this error
`Refused to set unsafe header “User-Agent”`

yatki:
### What was the problem?
lisk-elements is throwing Refused to set unsafe header “User-Agent”warning when it's used in browsers.
### How did I fix it?
I updated getUserAgent function to return empty object when clientOptions were not provided.
### How to test it?
should not set User-Agent header when client options were not given
### Review checklist

* The PR solves #696 
* Linting passes
* Tests pass
* Commit messages follow the

### DIFF
--- a/packages/lisk-api-client/src/api_client.js
+++ b/packages/lisk-api-client/src/api_client.js
@@ -38,9 +38,13 @@ const commonHeaders = {
 	'Content-Type': 'application/json',
 };
 
-const getUserAgent = (
-	{ name = '????', version = '????', engine = '????' } = {},
-) => {
+const getClientHeaders = clientOptions => {
+	if (!clientOptions) {
+		return {};
+	}
+
+	const { name = '????', version = '????', engine = '????' } = clientOptions;
+
 	const liskElementsInformation =
 		'LiskElements/1.0 (+https://github.com/LiskHQ/lisk-elements)';
 	const locale =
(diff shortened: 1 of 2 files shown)


willclarktech:
@yatki This is failing on CI.

packages/lisk-api-client/src/api_client.js
```diff
@@ -107,9 +114,7 @@ export default class APIClient {
 			{},
 			commonHeaders,
 			options.nethash ? { nethash: options.nethash } : {},
-			{
-				'User-Agent': getUserAgent(options.client),
-			},
+			getUserAgent(options.client),
```
willclarktech: This `Object.assign` call combines various headers objects. User agent is on the wrong semantic level for this. The function as written doesn't just get the user agent, it gets a headers object which happens to only include the user agent (currently). We should either use the previous pattern or rename the function something like `getClientHeaders`.

packages/lisk-api-client/test/api_client.js
```diff
@@ -210,6 +212,14 @@ describe('APIClient module', () => {
 					.to.have.property('headers')
 					.and.eql(customHeaders);
 			});
+
+			it('should not set User-Agent header when client options were not given', () => {
+				apiClient = new APIClient(defaultNodes, {
+					version: customHeaders.version,
```
willclarktech: Actually `version`/`customHeaders.version` doesn't make sense here. Could you remove it from the other test too?

PR #726, opened 2018-07-16
Reorganize browser tests - lisk-api-client - closes #717

### LINKED ISSUE #717: Reorganize browser tests - lisk-api-client
Child issue of #714

yatki:
### What was the problem?
We split submodules of lisk-elements to be separate npm packages. So we have to write package specific browser tests.

### How did I fix it?
I organised browser tests structure to support running node unit tests in browser for lisk-api-client.

### How to test it?
```
cd packages/lisk-api-client
npm run build:browsertest
npm run test:browser
```
### Review checklist

* The PR solves #717 
* Linting passes
* Tests pass
* Commit messages follow the


.gitignore
```diff
@@ -48,6 +48,10 @@ tmp
 
 # Distribution folders
 dist/
+dist-node/
+dist-browser/
```
willclarktech: Do these need to be preceded with `*/**`?

Jenkinsfile
```diff
@@ -44,17 +44,6 @@ pipeline {
 				}
 			}
 		}
-		stage('Run browser tests') {
```
willclarktech: Do we have an issue to reinstate these once the refactor has been completed?

packages/lisk-api-client/.eslintrc.json
```diff
@@ -1,3 +1,7 @@
 {
-	"extends": ["../../.eslintrc.json"]
+	"extends": ["../../.eslintrc.json"],
+	"globals": {
+		"mocha": true,
+		"document": true
+	}
```
willclarktech: These globals don't belong in the root of the package. They should be in `.eslintrc.json` files in the relevant subdirectory/ies. Also these are better set via the `mocha` and `browser` `env` settings.

packages/lisk-api-client/package.json
```diff
@@ -19,22 +19,46 @@
 		"node": ">=6.3 <=9.5",
 		"npm": ">=3 <=5"
 	},
-	"main": "dist/index.js",
+	"browser": "dist-browser/index.min.js",
+	"main": "dist-node/index.js",
 	"scripts": {
+		"babel": "babel src -d dist-node",
```
willclarktech: We should take the opportunity to give this script the generic name `transpile`.

packages/lisk-api-client/package.json
```diff
@@ -19,22 +19,46 @@
...
+		"babel:browsertest":
```
willclarktech: Likewise this one.

packages/lisk-api-client/package.json
```diff
@@ -19,22 +19,46 @@
...
+			"babel src -d ./browsertest/src && BABEL_ENV=browsertest npx babel test --ignore test/transactions/dapp.js -d ./browsertest/test",
```
willclarktech: We can't use `npx` until we support npm >=5.

packages/lisk-api-client/package.json
```diff
@@ -19,22 +19,46 @@
...
+			"uglifyjs -nm -o ./dist-browser/index.min.js ./dist-browser/index.js",
+		"uglify:browsertest":
+			"uglifyjs -o ./browsertest/browsertest.min.js ./browsertest/browsertest.js",
 		"format": "prettier --write \"*.{js,json}\" \"{src,test}/**/*.{js,json}\"",
 		"lint": "eslint .",
 		"lint:fix": "npm run lint -- --fix",
 		"test": "NODE_ENV=test nyc mocha test",
 		"test:watch": "npm test -- --watch",
 		"test:watch:min": "npm run test:watch -- --reporter=min",
+		"test:node":
```
willclarktech: This script fails for me.

.gitignore
```diff
@@ -48,6 +48,10 @@ tmp
...
+*/**/browsertest/browsertest.js
+*/**/browsertest/browsertest.min.js
```
willclarktech: We also need to add some cypress auto-generated files here. See https://github.com/LiskHQ/lisk-elements/blob/1.0.0/.gitignore#L26-L29

packages/lisk-api-client/package.json
willclarktech: We don't need the `--ignore` here.

packages/lisk-api-client/package.json
```diff
@@ -19,22 +19,46 @@
...
+		"build:browsertest":
+			"npm run babel:browsertest && npm run browserify:browsertest && npm run uglify:browsertest",
+		"postbuild:browsertest": "rm -r browsertest/src browsertest/test",
+		"prebuild:node": "rm -r dist-node/* || mkdir dist-node | echo",
+		"build:node": "npm run babel",
+		"prebuild:browser":
+			"rm ./dist-browser/index.js ./dist-browser/index.min.js | echo",
+		"build:browser": "npm run babel && npm run browserify && npm run uglify",
+		"prebuild": "npm run prebuild:node && npm run prebuild:browser",
+		"build": "npm run babel && npm run browserify && npm run uglify",
```
willclarktech: Let's make this `npm run build:node && npm run build:browser`. Then we can remove the `npm run babel` from `build:browser`.

packages/lisk-api-client/package.json
```diff
@@ -19,22 +19,46 @@
...
+		"build:check": "node -e \"require('./dist-node')\"",
 		"prepublishOnly":
```
willclarktech: This script breaks now.

packages/lisk-api-client/cypress.json
```diff
@@ -0,0 +1,5 @@
+{
+	"baseUrl": "http://localhost:8080",
+	"defaultCommandTimeout": 300e3,
```
willclarktech: Probably we can decrease this without video.

packages/lisk-api-client/browsertest/setup.js
```diff
@@ -0,0 +1,21 @@
...
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ *
+ */
+mocha.setup({
+	ui: 'bdd',
+	timeout: 30000,
```
willclarktech: Can we decrease this?

.gitignore
yatki:
```
If the pattern ends with a slash, it is removed for the purpose of the following description, but it would only find a match with a directory. In other words, foo/ will match a directory foo and paths underneath it, but will not match a regular file or a symbolic link foo (this is consistent with the way how pathspec works in general in Git).
```
It finds child folders when the pathname ends with /. However, doesn't do it for files. That's why i set it like this.

packages/lisk-api-client/package.json
yatki: @willclarktech  I did it, however now `build:browser` will fail if `build:node` wasn't run before which feels a bit wrong. I would expect `build:browser` to build browser bundle without depending on another command.

packages/lisk-api-client/cypress.json
yatki: @willclarktech I deleted `defaultCommandTimeout` and the default value is 4000ms.

.eslintrc.json
```diff
@@ -6,9 +6,8 @@
 		"import/no-extraneous-dependencies": [
 			"error",
 			{
-				"devDependencies": ["test/**"],
-				"optionalDependencies": false,
-				"packageDir": "./"
+				"devDependencies": true,
```
willclarktech: This should be a pattern, not a blanket `true`. It needs to exclude `src` in any case.

.gitignore
```diff
@@ -23,6 +23,12 @@ test/.coverage-unit
 # nyc test coverage
 .nyc_output
 
+# cypress media
+**/cypress/screenshots/
+**/cypress/videos/
+**/cypress/plugins/
+**/cypress/support/
```
willclarktech: Do we need the `**/` here then? (C.f. below.)

packages/lisk-api-client/browsertest/.eslintrc.json
```diff
@@ -0,0 +1,8 @@
+{
+	"env": {
+		"mocha": true
+	},
+	"globals": {
+		"document": true
```
willclarktech: This should be set via the `browser` env setting.

packages/lisk-api-client/package.json
```diff
@@ -19,24 +19,48 @@
...
+		"uglify:browsertest":
+			"uglifyjs -o ./browsertest/browsertest.min.js ./browsertest/browsertest.js",
 		"format": "prettier --write \"*.{js,json}\" \"{src,test}/**/*.{js,json}\"",
 		"lint": "eslint .",
 		"lint:fix": "npm run lint -- --fix",
 		"test": "NODE_ENV=test nyc mocha test",
 		"test:watch": "npm test -- --watch",
 		"test:watch:min": "npm run test:watch -- --reporter=min",
+		"test:node":
+			"npm run build:check && BABEL_ENV=buildcheck mocha test --compilers js:babel-register --recursive",
```
willclarktech: This doesn't make sense anymore. We should reinstate the buildcheck `BABEL_ENV`: https://github.com/LiskHQ/lisk-elements/blob/1.0.0/.babelrc#L17-L26

.gitignore
yatki: @willclarktech weirdly, .gitignore doesn't support if we define subfolders. Eg. if I just write `cypress/` git ignores the cypress folder under packages/lisk-api-client. but if we define `cypress\/support/` it doesn't match with the folder ¯\_(ツ)_/¯.

PR #732, opened 2018-07-19
Remove coveralls - Closes #727

### LINKED ISSUE #727: Remove coveralls
### Expected behavior

Our tools should be useful.

### Actual behavior

We constantly check coverage locally, the coveralls plugin doesn't cover all our needs, is frequently ignored, and it will be complex to adapt to the monorepo change.

1.1.0

willclarktech:
### What was the problem?

Coveralls was adding overhead and not providing much value.

### How did I fix it?

Removed coveralls.

### How to test it?

Check Jenkins.

### Review checklist

* The PR solves #727 
* Linting passes
* Tests pass
* Commit messages follow the

### DIFF
--- a/packages/lisk-api-client/package.json
+++ b/packages/lisk-api-client/package.json
@@ -47,7 +47,7 @@
 			"if [ -z $JENKINS_HOME ]; then npm run cover:local; else npm run cover:ci; fi",
 		"cover:base": "NODE_ENV=test nyc report",
 		"cover:local": "npm run cover:base -- --reporter=html --reporter=text",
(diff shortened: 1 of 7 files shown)


packages/lisk-api-client/package.json
```diff
@@ -47,7 +47,7 @@
 			"if [ -z $JENKINS_HOME ]; then npm run cover:local; else npm run cover:ci; fi",
```
shuse2: I think we need to change this script too
Also, other package.json too

PR #734, opened 2018-07-20
Add transaction validate feature

shuse2:
### What was the problem?
Schema check of the transactions were missing.

### Description
Add schema definition with `JSON schema` and use `ajv` to validate them.

### How to test it?
`npm t`

### Review checklist

* The PR solves #676 
* Linting passes
* Tests pass
* Commit messages follow the


packages/lisk-transactions/src/utils/validation/validator.js
```diff
@@ -0,0 +1,98 @@
...
+
+const getValidator = type => {
+	const schema = schemaMap[type];
+	if (!schema) {
+		throw new Error('Unsupported transaction type.');
+	}
+	return schema;
+};
+
+export const validateTransaction = tx => {
```
willclarktech: Surely this function belongs in the `validation` file?

packages/lisk-transactions/src/utils/validation/schema.js
```diff
@@ -0,0 +1,282 @@
...
+		'asset',
+		'signature',
+	],
+	properties: {
+		id: {
+			type: 'string',
+			format: 'number',
+			minLength: 1,
+			maxLength: 20,
+		},
```
willclarktech: Is it possible to set a range of values here?

packages/lisk-transactions/src/utils/validation/schema.js
```diff
@@ -0,0 +1,282 @@
...
+		amount: {
+			type: 'string',
+			format: 'number',
+		},
```
willclarktech: Same question but we can at least set the maximum length.

packages/lisk-transactions/src/utils/validation/schema.js
```diff
@@ -0,0 +1,282 @@
...
+		fee: {
+			type: 'string',
+			format: 'number',
+		},
```
willclarktech: As above.

packages/lisk-transactions/src/utils/validation/schema.js
willclarktech: Maybe we need an `amount` format?

packages/lisk-transactions/src/utils/validation/schema.js
```diff
@@ -0,0 +1,282 @@
...
+			type: 'string',
+			format: 'publicKey',
+		},
+		senderSecondPublicKey: {
+			type: 'string',
+			format: 'publicKey',
+		},
+		recipientId: {
+			type: 'string',
+		},
```
willclarktech: `address` format?

packages/lisk-transactions/src/utils/validation/schema.js
```diff
@@ -0,0 +1,282 @@
...
+		signatures: {
+			type: 'array',
+			uniqueItems: true,
+			items: {
+				type: 'string',
+				format: 'signature',
+			},
+		},
+		asset: {
+			type: 'object',
```
willclarktech: Should we be more specific here already?

packages/lisk-transactions/src/utils/validation/schema.js
```diff
@@ -0,0 +1,282 @@
...
+				},
+				recipientId: {
+					format: 'address',
+				},
+				asset: {
+					type: 'object',
+					properties: {
+						data: {
+							type: 'string',
+							maxLength: 32,
```
willclarktech: 64?

packages/lisk-transactions/src/utils/validation/schema.js
```diff
@@ -0,0 +1,282 @@
...
+					type: 'object',
+					required: ['delegate'],
+					properties: {
+						delegate: {
+							type: 'object',
+							required: ['username'],
+							properties: {
+								username: {
+									type: 'string',
+									maxLength: 10,
```
willclarktech: Where's this maximum from?

packages/lisk-transactions/src/utils/validation/schema.js
```diff
@@ -0,0 +1,282 @@
...
+		source: { $ref: 'lisk/base-transaction' },
+		with: {
+			properties: {
+				asset: {
+					type: 'object',
+					required: ['votes'],
+					properties: {
+						votes: {
+							type: 'array',
+							uniqueItems: true,
```
willclarktech: This won't include voting and unvoting for the same delegate in a vote transaction. (Is that scenario even invalid?)

packages/lisk-transactions/src/utils/validation/schema.js
```diff
@@ -0,0 +1,282 @@
...
+									maximum: 24,
+								},
+								keysgroup: {
+									type: 'array',
+									uniqueItems: true,
+									minItems: 2,
+									maxItems: 16,
+									items: {
+										type: 'string',
+										format: 'actionPublicKey',
```
willclarktech: Actually `-`-prepended aren't valid for this one.

packages/lisk-transactions/src/utils/validation/validator.js
```diff
@@ -0,0 +1,98 @@
...
+ * Removal or modification of this copyright notice is prohibited.
+ *
+ */
+import Ajv from 'ajv';
+import ajvMergePatch from 'ajv-merge-patch';
+import { validateAddress, validatePublicKey } from './validation';
+import * as schemas from './schema';
+
+const validator = new Ajv({ allErrors: true });
+// Add $merge and $patch keywords
```
willclarktech: Prefer importing under an informative name to relying on comments.

packages/lisk-transactions/src/utils/validation/validator.js
```diff
@@ -0,0 +1,98 @@
...
+ajvMergePatch(validator);
+
+validator.addFormat('number', data => data === '' || /^[0-9]+$/g.test(data));
```
willclarktech: Name is a bit misleading here: a blank string doesn't seem like a number to me.

packages/lisk-transactions/src/utils/validation/validator.js
```diff
@@ -0,0 +1,98 @@
...
+
+validator.addFormat(
+	'signature',
+	data => data === '' || /^[a-f0-9]{128}$/i.test(data),
```
willclarktech: Same concern here.

willclarktech: How confident are you this can be case insensitive?

packages/lisk-transactions/src/utils/validation/validator.js
```diff
@@ -0,0 +1,98 @@
...
+	const schema = schemaMap[type];
+	if (!schema) {
+		throw new Error('Unsupported transaction type.');
+	}
+	return schema;
+};
+
+export const validateTransaction = tx => {
+	if (typeof tx.type !== 'number') {
+		throw new Error('Transaction type must be a number.');
```
willclarktech: Why do this validation here? It's already covered by `getValidator` plus `baseTransaction`.

packages/lisk-transactions/src/utils/validation/validator.js
```diff
@@ -0,0 +1,98 @@
...
+	const validate = getValidator(tx.type);
+	const valid = validate(tx);
+	// Ajv produces merge error when error happens within $merge
+	const errors = validate.errors
+		? validate.errors.filter(e => e.keyword !== '$merge')
+		: null;
+	return {
+		valid,
+		errors,
+	};
```
willclarktech: Should we update our other validation functions to match this pattern?

packages/lisk-transactions/src/utils/validation/schema.js
shuse2: I think we need to add different validation function, ill look into it

packages/lisk-transactions/src/utils/validation/schema.js
shuse2: depending on type, it would be different, and merged so I think as `base` it should be enough

packages/lisk-transactions/src/utils/validation/schema.js
shuse2: ahh I think i have calculate wrong in general...it's 20byte max

packages/lisk-transactions/src/utils/validation/validator.js
shuse2: I think it's out of scope, so Ill create another issue for this

packages/lisk-transactions/src/utils/validation/schema.js
shuse2: same as below ='(

packages/lisk-transactions/src/utils/validation/validator.js
shuse2: I thought it would be more understandable to put this function where all the other function, which used in this function is defined

packages/lisk-transactions/src/utils/validation/schema.js
willclarktech: Is it possible to have union types?

packages/lisk-transactions/src/utils/validation/schema.js
willclarktech: Where's it from though?

packages/lisk-transactions/src/utils/validation/schema.js
willclarktech: 🙀

packages/lisk-transactions/src/utils/validation/validator.js
willclarktech: I think it makes more sense to separate schemas, validators, and end-user validation functions.

packages/lisk-transactions/src/utils/validation/schema.js
shuse2: since it's a format so would be weird, also we are using merge for each specific type of transaction

packages/lisk-transactions/src/utils/validation/validator.js
```diff
@@ -0,0 +1,130 @@
...
+	isGreaterThanMaxTransactionAmount,
+	isGreaterThanMaxTransactionId,
+	isNumberString,
+} from './validation';
+import * as schemas from './schema';
+
+const validator = new Ajv({ allErrors: true });
+addMergePatchKeywords(validator);
+
+validator.addFormat('signature', data => /^[a-f0-9]{128}$/i.test(data));
```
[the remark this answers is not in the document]
shuse2: seems like it...
https://github.com/LiskHQ/lisk/blob/development/helpers/z_schema.js#L150

packages/lisk-transactions/src/utils/validation/validator.js
```diff
@@ -0,0 +1,130 @@
...
+		validatePublicKey(data);
+		return true;
+	} catch (error) {
+		return false;
+	}
+});
+
+validator.addFormat('actionPublicKey', data => {
+	try {
+		const action = data[0];
```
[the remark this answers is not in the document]
shuse2: If we use this, then we need to `join` the publicKey again

packages/lisk-transactions/src/utils/validation/schema.js
```diff
@@ -0,0 +1,282 @@
...
+		source: { $ref: 'lisk/base-transaction' },
+		with: {
+			properties: {
+				asset: {
+					type: 'object',
+					required: ['votes'],
+					properties: {
+						votes: {
+							type: 'array',
+							uniquePublicKeys: true,
```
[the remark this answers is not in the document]
shuse2: yep
`{"error":"Could not broadcast transaction: Status 409 : Multiple votes for same delegate are not allowed"}`

packages/lisk-transactions/src/utils/validation/index.js
```diff
@@ -0,0 +1,25 @@
...
+export {
+	checkPublicKeysForDuplicates,
+	validatePublicKey,
+	validatePublicKeys,
+	validateKeysgroup,
+	validateAddress,
+	isGreaterThanMaxTransactionAmount,
+} from './validation';
+
+export { validateTransaction } from './validator';
```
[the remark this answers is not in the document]
shuse2: mmm yea i think it's either putting most of the stuff on one file or split like this... ='(

packages/lisk-transactions/src/utils/validation/schema.js
```diff
@@ -0,0 +1,282 @@
...
+
+export const transferTransaction = {
+	$merge: {
+		source: { $ref: 'lisk/base-transaction' },
+		with: {
+			properties: {
+				amount: {
+					type: 'string',
+					format: 'amount',
+					minLength: 1,
```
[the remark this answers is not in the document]
shuse2: it's was required anyway... so removed

packages/lisk-transactions/test/utils/validation/validator.js
```diff
@@ -0,0 +1,999 @@
...
+		it('should validate to be false with errors when id contains more than 20 digits', () => {
+			const invalidTransaction = Object.assign({}, validTransaction, {
+				id: '123456789012345678901',
+			});
+			const { valid, errors } = validateTransaction(invalidTransaction);
+			expect(errors[0].dataPath).to.equal('.id');
+			return expect(valid).to.be.false;
+		});
+
+		it('should validate to be false with errors when id contains more maximum id possible', () => {
```
willclarktech: `when id is too large`

packages/lisk-transactions/test/utils/validation/validator.js
```diff
@@ -0,0 +1,1019 @@
...
+		});
+
+		it('should validate to be false with errors when timestamp is not integer', () => {
+			const invalidTransaction = Object.assign({}, validTransaction, {
+				timestamp: 12.34,
+			});
+			const { valid, errors } = validateTransaction(invalidTransaction);
+			expect(errors[0].dataPath).to.equal('.timestamp');
+			return expect(valid).to.be.false;
+		});
```
willclarktech: Is there a maximum for timestamp? Negative?

packages/lisk-transactions/test/utils/validation/validator.js
```diff
@@ -0,0 +1,1019 @@
...
+				recipientId: undefined,
+			});
+			const { valid, errors } = validateTransaction(invalidTransaction);
+			expect(errors[0].params.missingProperty).to.equal('recipientId');
+			return expect(valid).to.be.false;
+		});
+
+		it('should validate to be false with errors when recipientId is not valid address', () => {
+			const invalidTransaction = Object.assign({}, validTransaction, {
+				recipientId: '1234567891011121314151617181920L',
```
willclarktech: Why is this invalid address different to the above example for `senderId`? Do we need two tests for each?

packages/lisk-transactions/test/utils/validation/validator.js
```diff
@@ -0,0 +1,1019 @@
...
+					'fad7b3b31c337b39190d206831c1eaadc6bbf3878a3507a868a5fbb03471b383042bf3bb7cee20d9844f2f4d1bb90d08bc3589b8b7d27a538be285deec7a9504',
+					'zzzzzzzzzzzzzzzzzzzzzz6831c1eaadc6bbf3878a3507a868a5fbb03471b383042bf3bb7cee20d9844f2f4d1bb90d08bc3589b8b7d27a538be285deec7a9504',
+				],
+			});
+			const { valid, errors } = validateTransaction(invalidTransaction);
+			expect(errors[0].dataPath).to.equal('.signatures[1]');
+			return expect(valid).to.be.false;
+		});
+
+		it('should validate to be true when signatures is contains valid signatures', () => {
```
willclarktech: `when signatures contains only non-duplicated valid signatures`

packages/lisk-transactions/test/utils/validation/validator.js
```diff
@@ -0,0 +1,1019 @@
...
+			});
+			const { valid, errors } = validateTransaction(invalidTransaction);
+			expect(errors).to.be.null;
+			return expect(valid).to.be.true;
+		});
+
+		it('should validate to be true when data is 64 bytes', () => {
+			const invalidTransaction = Object.assign({}, validTransaction, {
+				asset: {
+					data: 'fad7b3b31c337b39190d206831c1eaad',
```
willclarktech: This should have a length of 64 and I think it's confusing if it only has hex characters.

packages/lisk-transactions/test/utils/validation/validator.js
```diff
@@ -0,0 +1,1019 @@
...
+					},
+				},
+				signature:
+					'98b0dbd77efadaee7112915fd4a08551a594460f1d30f0ead7efb328ff60521a5d88c7f8eaf700f2d91c86228a41eacf34729268269dee5d74dd61ed1a79d40b',
+				id: '9795926042598492108',
+			};
+			return Promise.resolve();
+		});
+
+		it('should validate to be true', () => {
```
willclarktech: `true without errors` to match the previous description.

packages/lisk-transactions/test/utils/validation/validator.js
```diff
@@ -0,0 +1,1019 @@
...
+						'-215b667a32a5cd51a94c9c2046c11fffb08c65748febec099451e3b164452bca',
+					],
+				},
+			});
+			const { valid, errors } = validateTransaction(invalidTransaction);
+			expect(errors[0].dataPath).to.equal('.asset.votes');
+			return expect(valid).to.be.false;
+		});
+
+		it('should validate to be false with errors when asset.votes has no action key', () => {
```
willclarktech: `when a public key in asset.votes has no sign`

packages/lisk-transactions/test/utils/validation/validator.js
```diff
@@ -0,0 +1,1019 @@
...
+						'215b667a32a5cd51a94c9c2046c11fffb08c65748febec099451e3b164452bca',
+					],
+				},
+			});
+			const { valid, errors } = validateTransaction(invalidTransaction);
+			expect(errors[0].dataPath).to.equal('.asset.votes[1]');
+			return expect(valid).to.be.false;
+		});
+
+		it('should validate to be false with errors when asset.votes has non public key', () => {
```
willclarktech: `invalid public key`

packages/lisk-transactions/test/utils/validation/validator.js
```diff
@@ -0,0 +1,1019 @@
...
+			});
+			const { valid, errors } = validateTransaction(invalidTransaction);
+			expect(errors[0].params.missingProperty).to.equal('multisignature');
+			return expect(valid).to.be.false;
+		});
+
+		it('should validate to be false with errors when asset.multisignature.min does not exist', () => {
+			const invalidTransaction = Object.assign({}, validTransaction, {
+				asset: {
+					multisignature: {
```
willclarktech: Consider using `Object.assign` on this level too to indicate the shared properties?

packages/lisk-transactions/test/utils/validation/validator.js
```diff
@@ -0,0 +1,1019 @@
...
+							'+215b667a32a5cd51a94c9c2046c11fffb08c65748febec099451e3b164452bca',
+							'+922fbfdd596fa78269bbcadc67ec2a1cc15fc929a19c462169568d7a3df1a1aa',
+						],
+					},
+				},
+			});
+			const { valid, errors } = validateTransaction(invalidTransaction);
+			expect(errors[0].params.missingProperty).to.equal('min');
+			return expect(valid).to.be.false;
+		});
```
willclarktech: Is there a general maximum for `min`? What about if `min` is higher than the number of keys in the keysgroup?

packages/lisk-transactions/test/utils/validation/validator.js
```diff
@@ -0,0 +1,1019 @@
...
+						keysgroup: [],
+					},
+				},
+			});
+			const { valid, errors } = validateTransaction(invalidTransaction);
+			expect(errors[0].dataPath).to.equal('.asset.multisignature.keysgroup');
+			return expect(valid).to.be.false;
+		});
+
+		it('should validate to be false with errors when asset.multisignature.keysgroup has only one elements', () => {
```
willclarktech: `element`

packages/lisk-transactions/test/utils/validation/validator.js
```diff
@@ -0,0 +1,1019 @@
...
+						],
+					},
+				},
+			});
+			const { valid, errors } = validateTransaction(invalidTransaction);
+			expect(errors[0].dataPath).to.equal('.asset.multisignature.keysgroup');
+			return expect(valid).to.be.false;
+		});
+
+		it('should validate to be false with errors when asset.multisignature.keysgroup has no action key', () => {
```
willclarktech: `a public key with no sign`

packages/lisk-transactions/test/utils/validation/validator.js
```diff
@@ -0,0 +1,1019 @@
...
+						],
+					},
+				},
+			});
+			const { valid, errors } = validateTransaction(invalidTransaction);
+			expect(errors[0].dataPath).to.equal('.asset.multisignature.keysgroup[0]');
+			return expect(valid).to.be.false;
+		});
+
+		it('should validate to be false with errors when asset.multisignature.keysgroup has "-" action key', () => {
```
willclarktech: `sign`

packages/lisk-transactions/test/utils/validation/validator.js
```diff
@@ -0,0 +1,1019 @@
...
+						],
+					},
+				},
+			});
+			const { valid, errors } = validateTransaction(invalidTransaction);
+			expect(errors[0].dataPath).to.equal('.asset.multisignature.keysgroup[0]');
+			return expect(valid).to.be.false;
+		});
+
+		it('should validate to be false with errors when asset.multisignature.keysgroup has non public key', () => {
```
willclarktech: `an invalid public key`

packages/lisk-transactions/test/utils/validation/validator.js
```diff
@@ -0,0 +1,1019 @@
...
+		});
+
+		it('should validate to be false with errors when asset.multisignature.keysgroup has more than 16 elements', () => {
+			const invalidTransaction = Object.assign({}, validTransaction, {
+				asset: {
+					multisignature: {
+						min: 2,
+						lifetime: 24,
+						keysgroup: [
+							'215b667a32a5cd51a94c9c2046c11fffb08c65748febec099451e3b164452bca',
```
willclarktech: This should be signed?

packages/lisk-transactions/test/utils/validation/validator.js
```diff
@@ -0,0 +1,1019 @@
...
+				timestamp: 54196080,
+				asset: {
+					dapp: {
+						category: 0,
+						name: 'OpwYfwHdAfO4ncn7D',
+						description: 'Ds',
+						tags: '1nkfYVcgsisKnXyn7k0',
+						type: 0,
+						link: 'WQDusb0DgH',
+						icon: 'RPAFQsBIsE',
```
willclarktech: Can we have more realistic data?

packages/lisk-transactions/test/utils/validation/validator.js
```diff
@@ -0,0 +1,1019 @@
...
+			});
+			const { valid, errors } = validateTransaction(invalidTransaction);
+			expect(errors).to.have.length(4);
+			return expect(valid).to.be.false;
+		});
+
+		it('should validate to be false with errors when asset.dapp.name is not string', () => {
+			const invalidTransaction = Object.assign({}, validTransaction, {
+				asset: {
+					dapp: {
```
willclarktech: As above re nested `Object.assign`

packages/lisk-transactions/test/utils/validation/validator.js
```diff
@@ -0,0 +1,1019 @@
...
+						type: 1,
+						link: 'link',
+						icon: 'RPAFQsBIsE',
+					},
+				},
+			});
+			const { valid, errors } = validateTransaction(invalidTransaction);
+			expect(errors[0].dataPath).to.equal('.asset.dapp.category');
+			return expect(valid).to.be.false;
+		});
```
willclarktech: What about tags, description, icon, category?

packages/lisk-transactions/test/utils/validation/validator.js
shuse2: it was taken from fixtures =p

willclarktech: :0

packages/lisk-transactions/test/utils/validation/validator.js
shuse2: it's tested in unit test for the function, but i think it's better to test different value than the same one

packages/lisk-transactions/test/utils/validation/validator.js
shuse2: I think cross field check is over the scope of validator ='(

packages/lisk-transactions/src/utils/validation/validator.js
```diff
@@ -0,0 +1,127 @@
...
+
+const getTransactionValidator = type => {
+	const schema = schemaMap[type];
+	if (!schema) {
+		throw new Error('Unsupported transaction type.');
+	}
+	return schema;
+};
+
+export const validateTransaction = tx => {
```
willclarktech: How about making this the default export and renaming the file `validateTransaction.js`?

packages/lisk-transactions/test/utils/validation/validator.js
```diff
@@ -0,0 +1,1079 @@
...
+		it('should validate to be false with errors when id contains non-number', () => {
+			const invalidTransaction = Object.assign({}, validTransaction, {
+				id: '123nonnumber',
+			});
+			const { valid, errors } = validateTransaction(invalidTransaction);
+			expect(errors[0].dataPath).to.equal('.id');
+			return expect(valid).to.be.false;
+		});
+
+		it('should validate to be false with errors when id contains more than 20 digits', () => {
```
willclarktech: This case is superseded by the below case.

packages/lisk-transactions/test/utils/validation/validator.js
willclarktech: Hmm, but a cross-field check is still something we can statically determine before hitting the API. Maybe it's not something we can use ajv for, but in that case perhaps we should have additional validation depending on the transaction type?

packages/lisk-transactions/src/utils/validation/validator.js
shuse2: I was thinking that we could add another `validateSignature` function later.
maybe keep the default as it is and move the `validateTransaction` in another file?

packages/lisk-transactions/src/utils/validation/validate_transaction.js
```diff
@@ -0,0 +1,71 @@
...
+	return {
+		valid: true,
+		errors: null,
+	};
+};
+
+const validateTransaction = tx => {
+	const validate = getTransactionValidator(tx.type);
+	const valid = validate(tx);
+	// Ajv produces merge error when error happens within $merge
```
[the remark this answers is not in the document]
shuse2: We don't have to, but it creates duplicate error (one for merge and one for actual error).
So I think it would be better to filter out error which is not really useful

PR #738, opened 2018-07-23
Use React Native/Android-compatible buffer reverse function - Closes #710

### LINKED ISSUE #710: Android doesn't behave with instances of Buffer like instances of Array
### Expected behavior
Lisk Element should work in Android environment seamlessly.

### Actual behavior
Since the variables of type Buffer are not instances of Array, they don't have methods like `reverse` or `slice`. Major browsers add these methods to the interface by default but this is not the case for Android. 
The same issue has been brought up in [some other projects](https://github.com/Airbitz/react-native-fast-crypto/issues/4) too.

### Steps to reproduce
Try running Lisk element in a simple RN application in the Android environment. it should throw errors similar to this
screenshot_1531237325

OS: Android
Environment: React Native
Version:

willclarktech:
### What was the problem?

Using `.reverse` on buffers isn't supported in the React Native/Android environment.

### How did I fix it?

Used a library which will implement reverse for buffers.

### How to test it?

Try to use `getFirstEightBytesReversed` in particular (plus other functions) in React Native/Android.

### Review checklist

* The PR solves #710 
* Linting passes
* Tests pass
* Commit messages follow the

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,6 +67,7 @@
     "babel-runtime": "=6.26.0",
     "bip39": "=2.4.0",
     "browserify-bignum": "=1.3.0-2",
+    "buffer-reverse": "=1.0.1",
     "ed2curve": "=0.2.1",
     "tweetnacl": "=1.0.0",
     "varuint-bitcoin": "=1.1.0"
--- a/src/cryptography/convert.js
+++ b/src/cryptography/convert.js
@@ -14,6 +14,7 @@
  */
 import querystring from 'querystring';
 import bignum from 'browserify-bignum';
+import reverseBuffer from 'buffer-reverse';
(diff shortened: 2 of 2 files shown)


packages/lisk-cryptography/src/convert.js
```diff
@@ -14,6 +14,7 @@
  */
 import querystring from 'querystring';
 import bignum from 'browserify-bignum';
+import reverse from 'buffer-reverse';
```
yatki: I think we should be a bit more clear on function names, so I suggest to change this to be `bufferReverse`.

willclarktech:
Waiting for feedback from @reyraa

reyraa:
@willclarktech This addresses the issue. Thank you

PR #749, opened 2018-07-26
Create Symlinks Script - #closes 748

### LINKED ISSUE #747: Add clean scripts
### Expected behavior

There should be an easy way to remove fragile files from a package, e.g. node_modules and distribution directories, and an easy way to remove them from all packages at once from the root.

### Actual behavior

We have no such scripts, but @yatki has some locally.

1.1.0

### LINKED ISSUE #748: Add init script
### Expected behavior

There should be an easy way to setup default symlinks for e.g. auxiliary files when creating a new package.

### Actual behavior

We have no such script, but @yatki has one locally.

1.1.0

yatki:
### Review checklist

* The PR resolves #748, resolves #747 
* Linting passes
* Tests pass
* Commit messages follow the

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,8 @@
 	},
 	"scripts": {
 		"bootstrap": "lerna bootstrap",
+		"clean": "lerna run clean",
+		"clean:node_modules": "lerna clean",
 		"format": "lerna run format",
 		"lint": "lerna run lint",
 		"lint:fix": "lerna run lint:fix",
@@ -32,6 +34,7 @@
 		"build": "lerna run build",
 		"build:browsertest": "lerna run build:browsertest",
 		"cover": "lerna run cover",
+		"init": "./scripts/init.sh",
 		"postinstall": "npm run bootstrap",
 		"precommit": "lint-staged && npm run lint",
 		"prepush": "npm run lint && npm test"
--- a/packages/lisk-api-client/package.json
+++ b/packages/lisk-api-client/package.json
@@ -28,6 +28,7 @@
 		"browserify:browsertest": "browserify ./browsertest.build/test/*.js ./browsertest.build/test/**/*.js -o ./browsertest.build/browsertest.js -s lisk",
 		"uglify": "uglifyjs -nm -o ./dist-browser/index.min.js ./dist-browser/index.js",
 		"uglify:browsertest": "uglifyjs -o ./browsertest.build/browsertest.min.js ./browsertest.build/browsertest.js",
+		"clean": "./scripts/clean.sh",
 		"format": "prettier --write \"*.{js,json}\" \"{src,test}/**/*.{js,json}\"",
 		"lint": "eslint .",
 		"lint:fix": "npm run lint -- --fix",
--- a/scripts/init.sh
+++ b/scripts/init.sh
@@ -1,0 +1,43 @@
+#!/bin/bash
+
+# Unofficial strict mode
+set -euo pipefail
+IFS=$'\n\t'
+
+packageName=${1:-}
+
+ROOT_PACKAGE_NAME=$(jq --raw-output '.name' package.json)
+if [ "$ROOT_PACKAGE_NAME" != "lisk-elements-monorepo" ]; then
+	echo "Please use the command in the project root directory."
+	exit 1
+fi
+
+if [ -z "$packageName" ] || [[ "$packageName" =~ [^a-zA-Z0-9-] ]]; then
+	echo "Usage: npm run init -- PACKAGE_NAME"
+	exit 1
+fi
+
+packageDir="./packages/$packageName"
+# Just in case package folder doesn't exist yet.
+mkdir -p "$packageDir"
+
+templates=(
+	"browsertest"
+	"cypress"
+	"scripts"
+	".babelrc"
+	".eslintignore"
+	".npmignore"
+	".npmrc"
+	".nycrc"
+	".prettierignore"
+	".prettierrc.json"
+	"cypress.json"
+)
+
+for i in "${templates[@]}"
+do
+	if [ ! -e "$packageDir/${i}" ];	then
+		ln -vs "../../templates/$i.tmpl" "$packageDir/$i"
+	fi
+done
--- a/templates/scripts.tmpl/clean.sh
+++ b/templates/scripts.tmpl/clean.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+# Unofficial strict mode
+set -euo pipefail
+IFS=$'\n\t'
+
+# Usage
+# npm run clean
+
+rm -rf .nyc_output/
+rm -rf browsertest.build/
+rm -rf coverage/
+rm -rf dist-browser/
+rm -rf dist-node/
+rm -rf lerna-debug.log
+rm -rf npm-debug.log
--- a/packages/lisk-api-client/scripts
+++ b/packages/lisk-api-client/scripts
@@ -1,0 +1,1 @@
+../../templates/scripts.tmpl
--- a/packages/lisk-constants/package.json
+++ b/packages/lisk-constants/package.json
@@ -28,6 +28,7 @@
 		"browserify:browsertest": "browserify ./browsertest.build/test/*.js -o ./browsertest.build/browsertest.js -s lisk",
 		"uglify": "uglifyjs -nm -o ./dist-browser/index.min.js ./dist-browser/index.js",
 		"uglify:browsertest": "uglifyjs -o ./browsertest.build/browsertest.min.js ./browsertest.build/browsertest.js",
+		"clean": "./scripts/clean.sh",
 		"format": "prettier --write \"*.{js,json}\" \"{src,test}/**/*.{js,json}\"",
 		"lint": "eslint .",
 		"lint:fix": "npm run lint -- --fix",
--- a/packages/lisk-constants/scripts
+++ b/packages/lisk-constants/scripts
@@ -1,0 +1,1 @@
+../../templates/scripts.tmpl
--- a/packages/lisk-cryptography/package.json
+++ b/packages/lisk-cryptography/package.json
@@ -29,6 +29,7 @@
 		"browserify:browsertest": "browserify ./browsertest.build/test/*.js -o ./browsertest.build/browsertest.js -s lisk",
 		"uglify": "uglifyjs -nm -o ./dist-browser/index.min.js ./dist-browser/index.js",
 		"uglify:browsertest": "uglifyjs -o ./browsertest.build/browsertest.min.js ./browsertest.build/browsertest.js",
+		"clean": "./scripts/clean.sh",
 		"format": "prettier --write \"*.{js,json}\" \"{src,test}/**/*.{js,json}\"",
 		"lint": "eslint .",
 		"lint:fix": "npm run lint -- --fix",
--- a/packages/lisk-cryptography/scripts
+++ b/packages/lisk-cryptography/scripts
@@ -1,0 +1,1 @@
+../../templates/scripts.tmpl
--- a/packages/lisk-elements/package.json
+++ b/packages/lisk-elements/package.json
@@ -31,6 +31,7 @@
 		"browserify:browsertest": "browserify ./browsertest.build/test/*.js -o ./browsertest.build/browsertest.js -s lisk --node",
 		"uglify": "uglifyjs -nm -o ./dist-browser/index.min.js ./dist-browser/index.js",
 		"uglify:browsertest": "uglifyjs -o ./browsertest.build/browsertest.min.js ./browsertest.build/browsertest.js",
+		"clean": "./scripts/clean.sh",
 		"format": "prettier --write \"*.{js,json}\" \"{src,test}/**/*.{js,json}\"",
 		"lint": "eslint .",
 		"lint:fix": "npm run lint -- --fix",
--- a/packages/lisk-elements/scripts/clean.sh
+++ b/packages/lisk-elements/scripts/clean.sh
@@ -1,0 +1,1 @@
+../../../templates/scripts.tmpl/clean.sh
--- a/packages/lisk-passphrase/package.json
+++ b/packages/lisk-passphrase/package.json
@@ -29,6 +29,7 @@
 		"browserify:browsertest": "browserify ./browsertest.build/test/*.js -o ./browsertest.build/browsertest.js -s lisk",
 		"uglify": "uglifyjs -nm -o ./dist-browser/index.min.js ./dist-browser/index.js",
 		"uglify:browsertest": "uglifyjs -o ./browsertest.build/browsertest.min.js ./browsertest.build/browsertest.js",
+		"clean": "./scripts/clean.sh",
 		"format": "prettier --write \"*.{js,json}\" \"{src,test}/**/*.{js,json}\"",
 		"lint": "eslint .",
 		"lint:fix": "npm run lint -- --fix",
--- a/packages/lisk-passphrase/scripts
+++ b/packages/lisk-passphrase/scripts
@@ -1,0 +1,1 @@
+../../templates/scripts.tmpl
--- a/packages/lisk-transactions/package.json
+++ b/packages/lisk-transactions/package.json
@@ -29,6 +29,7 @@
 		"browserify:browsertest": "browserify ./browsertest.build/test/*.js -o ./browsertest.build/browsertest.js -s lisk",
 		"uglify": "uglifyjs -nm -o ./dist-browser/index.min.js ./dist-browser/index.js",
 		"uglify:browsertest": "uglifyjs -o ./browsertest.build/browsertest.min.js ./browsertest.build/browsertest.js",
+		"clean": "./scripts/clean.sh",
 		"format": "prettier --write \"*.{js,json}\" \"{src,test}/**/*.{js,json}\"",
 		"lint": "eslint .",
 		"lint:fix": "npm run lint -- --fix",
--- a/packages/lisk-transactions/scripts
+++ b/packages/lisk-transactions/scripts
@@ -1,0 +1,1 @@
+../../templates/scripts.tmpl


scripts/create-symlinks.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+# Usage in the project root folder:
+# ./scripts/create-symlinks.sh PACKAGE_NAME
+#
+# Usage in package folder
+# ../../scripts/create-symlinks.sh
+
+packageDir=".";
```
willclarktech: The root directory is a one-off situation, we shouldn't need to cover that case.

scripts/create-symlinks.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+# Usage in the project root folder:
+# ./scripts/create-symlinks.sh PACKAGE_NAME
```
willclarktech: Can we make this an npm script with an argument?

scripts/create-symlinks.sh
```diff
@@ -0,0 +1,25 @@
...
+# ../../scripts/create-symlinks.sh
+
+packageDir=".";
+if [ -n "$1" ]
+	then
+		# Override $pageDir if argument was provided.
+		packageDir="./packages/$1";
+
+		# Just in case package folder doesn't exist yet.
+		mkdir -p $packageDir;
```
willclarktech: I think this script would make more sense as an `init` script: whenever we want to add a new package we run it, and it creates our package with defaults.

scripts/create-symlinks.sh
```diff
@@ -0,0 +1,25 @@
...
+fi
+
+declare -a arr=(".prettierrc.json" ".prettierignore" ".npmrc" ".nycrc" ".npmignore" ".babelrc" ".eslintignore" "cypress.json" "cypress" "browsertest")
+
+for i in "${arr[@]}"
+do
+   ln -s ../../templates/$i.tmpl $packageDir/$i;
```
willclarktech: This wouldn't work for the root directory btw.

scripts/create-symlinks.sh
```diff
@@ -0,0 +1,25 @@
...
+   printf "Symlink Created: $packageDir/$i -> ../../templates/$i.tmpl\n";
```
willclarktech: Not sure we need feedback in the success case here.

scripts/create-symlinks.sh
yatki: `npm run init -- packageName` would this usage make sense? @willclarktech

scripts/create-symlinks.sh
yatki: This was in case of running script inside the package folder. But i think it's better to run the script from only one place.

scripts/create-symlinks.sh
yatki: I didn't intend to support creating of files in root folder.

scripts/create-symlinks.sh
yatki: I think it's better to have this in case of a typo and would be good to see if they were created during CI process ?

package.json
```diff
@@ -30,6 +30,7 @@
 		"build": "lerna run build",
 		"build:browsertest": "lerna run build:browsertest",
 		"cover": "lerna run cover",
+		"init": "./scripts/create-symlinks.sh",
```
willclarktech: Let's rename the file too.

scripts/create-symlinks.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
```
willclarktech: Can we get away with `#!/bin/sh`?

scripts/create-symlinks.sh
```diff
@@ -0,0 +1,29 @@
...
+
+# Usage in the project root folder:
+# npm run init -- PACKAGE_NAME
+
+
+if [ -z "$1" ]
+	then
+		echo "Please provide a package name!";
```
willclarktech: `echo "Usage: npm run init -- PACKAGE_NAME"`

scripts/create-symlinks.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
+# Usage in the project root folder:
+# npm run init -- PACKAGE_NAME
```
willclarktech: We can remove this comment if we have the below change to `echo`.

scripts/create-symlinks.sh
```diff
@@ -0,0 +1,29 @@
...
+
+
+if [ -z "$1" ]
+	then
+		echo "Please provide a package name!";
+		exit 1;
+fi
+
+packageDir="./packages/$1";
+# Just in case package folder doesn't exist yet.
```
willclarktech: Should it error if the package dir does exist?

scripts/create-symlinks.sh
```diff
@@ -0,0 +1,29 @@
...
+mkdir -p $packageDir;
+
+files=(".prettierrc.json" ".prettierignore" ".npmrc" ".nycrc" ".npmignore" ".babelrc" ".eslintignore" "cypress.json" "cypress" "browsertest" "scripts")
```
willclarktech: Can we split these on multiple lines for cleaner git diffs?

scripts/create-symlinks.sh
```diff
@@ -0,0 +1,29 @@
...
+
+
+for i in "${files[@]}"
+do
+	if [ ! -e "$packageDir/${i}" ]
+		then
+			ln -s ../../templates/$i.tmpl $packageDir/$i;
+			printf "Symlink Created: $packageDir/$i -> ../../templates/$i.tmpl\n";
```
willclarktech: `created`

willclarktech: Why have you used `printf` here and `echo` above?

scripts/create-symlinks.sh
```diff
@@ -0,0 +1,29 @@
...
+	fi;
+done
+
+chmod +x $packageDir/scripts/*.sh
```
willclarktech: 👍

scripts/create-symlinks.sh
```diff
@@ -0,0 +1,29 @@
...
+mkdir -p $packageDir;
+
+files=(".prettierrc.json" ".prettierignore" ".npmrc" ".nycrc" ".npmignore" ".babelrc" ".eslintignore" "cypress.json" "cypress" "browsertest" "scripts")
+
+
+for i in "${files[@]}"
+do
+	if [ ! -e "$packageDir/${i}" ]
+		then
+			ln -s ../../templates/$i.tmpl $packageDir/$i;
```
willclarktech: Maybe wrap in `"`?

templates/scripts.tmpl/cleanup.sh
```diff
@@ -0,0 +1,22 @@
...
+rm -r dist-node/
+rm -r dist-browser/
+rm -r browsertest.build/
+rm -r .nyc_output/
+rm -r *.log
+
+if [ "$1" == "all" ]
+	then
+		rm -rf node_modules/;
+		echo "'node_modules' folders were deleted from all packages!"
```
willclarktech: Why `echo` here and not for the others?

willclarktech: "all packages" is a bit misleading.

templates/scripts.tmpl/cleanup.sh
```diff
@@ -0,0 +1,22 @@
...
+#
+# To delete node_modules folder:
+# npm run clean -- all
+
+rm -r dist/
+rm -r dist-node/
+rm -r dist-browser/
+rm -r browsertest.build/
+rm -r .nyc_output/
+rm -r *.log
```
willclarktech: I'm worried this will delete logs people might want to keep. Is there anything besides `npm-debug.log` and `lerna-debug.log`? Also I think npm debug logs can end with timestamps or something sometimes.

templates/scripts.tmpl/cleanup.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+# Please run this script only in the project root folder.
+#
+# Usage
+# npm run clean
+#
+# To delete node_modules folder:
+# npm run clean -- all
```
willclarktech: Prefer something more explicit than `all`. Maybe `--include-node-modules`?

scripts/create-symlinks.sh
```diff
@@ -0,0 +1,29 @@
...
+# npm run init -- PACKAGE_NAME
+
+
+if [ -z "$1" ]
+	then
+		echo "Please provide a package name!";
+		exit 1;
+fi
+
+packageDir="./packages/$1";
```
willclarktech: What if `$1` contains a `/`?

scripts/create-symlinks.sh
fchavant: No, there are bashisms like the files array below.

scripts/create-symlinks.sh
fchavant: Use `ln --verbose --symblic` instead.

scripts/create-symlinks.sh
fchavant: Why not make the scripts executable in git?

scripts/create-symlinks.sh
fchavant: Superfluous  `;` (one of many)

scripts/create-symlinks.sh
yatki: We may wanna create symlinks after creating the folder. So let's keep it just in case.

scripts/create-symlinks.sh
fchavant: You might want to use the unofficial [strict mode](http://redsymbol.net/articles/unofficial-bash-strict-mode/).

scripts/create-symlinks.sh
```diff
@@ -0,0 +1,29 @@
...
+# Just in case package folder doesn't exist yet.
+mkdir -p $packageDir;
```
fchavant: Double quote to prevent globbing and word splitting.

templates/scripts.tmpl/cleanup.sh
```diff
@@ -0,0 +1,22 @@
...
+
+rm -r dist/
```
fchavant: Use `-f` switch.

templates/scripts.tmpl/cleanup.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+# Please run this script only in the project root folder.
```
fchavant: Should be enforced.

templates/scripts.tmpl/cleanup.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
```
fchavant: General piece of advice: you might want to install `shellcheck` and run it against those scripts.

scripts/create-symlinks.sh
alepop: @fchavant this works for me with `sh`.

templates/scripts.tmpl/cleanup.sh
yatki: This is not the case anymore. We are copying cleanup script in every package folder separately. So I deleted the comment.

scripts/init.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
+
+set -euo pipefail
+IFS=$'\n\t'
+
+PACKAGE_NAME=$(cat package.json \
```
fchavant: Use `jq`.

packages/lisk-api-client/package.json
```diff
@@ -28,6 +28,7 @@
 		"browserify:browsertest": "browserify ./browsertest.build/test/*.js ./browsertest.build/test/**/*.js -o ./browsertest.build/browsertest.js -s lisk",
 		"uglify": "uglifyjs -nm -o ./dist-browser/index.min.js ./dist-browser/index.js",
 		"uglify:browsertest": "uglifyjs -o ./browsertest.build/browsertest.min.js ./browsertest.build/browsertest.js",
+		"clean": "./scripts/cleanup.sh",
```
willclarktech: Why not just use the same name for the script?

scripts/init.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
+
+set -euo pipefail
```
willclarktech: Let's have a comment `Unofficial strict mode`.

scripts/init.sh
```diff
@@ -0,0 +1,42 @@
...
+if [ -z "$packageName" ] || [[ "$packageName" =~ [^a-zA-Z0-9-] ]]; then
+	echo "Usage: npm run init -- PACKAGE_NAME"
+	exit 1
+fi
+
+packageDir="./packages/$packageName"
+# Just in case package folder doesn't exist yet.
+mkdir -p "$packageDir"
+
+files=(
```
willclarktech: `sources` is maybe more accurate since this includes directories.

scripts/init.sh
```diff
@@ -0,0 +1,42 @@
...
+	"cypress.json"
+	"cypress"
+	"browsertest"
+	"scripts"
+)
+
+for i in "${files[@]}"
+do
+	if [ ! -e "$packageDir/${i}" ];	then
+		ln -vs "../../templates/$i.tmpl" "$packageDir/$i"
```
willclarktech: We should use the long form of these options for intelligibility.

templates/scripts.tmpl/cleanup.sh
```diff
@@ -0,0 +1,19 @@
...
+# OR
+# npm run clean -- --m
+
+rm -rf dist/
+rm -rf dist-node/
+rm -rf dist-browser/
+rm -rf browsertest.build/
+rm -rf .nyc_output/
+
+if [ "$1" == "--include-node-modules" ] ||  [ "$1" == "--m" ]; then
```
willclarktech: `-m` for the short option. Or `-n`?

templates/scripts.tmpl/cleanup.sh
```diff
@@ -0,0 +1,19 @@
...
+# To delete node_modules folder:
+# npm run clean -- --include-node-modules
+# OR
+# npm run clean -- --m
+
+rm -rf dist/
+rm -rf dist-node/
+rm -rf dist-browser/
+rm -rf browsertest.build/
+rm -rf .nyc_output/
```
willclarktech: I still think it's worth specifying `npm-debug.log` and `lerna-debug.log`, just not a glob for logfiles.

templates/scripts.tmpl/cleanup.sh
```diff
@@ -0,0 +1,19 @@
...
+# OR
+# npm run clean -- -m
+
+rm -rf dist/
+rm -rf dist-node/
+rm -rf dist-browser/
+rm -rf browsertest.build/
+rm -rf .nyc_output/
+
+if [ "$1" == "--include-node-modules" ] ||  [ "$1" == "-m" ]; then
```
willclarktech: Does `-n` make more sense than `-m`?

scripts/init.sh
```diff
@@ -0,0 +1,42 @@
...
+
+set -euo pipefail
+IFS=$'\n\t'
+
+packageName=${1:-}
+
+ROOT_PACKAGE_NAME=$(jq -r ".name" < package.json)
+if [ "$ROOT_PACKAGE_NAME" != "lisk-elements-monorepo" ]; then
+	echo "Please use the command in the project root directory."
+	exit 1;
```
willclarktech: `;`

templates/scripts.tmpl/cleanup.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
+set -euo pipefail
```
willclarktech: As above re explanatory comment.

packages/lisk-api-client/package.json
yatki: Should we stick with clean or cleanup ? @willclarktech ?

willclarktech: I have a mild preference for `clean`

scripts/init.sh
yatki: I agree files is not a good choice but then I prefer to use `templates`.

scripts/init.sh
yatki: That causes the error below:

```

Comment (willclarktech) [reply scripts/init.sh]: ?

File: scripts/init.sh [hunk 33]
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
+
+set -euo pipefail
+IFS=$'\n\t'
+
+packageName=${1:-}
+
+ROOT_PACKAGE_NAME=$(jq -r ".name" < package.json)
```
fchavant: You don't need any redirection, the following will work just fine: `jq --raw-output '.name' package.json`

templates/scripts.tmpl/clean.sh
```diff
@@ -0,0 +1,27 @@
...
+# OR
+# npm run clean -- -m
+
+rm -rf dist/
+rm -rf dist-node/
+rm -rf dist-browser/
+rm -rf browsertest.build/
+rm -rf .nyc_output/
+rm -rf npm-debug.log
+rm -rf larna-debug.log
```
willclarktech: `lerna`

scripts/init.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
+
+#Unofficial strict mode
```
willclarktech: Prefer a space after `#` for uniformity/legibility.

templates/scripts.tmpl/clean.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+
+#Unofficial strict mode
```
willclarktech: Space

package.json
```diff
@@ -30,6 +30,7 @@
 		"build": "lerna run build",
 		"build:browsertest": "lerna run build:browsertest",
 		"cover": "lerna run cover",
```
willclarktech: Since we're combining the two scripts now, we should have a root `clean` script which just runs `lerna run clean`.

templates/scripts.tmpl/clean.sh
```diff
@@ -0,0 +1,27 @@
...
+set -euo pipefail
+IFS=$'\n\t'
+
+arg1=${1:-}
```
willclarktech: This doesn't work if you call it via `npm run`.

scripts/create-symlinks.sh
willclarktech: @alepop Your default shell is probably bash (or at least bash compatible). If you run `sh --version` does it tell you it's bash?

templates/scripts.tmpl/clean.sh
yatki: @willclarktech lerna run doesn't pass the arguments to the child commands. I switched it to lerna exec. Also there was a lerna command for cleaning but just for 'node_modules'. see: https://www.npmjs.com/package/@lerna/clean

scripts/init.sh
shuse2: do we need to install `jq` locally? 😼

shuse2: `./scripts/init.sh: line 9: jq: command not found` ='(

yatki: @shuse2 yeap. Will told me that jq is already a dependency for lisk core. So `brew install jq` .

scripts/init.sh
```diff
@@ -0,0 +1,43 @@
...
+	".prettierignore"
+	".npmrc"
+	".nycrc"
+	".npmignore"
+	".babelrc"
+	".eslintignore"
+	"cypress.json"
+	"cypress"
+	"browsertest"
+	"scripts"
```
willclarktech: Prefer alphabetical.

templates/scripts.tmpl/clean.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+# Unofficial strict mode
+set -euo pipefail
+IFS=$'\n\t'
+
+# Usage
+# npm run clean
+
+rm -rf dist/
```
willclarktech: We shouldn't need this anymore, but we do need `coverage/`. Also prefer alphabetical (can still be directories first).

PR #762, opened 2018-08-03
Validate transaction inputs - Closes #752

### LINKED ISSUE #752: Validate transaction creator inputs
### Expected behavior

When attempting to create transactions with invalid inputs, the user should be informed which inputs are invalid and how. We should be able to use the new schema validation introduced in https://github.com/LiskHQ/lisk-elements/issues/734

### Actual behavior

We don't validate these inputs thoroughly.

### Steps to reproduce

Try to create a transfer transaction with a recipient ID that's out of range for example.

1.1.0+

yatki:
### Review checklist

* The PR resolves #752 
* Linting passes
* Tests pass


packages/lisk-transactions/src/4_register_multisignature_account.js
```diff
@@ -17,11 +17,28 @@ import {
...
 	validateKeysgroup,
 	wrapTransactionCreator,
+	isValidInteger,
 } from './utils';
 
-const registerMultisignatureAccount = ({ keysgroup, lifetime, minimum }) => {
+const validateInputs = ({ keysgroup, lifetime, minimum }) => {
 	validateKeysgroup(keysgroup);
 
+	if (!isValidInteger(lifetime) || lifetime < 1 || lifetime > 72) {
```
shuse2: It would be better to split this magic numbers in to some place.
In core, these are network `constants`

packages/lisk-transactions/src/4_register_multisignature_account.js
```diff
@@ -17,11 +17,28 @@ import {
...
+		throw new Error(
+			'Please provide a valid lifetime value. Expected integer between 1 and 72.',
+		);
+	}
+
+	if (!isValidInteger(minimum) || minimum < 1 || minimum > 15) {
```
shuse2: Same as above

packages/lisk-transactions/src/0_transfer.js
```diff
@@ -27,12 +51,11 @@ const createAsset = data => {
 			throw new Error(
 				'Invalid encoding in transaction data. Data must be utf-8 encoded.',
 			);
-		return { data };
 	}
-	return {};
 };
 
 const transfer = ({ amount, recipientId, recipientPublicKey, data }) => {
+	validateInputs({ amount, recipientId, recipientPublicKey, data });
```
willclarktech: Prefer something like this (also applies to other functions):
```js
const transfer = (inputs) => {
    validateInputs(inputs);
    const { amount, recipientId, recipientPublicKey, data } = inputs;
    // ...
};

File: packages/lisk-transactions/src/utils/get_address_and_public_key_from_recipient_data.js [hunk 4]
```diff
@@ -18,19 +18,7 @@ const getAddressAndPublicKeyFromRecipientData = ({
...
 	recipientPublicKey,
 }) => {
-	if (recipientId && recipientPublicKey) {
-		const addressFromPublicKey = cryptography.getAddressFromPublicKey(
-			recipientPublicKey,
-		);
-		if (recipientId === addressFromPublicKey) {
-			return { address: recipientId, publicKey: recipientPublicKey };
-		}
-		throw new Error(
```
shuse2: Shouldn't it still check the matching `address` vs `publicKey`?

packages/lisk-transactions/test/0_transfer.js
```diff
@@ -18,13 +18,16 @@ const time = require('../src/utils/time');
...
 	const fixedPoint = 10 ** 8;
-	const recipientId = '58191285901858109L';
 	const testData = 'data';
 	const passphrase = 'secret';
 	const secondPassphrase = 'second secret';
 	const transactionType = 0;
 	const publicKey =
 		'5d036a858ce89f844491762eb89e2bfbd50a4a0a0da658e4b2628b25b117ae09';
+	const recipientId = '18160565574430594874L';
+	// const recipientPublicKey = '5d036a858ce89f844491762eb89e2bfbd50a4a0a0da658e4b2628b25b117ae09';
```
shuse2: Why comment out?

packages/lisk-transactions/src/0_transfer.js
```diff
@@ -12,13 +12,37 @@
...
 } from './utils';
 
 const createAsset = data => {
+	if (data && data.length > 0) {
+		return { data };
+	}
+	return {};
+};
+
+const validateInputs = ({ amount, recipientId, recipientPublicKey, data }) => {
```
willclarktech: This function should also validate the format of `recipientId` and `recipientPublicKey`.

packages/lisk-transactions/src/0_transfer.js
```diff
@@ -27,12 +51,11 @@ const createAsset = data => {
 			throw new Error(
 				'Invalid encoding in transaction data. Data must be utf-8 encoded.',
```
willclarktech: I'm not actually confident we should enforce UTF8.

packages/lisk-transactions/src/4_register_multisignature_account.js
willclarktech: Should also check if it's larger than the keysgroup length.

packages/lisk-transactions/src/4_register_multisignature_account.js
```diff
@@ -17,11 +17,28 @@ import {
...
+		throw new Error(
+			'Please provide a valid minimum value. Expected integer between 1 and 15.',
```
willclarktech: This message will need to be updated too.

packages/lisk-transactions/src/5_create_dapp.js
```diff
@@ -38,7 +38,7 @@ const validateOptions = options => {
 };
```
willclarktech: We should probably also have minimal checks for the other properties.

packages/lisk-transactions/src/utils/get_address_and_public_key_from_recipient_data.js
```diff
@@ -18,19 +18,7 @@ const getAddressAndPublicKeyFromRecipientData = ({
 	recipientId,
 	recipientPublicKey,
 }) => {
```
willclarktech: I don't think this function adds much value anymore. Can we remove it?

packages/lisk-transactions/src/utils/validation/validation.js
```diff
@@ -86,3 +86,9 @@ export const isNumberString = str => {
 	}
 	return /^[0-9]+$/g.test(str);
 };
+
+export const validateAmount = data =>
+	isNumberString(data) && !isGreaterThanMaxTransactionAmount(bignum(data));
+
+export const isValidInteger = num =>
+	typeof num === 'number' && parseInt(num, 10) === num;
```
willclarktech: Is the `typeof` check necessary?

packages/lisk-transactions/test/0_transfer.js
```diff
@@ -220,6 +223,26 @@ describe('#transfer transaction', () => {
...
 
+			it('should throw error when invalid amount provided', () => {
+				return expect(
+					transfer.bind(null, {
+						amount: '18446744073709551616',
+					}),
+				).to.throw('Please enter a valid amount!');
+			});
+
+			it('should throw error when recipientId & non-matching publicKey provided', () => {
```
willclarktech: Either `recipientPublicKey` or `public key`.

packages/lisk-transactions/test/1_register_second_passphrase.js
```diff
@@ -181,6 +181,18 @@ describe('#registerSecondPassphrase transaction', () => {
...
+			it('should throw error when secondPassphrase was not provided', () => {
+				return expect(registerSecondPassphrase.bind(null, {})).to.throw(
+					'Please provide a secondPassphrase. Expected string.',
+				);
+			});
+
+			it('should not throw error when secondPassphrase is empty string', () => {
+				return expect(
+					registerSecondPassphrase.bind(null, { secondPassphrase: '' }),
+				).to.not.throw('Please provide a secondPassphrase. Expected string.');
```
willclarktech: `.not.to.throw()`: no need to check the message.

packages/lisk-transactions/test/4_register_multisignature_account.js
```diff
@@ -312,6 +312,111 @@ describe('#registerMultisignatureAccount transaction', () => {
 				return Promise.resolve();
 			});
 
+			describe('shoud throw error', () => {
```
willclarktech: `validation errors`: This is a suite description, but `should` implies a test.

packages/lisk-transactions/test/4_register_multisignature_account.js
```diff
@@ -312,6 +312,111 @@ describe('#registerMultisignatureAccount transaction', () => {
...
+				describe('when #lifetime', () => {
```
willclarktech: `#` is used fairly inconsistently, but generally indicates a function or method under test. This should just be `lifetime`.

packages/lisk-transactions/test/4_register_multisignature_account.js
```diff
@@ -312,6 +312,111 @@ describe('#registerMultisignatureAccount transaction', () => {
...
+							registerMultisignatureAccount.bind(null, {
+								keysgroup,
+								lifetime: -1,
+							}),
+						).to.throw(lifetimeErrorMessage);
+					});
+				});
+			});
+
+			describe('when #minimum', () => {
```
willclarktech: As above.

packages/lisk-transactions/src/0_transfer.js
```diff
@@ -12,13 +12,37 @@
...
+	}
+	return {};
+};
+
+const validateInputs = ({ amount, recipientId, recipientPublicKey, data }) => {
+	if (!validateAmount(amount)) {
+		throw new Error('Please enter a valid amount!');
+	}
+
+	if (recipientId && recipientPublicKey) {
```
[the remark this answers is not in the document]
yatki: @shuse2 I don't think so. This case only covers the situation publicKey and recipientId don't match. if only recipientId is provided then it can go to transaction and when we broadcast get an error message from network if there is any error with it. 

In the schema recipientId was only defined as a string. If there is a unique pattern I can validate but right now I think this is fine ?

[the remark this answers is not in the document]
shuse2: Reason why `recipientId` is defined in base Transaction is that it is required, but depending on the type of transaction the requirement defers.
For type `0`, it's defined as `format: address`, but in other cases, it is empty string.

But here, I think it should check if the `recipientId` exists.

packages/lisk-transactions/src/0_transfer.js
yatki: @shuse2 could you please let me know what do you think on this? I have no idea :D

packages/lisk-transactions/src/utils/validation/validation.js
yatki: If the num is undefined then parseInt is throwing exception. So I think we need it. Also proper polyfill is here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isInteger#Polyfill

packages/lisk-transactions/src/utils/get_address_and_public_key_from_recipient_data.js
yatki: @willclarktech we are still using it in type o transactions.

yatki: @willclarktech are you suggesting to use `cryptography.getAddressFromPublicKey` directly in the tx function ?

willclarktech: Yes.

packages/lisk-transactions/src/5_create_dapp.js
yatki: @willclarktech Are they optional keys and we already have schema check. Is it really necessary ?

packages/lisk-transactions/src/0_transfer.js
yatki: @shuse2 @willclarktech could you elaborate this?

packages/lisk-transactions/src/0_transfer.js
```diff
@@ -12,13 +12,48 @@
...
+	if (data && data.length > 0) {
+		return { data };
+	}
+	return {};
+};
+
+const validateInputs = ({ amount, recipientId, recipientPublicKey, data }) => {
+	if (!validateAmount(amount)) {
+		throw new Error(
+			'Please provide an amount. Expected a valid number in string format!',
```
willclarktech: `!` -> `.`

packages/lisk-transactions/src/2_register_delegate.js
```diff
@@ -15,14 +15,31 @@
...
-	fee: DELEGATE_FEE.toString(),
-	asset: {
-		delegate: {
-			username,
+const validateInputs = ({ username }) => {
+	if (!username || typeof username !== 'string') {
+		throw new Error('Please provide a username. Expected string.');
+	}
+
+	if (username.length > 20) {
```
willclarktech: This could also be moved into constants.

packages/lisk-transactions/src/4_register_multisignature_account.js
```diff
@@ -12,16 +12,52 @@
...
+		!isValidInteger(minimum) ||
+		minimum < MULTISIGNATURE_MIN_SIGNATURES ||
+		minimum > MULTISIGNATURE_MAX_SIGNATURES
+	) {
+		throw new Error(
+			`Please provide a valid minimum value. Expected integer between ${MULTISIGNATURE_MIN_SIGNATURES} and ${MULTISIGNATURE_MAX_SIGNATURES}.`,
+		);
+	}
+
+	if (keysgroup.length < minimum) {
```
willclarktech: @shuse2 This should be `keysgroup.length < minimum - 1`, right?

packages/lisk-transactions/src/constants.js
```diff
@@ -22,6 +22,10 @@ export const SIGNATURE_FEE = 5 * FIXED_POINT;
 export const DELEGATE_FEE = 25 * FIXED_POINT;
 export const VOTE_FEE = 1 * FIXED_POINT;
 export const MULTISIGNATURE_FEE = 5 * FIXED_POINT;
+export const MULTISIGNATURE_MAX_LIFETIME = 72;
+export const MULTISIGNATURE_MIN_LIFETIME = 1;
+export const MULTISIGNATURE_MAX_SIGNATURES = 15;
```
willclarktech: @shuse2 Please confirm this value.

packages/lisk-transactions/src/utils/validation/validation.js
```diff
@@ -44,8 +48,13 @@ export const validatePublicKeys = publicKeys =>
 	checkPublicKeysForDuplicates(publicKeys);
 
 export const validateKeysgroup = keysgroup => {
-	if (keysgroup.length === 0 || keysgroup.length > 16) {
-		throw new Error('Expected between 1 and 16 public keys in the keysgroup.');
+	if (
+		keysgroup.length < MULTISIGNATURE_MIN_SIGNATURES ||
+		keysgroup.length > MULTISIGNATURE_MAX_SIGNATURES
```
willclarktech: @shuse2 There's a difference between the range allowed for number of signatures and the range allowed for keysgroup length, right?

packages/lisk-transactions/src/0_transfer.js
```diff
@@ -27,16 +62,19 @@ const createAsset = data => {
...
-const transfer = ({ amount, recipientId, recipientPublicKey, data }) => {
-	const { address, publicKey } = getAddressAndPublicKeyFromRecipientData({
-		recipientId,
-		recipientPublicKey,
-	});
+const transfer = inputs => {
+	validateInputs(inputs);
+	const { amount, recipientId, recipientPublicKey, data } = inputs;
+	let address = recipientId;
+	let publicKey = null;
```
willclarktech: Prefer a solution without `let`.

packages/lisk-transactions/src/3_cast_votes.js
```diff
@@ -21,9 +21,13 @@ import {
...
 } from './utils';
 
-const castVotes = ({ passphrase, votes = [], unvotes = [] }) => {
+const validateInputs = ({ votes, unvotes }) => {
 	validatePublicKeys([...votes, ...unvotes]);
+};
 
+const castVotes = inputs => {
+	const { passphrase, votes = [], unvotes = [] } = inputs;
+	validateInputs({ votes, unvotes });
```
willclarktech: Prefer a consistent approach here: `validateInputs(inputs)` followed by destructuring. (You can provide default values in the `validateInputs` function.)

packages/lisk-transactions/src/0_transfer.js
willclarktech: Issue created: https://github.com/LiskHQ/lisk-elements/issues/771

packages/lisk-transactions/src/5_create_dapp.js
willclarktech: If the schema check was sufficient we wouldn't need this issue at all. A thorough solution will include these keys. Feel free to create a follow-up issue if you think we should merge this first and address them separately.

packages/lisk-transactions/test/4_register_multisignature_account.js
```diff
@@ -312,6 +312,111 @@ describe('#registerMultisignatureAccount transaction', () => {
 				return Promise.resolve();
 			});
 
+			describe('validation errors:', () => {
```
willclarktech: Remove `:`

packages/lisk-transactions/test/utils/validation/validation.js
```diff
@@ -243,6 +245,12 @@ describe('validation', () => {
 		});
 	});
 
+	describe('#validateAmount', () => {
+		it('should return true when amount is a number and is not greater than MaxTransactionAmount', () => {
```
willclarktech: `maximum transaction amount`

packages/lisk-transactions/src/4_register_multisignature_account.js
shuse2: Yes!
On the other PR, i fixed it on schema check in opposite way (but same thing)
https://github.com/LiskHQ/lisk-elements/pull/770/files#diff-bc8696a3bc628a23ff3642d6f438a7e5R37

packages/lisk-transactions/src/constants.js
shuse2: so it should be ok
https://github.com/LiskHQ/lisk/blob/development/config/default/constants.js#L80-L92

packages/lisk-transactions/src/utils/validation/validation.js
shuse2: I think `signatures` is misleading. the name should reflect `keysgroup`

packages/lisk-transactions/src/4_register_multisignature_account.js
shuse2: https://github.com/LiskHQ/lisk/blob/development/logic/multisignature.js#L135-L136
This is correct according to the `core`

packages/lisk-transactions/src/0_transfer.js
willclarktech: Addressed.

packages/lisk-transactions/src/0_transfer.js
```diff
@@ -12,13 +12,54 @@
...
+	if (data && data.length > 0) {
+		return { data };
+	}
+	return {};
+};
+
+const validateInputs = ({ amount, recipientId, recipientPublicKey, data }) => {
+	if (!validateAmount(amount)) {
+		throw new Error(
+			'Please provide an amount. Expected a valid number in string format.',
```
willclarktech: We should be consistent with the structure of our messages. How about `Could not create transaction: Amount must be a valid number in string format.`? (Also applies to the other messages.)

packages/lisk-transactions/src/0_transfer.js
```diff
@@ -12,13 +12,54 @@
...
+
+	if (typeof recipientId !== 'undefined') {
+		validateAddress(recipientId);
+	}
+
+	if (typeof recipientPublicKey !== 'undefined') {
+		validatePublicKey(recipientPublicKey);
+	}
+
+	if (recipientId && recipientPublicKey) {
```
willclarktech: We could make this a bit more succinct:
```js
if (recipientId && recipientPublicKey && recipientId !== getAddressFromPublicKey(recipientPublicKey)) {
    throw new Error('...');
}
```

packages/lisk-transactions/src/2_register_delegate.js
```diff
@@ -12,17 +12,34 @@
...
-		delegate: {
-			username,
+const validateInputs = ({ username }) => {
+	if (!username || typeof username !== 'string') {
+		throw new Error('Please provide a username. Expected string.');
+	}
+
+	if (username.length > USERNAME_LENGTH) {
+		throw new Error(
+			`Username length does not match requirements. Expected to be less than ${USERNAME_LENGTH} characters.`,
```
willclarktech: `less than` -> `no more than`

packages/lisk-transactions/src/3_cast_votes.js
```diff
@@ -21,9 +21,30 @@ import {
...
+		publicKeys.push(...votes);
+	}
 
+	if (typeof unvotes !== 'undefined') {
+		if (!Array.isArray(unvotes)) {
+			throw new Error(
+				'Please provide a valid unvotes value. Expected an array.',
+			);
+		}
+		publicKeys.push(...unvotes);
```
willclarktech: As above.

packages/lisk-transactions/src/3_cast_votes.js
```diff
@@ -21,9 +21,30 @@ import {
...
 
-const castVotes = ({ passphrase, votes = [], unvotes = [] }) => {
-	validatePublicKeys([...votes, ...unvotes]);
+const validateInputs = ({ votes, unvotes }) => {
+	const publicKeys = [];
+	if (typeof votes !== 'undefined') {
+		if (!Array.isArray(votes)) {
+			throw new Error('Please provide a valid votes value. Expected an array.');
+		}
+		publicKeys.push(...votes);
```
willclarktech: Just validate them directly here. (`.push` is as bad as `let` 😉 )

packages/lisk-transactions/src/4_register_multisignature_account.js
```diff
@@ -12,16 +12,52 @@
...
+		minimum > MULTISIGNATURE_MAX_KEYSGROUP
+	) {
+		throw new Error(
+			`Please provide a valid minimum value. Expected integer between ${MULTISIGNATURE_MIN_KEYSGROUP} and ${MULTISIGNATURE_MAX_KEYSGROUP}.`,
+		);
+	}
+
+	if (keysgroup.length < minimum) {
+		throw new Error(
+			`Expected minimum ${minimum} public keys in the keysgroup.`,
```
willclarktech: This message makes an assumption about what went wrong. Something like this would be better: `Minimum number of signatures is larger than the number of keys in the keysgroup.`

packages/lisk-transactions/src/5_create_dapp.js
```diff
@@ -35,10 +35,23 @@ const validateOptions = options => {
 	if (typeof link !== 'string') {
 		throw new Error('Dapp link must be a string.');
 	}
+
+	if (typeof description !== 'undefined' && typeof description !== 'string') {
+		throw new Error('Dapp description must be a string.');
```
willclarktech: `if provided`

packages/lisk-transactions/src/5_create_dapp.js
```diff
@@ -35,10 +35,23 @@ const validateOptions = options => {
...
+	}
+
+	if (typeof tags !== 'undefined' && typeof tags !== 'string') {
+		throw new Error('Dapp tags must be a string.');
```
willclarktech: As above.

packages/lisk-transactions/src/5_create_dapp.js
```diff
@@ -35,10 +35,23 @@ const validateOptions = options => {
...
+	}
+
+	if (typeof icon !== 'undefined' && typeof icon !== 'string') {
+		throw new Error('Dapp icon must be a string.');
```
willclarktech: As above.

packages/lisk-transactions/src/3_cast_votes.js
```diff
@@ -21,9 +21,30 @@ import {
 	wrapTransactionCreator,
 } from './utils';
 
-const castVotes = ({ passphrase, votes = [], unvotes = [] }) => {
-	validatePublicKeys([...votes, ...unvotes]);
+const validateInputs = ({ votes, unvotes }) => {
+	const publicKeys = [];
+	if (typeof votes !== 'undefined') {
+		if (!Array.isArray(votes)) {
+			throw new Error('Please provide a valid votes value. Expected an array.');
```
willclarktech: `if present`

packages/lisk-transactions/src/3_cast_votes.js
```diff
@@ -21,9 +21,30 @@ import {
...
+		}
+		publicKeys.push(...votes);
+	}
 
+	if (typeof unvotes !== 'undefined') {
+		if (!Array.isArray(unvotes)) {
+			throw new Error(
+				'Please provide a valid unvotes value. Expected an array.',
```
willclarktech: `if present`

packages/lisk-transactions/src/constants.js
```diff
@@ -22,7 +22,12 @@ export const SIGNATURE_FEE = 5 * FIXED_POINT;
 export const DELEGATE_FEE = 25 * FIXED_POINT;
 export const VOTE_FEE = 1 * FIXED_POINT;
 export const MULTISIGNATURE_FEE = 5 * FIXED_POINT;
+export const MULTISIGNATURE_MAX_LIFETIME = 72;
+export const MULTISIGNATURE_MIN_LIFETIME = 1;
+export const MULTISIGNATURE_MAX_KEYSGROUP = 15;
+export const MULTISIGNATURE_MIN_KEYSGROUP = 1;
 export const DAPP_FEE = 25 * FIXED_POINT;
+export const USERNAME_LENGTH = 20;
```
willclarktech: `USERNAME_MAX_LENGTH`

packages/lisk-transactions/src/1_register_second_passphrase.js
```diff
@@ -16,7 +16,15 @@ import cryptography from 'lisk-cryptography';
 import { SIGNATURE_FEE } from './constants';
 import { wrapTransactionCreator } from './utils';
 
-const registerSecondPassphrase = ({ secondPassphrase }) => {
+const validateInputs = ({ secondPassphrase }) => {
+	if (typeof secondPassphrase !== 'string') {
+		throw new Error('Please provide a secondPassphrase. Expected string.');
```
willclarktech: Should we prefix every such error message with `Could not create transaction: `? It would be more consistent.

yatki: @willclarktech actually only transfer tx creator has `Could not create transaction: `. All the other error messages don't have this prefix. I'd prefer to remove it since we will get it during transfer creation anyway.

PR #778, opened 2018-09-07
:bug: Fix circular dependency in lisk-cryptography

### LINKED ISSUE #775: Remove circular dependency in lisk-cryptography
### Expected behavior

There should be no circular dependencies.

### Actual behavior

`convert` depends on `hash` and vice versa.

1.0.0+

mitsujutsu:
### What was the problem?

Circular dependency in lisk-cryptography, specifically the "hexToBuffer" function in convert module being used in hash module.

### How did I fix it?

Separated buffer specific logic into a new buffer module along with the corresponding tests.

### How to test it?

Install, build, lint, test etc.

### Review checklist
* The PR resolves #775
* Linting passes
* Tests pass
* Commit messages follow the commit guidelines

### DIFF
--- a/packages/lisk-cryptography/src/index.js
+++ b/packages/lisk-cryptography/src/index.js
@@ -12,10 +12,19 @@
  * Removal or modification of this copyright notice is prohibited.
  *
  */
+import * as buffer from './buffer';
 import * as convert from './convert';
 import * as encrypt from './encrypt';
 import hash from './hash';
 import * as keys from './keys';
 import * as sign from './sign';
 
-export default Object.assign({}, convert, encrypt, { hash }, keys, sign);
+export default Object.assign(
+	{},
+	buffer,
+	convert,
+	encrypt,
+	{ hash },
+	keys,
+	sign,
+);
--- a/packages/lisk-cryptography/test/buffer.js
+++ b/packages/lisk-cryptography/test/buffer.js
@@ -1,0 +1,125 @@
+/*
+ * Copyright © 2018 Lisk Foundation
+ *
+ * See the LICENSE file at the top-level directory of this distribution
+ * for licensing information.
+ *
+ * Unless otherwise agreed in a custom licensing agreement with the Lisk Foundation,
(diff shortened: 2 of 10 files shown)


packages/lisk-cryptography/src/index.js
```diff
@@ -12,10 +12,19 @@
...
  *
  */
+import * as buffer from './buffer';
 import * as convert from './convert';
 import * as encrypt from './encrypt';
 import hash from './hash';
 import * as keys from './keys';
 import * as sign from './sign';
 
-export default Object.assign({}, convert, encrypt, { hash }, keys, sign);
```
willclarktech: This is a breaking change, and requires a major release. @shuse2 thoughts?

packages/lisk-cryptography/test/buffer.js
```diff
@@ -0,0 +1,125 @@
...
+			);
+		});
+
+		it('should throw an error for an odd-length hex string input with custom argument name', () => {
+			return expect(hexToBuffer.bind(null, 'c3a5c3a4c3b6a', 'Custom')).to.throw(
+				'Custom must have a valid length of hex string.',
+			);
+		});
+
+		describe('#bigNumberToBuffer', () => {
```
willclarktech: These suites should be nested one level shallower.

packages/lisk-cryptography/src/index.js
shuse2: I think we should expose buffer function, but also make alias in the old format as well.

willclarktech: There should be no collisions because they've just been ripped out of `convert` directly.

willclarktech: So
```
export default Object.assign(
    {},
    // ...
    buffer,
    convert: Object.assign({}, buffer, convert),
    // ...
);
```
?

willclarktech: Sorry, my mistake, using `Object.assign` in this way means the export won't change. We can include this in a minor release just fine.

PR #782, opened 2018-09-13
 Add support for sodium-native library in cryptography - Closes #772

### LINKED ISSUE #772: Add support for sodium-native library in cryptography
### Description

For cryptography package, 
we should support [Sodium Native](https://github.com/sodium-friends/sodium-native) for server side usage which require high performance.

mitsujutsu:
### What was the problem?

For cryptography package, we were using tweetnacl library which is slower than sodium-native in benchmark tests.

### How did I fix it?

Added support for sodium-native library. Users can opt to use sodium-native with "NACL_FAST" environment variable. Sodium-native is listed in optionalDependencies and devDependencies in package.json. 

### How to test it?

Install, build, lint, test.

### Review checklist

* The PR resolves #772
* Linting passes
* Tests pass


package-lock.json
```diff
@@ -0,0 +1,3408 @@
+{
+	"name": "lisk-elements-monorepo",
```
shuse2: `package-lock.json` should not be committed as it's not officially npm5 yet

packages/lisk-cryptography/src/nacl/index.js
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ *
+ */
+let lib; /* eslint-disable-line */
+
+try {
+	if (process.env.NACL_FAST === 'disable') throw new Error('Use tweetnacl');
+	// Require used for conditional importing
+	lib = require('./sodium'); /* eslint-disable-line */
```
shuse2: I think we use `// eslint-disable-next-line` in other places

packages/lisk-cryptography/src/keys.js
```diff
@@ -12,29 +12,33 @@
...
+
 export const getPrivateAndPublicKeyBytesFromPassphrase = passphrase => {
 	const hashed = hash(passphrase, 'utf8');
-	const { publicKey, secretKey } = nacl.sign.keyPair.fromSeed(hashed);
+	const { publicKeyBytes, privateKeyBytes } = signKeyPair(hashed);
 
 	return {
-		privateKey: secretKey,
-		publicKey,
+		privateKeyBytes,
```
shuse2: Just a remark here as this is a breaking change, but considering the fact that it will be `v2.0`, and it's more clear what the content is, I think this would be a good change.

packages/lisk-cryptography/test/nacl.js
```diff
@@ -0,0 +1,238 @@
+/*
```
shuse2: I think folder structure should be the same as the `src`.
In this case, `test/nacl/nacl.js`

packages/lisk-cryptography/test/sodium.js
```diff
@@ -0,0 +1,234 @@
+/*
```
shuse2: Also, test file structure for this one too.

packages/lisk-cryptography/src/nacl/index.js
```diff
@@ -0,0 +1,26 @@
+/*
```
shuse2: We should have test for this `index.js` behavior too

packages/lisk-cryptography/test/nacl.js
```diff
@@ -0,0 +1,238 @@
...
+	boxOpen,
+} = nacl;
+
+const makeInvalid = str => {
+	const char = str[0] === '0' ? '1' : '0';
+	return `${char}${str.slice(1)}`;
+};
+
+describe('nacl', () => {
+	const CRYPTO_SIGN_PUBLICKEYBYTES = 32;
```
shuse2: maybe these one doesn't need to be capital since other constants are not capital?

packages/lisk-cryptography/src/index.js
```diff
@@ -12,6 +12,7 @@
  * Removal or modification of this copyright notice is prohibited.
  *
  */
+import * as nacl from './nacl';  /* eslint-disable-line */
```
shuse2: `nacl` not used in this file

packages/lisk-cryptography/test/nacl.js
```diff
@@ -0,0 +1,238 @@
...
+	describe('#randombytes', () => {
+		const size = 24;
+		let randomBuffer;
+		beforeEach(() => {
+			randomBuffer = randombytes(size);
+			return Promise.resolve();
+		});
+
+		it('should return a buffer', () => {
+			return expect(Buffer.from(randomBuffer)).to.be.instanceOf(Buffer);
```
willclarktech: This seems to be testing the native method `Buffer.from`, not our `randombytes` function.

packages/lisk-cryptography/test/nacl.js
```diff
@@ -0,0 +1,238 @@
...
+		'a232e5ea10e18249efc5a0aa8ed68271fc494d02245c52277ee2e14cddd960144a65';
+	const defaultNonce = 'df4c8b09e270d2cb3f7b3d53dfa8a6f3441ad3b14a13fb66';
+
+	let hashStub;
+	let digestMessageStub;
+	let convertPublicKeyEd2CurveStub;
+	let convertPrivateKeyEd2CurveStub;
+
+	beforeEach(() => {
+		hashStub = sandbox
```
willclarktech: Prefer stubbing without assignment and then asserting on the relevant method directly.

packages/lisk-cryptography/test/nacl.js
```diff
@@ -0,0 +1,238 @@
...
+			.stub(hash, 'default')
+			.returns(
+				Buffer.from(
+					'314852d7afb0d4c283692fef8a2cb40e30c7a5df2ed79994178c10ac168d6d97',
```
willclarktech: It's a little confusing that this is also the start of the default hash. Prefer totally separate data.

packages/lisk-cryptography/test/nacl.js
```diff
@@ -0,0 +1,238 @@
...
+		it('should return a buffer', () => {
+			return expect(Buffer.from(randomBuffer)).to.be.instanceOf(Buffer);
+		});
+	});
+
+	describe('#signKeyPair', () => {
+		let hashedSeed;
+		let signedKeys;
+		beforeEach(() => {
+			hashedSeed = hashStub();
```
willclarktech: If the hash was defined as a `const`, it could be used directly here as well as in the stub call. I think calling a stub at this point complicates things.

packages/lisk-cryptography/test/nacl.js
```diff
@@ -0,0 +1,238 @@
...
+			signedKeys = signKeyPair(Buffer.from(hashedSeed));
```
willclarktech: If we go with the above option, it probably makes more sense to define the hash as a string, but as things are currently the `hashStub` call already returns a `Buffer` so no need to create a new one here.

packages/lisk-cryptography/test/nacl.js
```diff
@@ -0,0 +1,238 @@
...
+			return Promise.resolve();
+		});
+
+		it('should create a buffer publicKey', () => {
+			return expect(
+				Buffer.from(signedKeys.publicKeyBytes).toString('hex'),
```
willclarktech: Again, if we use `Buffer.from` here, we're not really testing what the test title says we are.

packages/lisk-cryptography/src/nacl/nacl.js
```diff
@@ -0,0 +1,61 @@
...
+ * Removal or modification of this copyright notice is prohibited.
+ *
+ */
+import tweetnacl from 'tweetnacl';
+
+export const naclSignPublicKeyLength = 32;
+
+export const naclSignSignatureLength = 64;
+
+export const randombytes = data => tweetnacl.randomBytes(data);
```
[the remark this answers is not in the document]
shuse2: I think Im ok with either option, but if we decide to go with `getRandomBytes` other wrapper function also should be in the same style imo

packages/lisk-cryptography/test/nacl.js
```diff
@@ -0,0 +1,238 @@
...
+			).to.be.eql(defaultPrivateKey);
+		});
+
+		it('should return publicKeyBytes with buffer of length 32', () => {
+			return expect(signedKeys.publicKeyBytes.length).to.be.eql(
+				CRYPTO_SIGN_PUBLICKEYBYTES,
+			);
+		});
+
+		it('should return privateKeyBytes with buffer of length 64', () => {
```
willclarktech: These tests are kind of redundant given the direct comparisons in the previous test, but I guess they're useful as a coherence check.

packages/lisk-cryptography/test/nacl.js
```diff
@@ -0,0 +1,238 @@
...
+				);
+				return Promise.resolve();
+			});
+
+			it('should decrypt a message', () => {
+				return expect(
+					Buffer.from(decryptedMessageBytes).toString('utf8'),
+				).to.be.eql(defaultMessage);
+			});
+		});
```
willclarktech: It would be nice to have an integration test to make sure the two play together nicely. Also potentially for sign/verify.

packages/lisk-cryptography/src/nacl/nacl.js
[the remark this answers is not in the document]
willclarktech: Yep, either all should be Lisk-style or all should be NaCl-style.

packages/lisk-cryptography/src/encrypt.js
```diff
@@ -29,23 +30,24 @@ export const encryptMessageWithPassphrase = (
...
 	const convertedPrivateKey = convertPrivateKeyEd2Curve(senderPrivateKeyBytes);
 	const recipientPublicKeyBytes = hexToBuffer(recipientPublicKey);
 	const convertedPublicKey = convertPublicKeyEd2Curve(recipientPublicKeyBytes);
 	const messageInBytes = Buffer.from(message, 'utf8');
 
-	const nonce = nacl.randomBytes(24);
-	const cipherBytes = nacl.box(
+	const nonce = randombytes(24);
+
+	const cipherText = box(
```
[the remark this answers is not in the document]
mitsujutsu: "cipherText" conforms to the naming in sodium-native library. I agree "cipherBytes" may be more descriptive.

packages/lisk-cryptography/src/nacl/nacl.js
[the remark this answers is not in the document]
mitsujutsu: I think getRandomBytes here may be clearer as well.

packages/lisk-cryptography/test/nacl.js
mitsujutsu: `Buffer.from` is used here as a way to easily convert the returned `Uint8Array` into a hex string, so that it can be directly compared to defaultPublicKey.  Rename the test to just `should create a publicKey` and leave the conversion as is?  @willclarktech

willclarktech: Maybe the most direct way is to assert on the type and keep the assertion you have now for the content. How does that sound?

packages/lisk-cryptography/src/nacl/index.js
```diff
@@ -0,0 +1,45 @@
...
+	process.env.NACL_FAST = 'disable';
+	// eslint-disable-next-line
+	lib = require('./slow');
+}
+
+export const NACL_SIGN_PUBLICKEY_LENGTH = 32;
+
+export const NACL_SIGN_SIGNATURE_LENGTH = 64;
+
+export const { box } = lib;
```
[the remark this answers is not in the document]
mitsujutsu: Exporting named constants here to allow importing specific functions, instead of destructing a default object each time.

packages/lisk-cryptography/test/helpers/index.js
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * Unless otherwise agreed in a custom licensing agreement with the Lisk Foundation,
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ *
+ */
+// eslint-disable-next-line
```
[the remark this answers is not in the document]
mitsujutsu: Yes it's for prefer default export

packages/lisk-cryptography/test/nacl/slow.js
```diff
@@ -0,0 +1,199 @@
...
+
+			it('should decrypt a message', () => {
+				return expect(
+					Buffer.from(decryptedMessageBytes).toString('utf8'),
+				).to.be.eql(defaultMessage);
+			});
+		});
+	});
+
+	describe('integration test', () => {
```
willclarktech: `integration tests`

packages/lisk-cryptography/src/encrypt.js
```diff
@@ -71,16 +72,20 @@ export const decryptMessageWithPassphrase = (
...
 		return Buffer.from(decoded).toString();
 	} catch (error) {
-		if (error.message.match(/bad nonce size/)) {
-			throw new Error('Expected 24-byte nonce but got length 1.');
+		if (
+			error.message.match(
+				/bad nonce size|nonce must be a buffer of size crypto_box_NONCEBYTES/,
+			)
+		) {
+			throw new Error('Nonce must be a buffer of size 24.');
```
willclarktech: This isn't true: we convert from a hex string. How about `Expected nonce to be 24 bytes.`?

packages/lisk-cryptography/src/nacl/slow.js
```diff
@@ -0,0 +1,34 @@
...
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ *
+ */
+import tweetnacl from 'tweetnacl';
+
+// eslint-disable-next-line prefer-destructuring
+export const box = tweetnacl.box;
```
willclarktech: Oh sorry, didn't realise this was going to require us to disable a linting rule. I would say conforming to lint rules trumps this kind of uniformity.

packages/lisk-cryptography/test/nacl/index.js
```diff
@@ -0,0 +1,130 @@
...
+
+	describe('nacl fast installed', () => {
+		beforeEach(() => {
+			resetTest();
+			return Promise.resolve();
+		});
+
+		it('should load nacl fast if process.env.NACL_FAST is set to enable', () => {
+			process.env.NACL_FAST = 'enable';
+			// eslint-disable-next-line global-require
```
willclarktech: It probably makes more sense to turn this off for the whole file, or at least for a block rather than individual lines.

packages/lisk-cryptography/test/nacl/nacl.js
```diff
@@ -0,0 +1,219 @@
...
+	{
+		name: 'fast',
+		fast,
+	},
+	{
+		name: 'slow',
+		slow,
+	},
+].forEach(nacl => {
+	describe(`nacl ${nacl.name}`, () => {
```
willclarktech: How about an outer suite `nacl` and `${nacl.name}` in the inner suites?

packages/lisk-cryptography/test/nacl/index.js
```diff
@@ -0,0 +1,130 @@
...
+import * as fast from '../../src/nacl/fast';
+import * as slow from '../../src/nacl/slow';
+// Require is used for stubbing
+const moduleLibrary = require('module');
+
+const resetTest = () => {
+	// Reset environment variable
+	delete process.env.NACL_FAST;
+	// Delete require cache to force it to re-load module
+	delete require.cache[require.resolve('../../src/nacl/index')];
```
willclarktech: No need for `/index` in this path.

packages/lisk-cryptography/test/nacl/index.js
```diff
@@ -0,0 +1,130 @@
...
+};
+
+const stripConstants = library => {
+	// Constants are added in /src/index.js
```
willclarktech: This isn't the correct path.

PR #787, opened 2018-09-20
Add TypeScript setup - Closes #781

### LINKED ISSUE #781: Set up TypeScript for lisk-elements
Child of #417 

### Expected behavior

We should have a monorepo-friendly TypeScript setup so we can convert existing packages to TypeScript.

### Actual behavior

We have no such setup.

2.0.0

willclarktech:
### What was the problem?

We needed some TypeScript setup.

### How did I fix it?

Added some.

### How to test it?

In `lisk-constants` try building, linting etc.

### Review checklist

* The PR resolves #781 
* Linting passes
* Tests pass


shuse2:
I will disable `test:node` to check only build file exists as before. In order to achieve this, we need to keep babel or add other build system.
Since it will be out of scope and too big PR, I will create another issue.

### Reason

For `test:node` command, I didn't find a way to both achieve
- test against `src`, test against `dist-node` and browser test.
Only way it seems possible by changing `tsconfig` is
1. test against `src` and test against `dist-node`
or
2. test against `src` and browser test

To test 1, we can use `paths` in `tsconfig` to use alias in the test we use `@liskhq/lisk-constants/index`, and switch the root folder.
However, typescript doesn't resolve the path after build, so it will not work in browser test. https://github.com/Microsoft/TypeScript/issues/10866

packages/lisk-constants/src/index.ts
```diff
@@ -12,9 +12,11 @@
  * Removal or modification of this copyright notice is prohibited.
  *
  */
+// tslint:disable-next-line:no-magic-numbers
 export const EPOCH_TIME = new Date(Date.UTC(2016, 4, 24, 17, 0, 0, 0));
 export const EPOCH_TIME_MILLISECONDS = EPOCH_TIME.getTime();
-export const EPOCH_TIME_SECONDS = Math.floor(EPOCH_TIME.getTime() / 1000);
+const MS_FACTOR = 1000;
```
yatki: 👍

tslint.json
```diff
@@ -0,0 +1,26 @@
+{
```
mitsujutsu: `npm run lint` produces `no-unused-variable is deprecated` warning. Maybe related to: [https://github.com/palantir/tslint/issues/4046](https://github.com/palantir/tslint/issues/4046)

shuse2: yes, it should be fixed by the next version of tslint

mitsuaki-u (dismissed):

packages/lisk-constants/test/index.ts
```diff
@@ -22,7 +23,7 @@ import {
 	TESTNET_NETHASH,
 	MAINNET_NETHASH,
 	SIGNED_MESSAGE_PREFIX,
-} from '../src';
+} from '../src/index';
```
yatki: I prefer to keep index out. Is there a reason that you added it ?

packages/lisk-constants/test/tslint.json
```diff
@@ -0,0 +1,3 @@
+{
+	"extends": ["../../../tslint.test.json"]
```
yatki: this file could be a template? we can also create a symlink for the root folder.

packages/lisk-constants/tsconfig.browsertest.json
```diff
@@ -0,0 +1,13 @@
...
+	"compilerOptions": {
+		"declaration": false,
+		"outDir": "browsertest.build",
+		"baseUrl": ".",
+		"rootDir": ".",
+		"paths": {
+			"chai": ["../../types/chai"]
+		}
+	},
+	"include": ["../../types/**/*", "src/**/*", "test/**/*"]
```
yatki: I prefer to have this as a template file as well. Seems like we are gonna use this.

packages/lisk-constants/tsconfig.json
```diff
@@ -0,0 +1,9 @@
+{
+	"extends": "../../tsconfig",
+	"compilerOptions": {
+		"declaration": true,
+		"outDir": "dist-node",
+		"rootDir": "src"
+	},
+	"include": ["src/**/*"]
```
yatki: same as above.

packages/lisk-constants/tslint.json
```diff
@@ -0,0 +1,3 @@
+{
+	"extends": "../../tslint.json"
```
yatki: same as above.

tsconfig.json
```diff
@@ -0,0 +1,18 @@
...
+		"newLine": "lf",
+		"noFallthroughCasesInSwitch": true,
+		"noImplicitReturns": true,
+		"noUnusedLocals": true,
+		"noUnusedParameters": true,
+		"pretty": true,
+		"removeComments": true,
+		"sourceMap": true,
+		"strict": true
+	}
```
yatki: I'm not sure about this but does it make sense to have this as template, instead of extending we can use the symlink?

yatki (dismissed):
looks good in general but maybe having template files would be better.

packages/lisk-constants/test/index.ts
shuse2: it seems in typescript, u need to specify index

yatki: It shouldn't be like that. http://www.typescriptlang.org/docs/handbook/module-resolution.html  (How TypeScript resolves modules section). I don't get any linter or build errors when I remove it.

tsconfig.json
shuse2: I think having the global setting and extending if they need to would be probably better considering in the future there will be lots of child packages and not all the project will have the same setup

yatki (approved):

mitsuaki-u (approved):

PR #790, opened 2018-09-27
Update package name

### LINKED ISSUE #788: Update child packages names
## Description
After moving to mono-repo, we need to update the naming of the child packages for npm publication.

- lisk-api-client => @liskhq/lisk-api-client
- lisk-contants => @liskhq/lisk-constants
- lisk-cryptography => @liskhq/lisk-cryptography
- lisk-passphrase => @liskhq/lisk-passphrase
- lisk-transactions => @liskhq/lisk-transactions

and `lisk-elements` will stay as it is.
Also, we should add `@liskhq/lisk-client` which is currently exactly the same as `lisk-elements` but it will only contain the packages regards to the `web` and `mobile` in the future.

shuse2:
### What was the problem?
After moving to monorepo, package naming strategy needed to be applied.
Also, separate `@liskhq/lisk-client` is created for client related packages.

### Review checklist

* The PR resolves #788
* Linting passes
* Tests pass


packages/lisk-client/scripts/prestart.sh
```diff
@@ -0,0 +1,5 @@
+read -r -p $'\e[96mDo you want to build Lisk Elements first? [y/N]\e[0m ' should_build
```
willclarktech: Rename

packages/lisk-client/test/index.js
```diff
@@ -0,0 +1,43 @@
...
+ */
+import {
+	APIClient,
+	constants,
+	cryptography,
+	passphrase,
+	transaction,
+} from '../src';
+
+describe('lisk-elements', () => {
```
willclarktech: Rename

packages/lisk-client/package.json
```diff
@@ -0,0 +1,94 @@
+{
+	"name": "@liskhq/lisk-client",
+	"version": "1.0.0",
+	"description": "A default set of elements for use by clients of the Lisk network",
```
willclarktech: I think "Elements" is branded, therefore should be capitalised.

packages/lisk-client/index.html
```diff
@@ -0,0 +1,14 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <title>Title</title>
+</head>
+<body>
+Open the console to get started with Lisk Client`.
```
willclarktech: Rogue `.

packages/lisk-client/README.md
```diff
@@ -0,0 +1 @@
+# lisk-client
```
willclarktech: Let's create a separate issue to add proper READMEs to all the packages.

packages/lisk-client/.gitignore
```diff
@@ -0,0 +1,17 @@
+# cypress media
```
willclarktech: Maybe this should be a template now since it's used in two places?

packages/lisk-client/package.json
```diff
@@ -0,0 +1,94 @@
+{
```
willclarktech: How about adding the other `package-lock.json`s at this point?

shuse2: in this branch, I think we are still at npm3 so no point to include `package-lock.json` now?

PR #794, opened 2018-10-01
Disallow 0 amounts for transfer transactions - Closes #783

### LINKED ISSUE #783: Disallow 0 amounts for transfer transactions
### Expected behavior

Transfer transactions should not allow an amount of 0.

### Actual behavior

We don't have a special check for transfer transactions. Amount validation currently allows 0 for all transaction types.

### Steps to reproduce

Create a transfer transaction with amount 0 and try to broadcast it to the network.

1.0.0+

mitsujutsu:
### What was the problem?

Transfer transactions (type 0) should not allow an amount of 0.

### How did I fix it?

Add validator for transfer transaction amount. Enforce 0 amount for other transaction types in schema. 

### How to test it?

Create a transfer transaction with amount 0.

### Review checklist

* The PR resolves #783
* Linting passes
* Tests pass

### DIFF
--- a/packages/lisk-transactions/src/utils/validation/validation.js
+++ b/packages/lisk-transactions/src/utils/validation/validation.js
@@ -83,7 +83,7 @@ export const validateAddress = address => {
 	return true;
 };
 
-export const isGreaterThanZero = amount => amount > 0;
+export const isGreaterThanZero = amount => amount.cmp(0) > 0;
 
 export const isGreaterThanMaxTransactionAmount = amount =>
 	amount.cmp(MAX_TRANSACTION_AMOUNT) > 0;
--- a/packages/lisk-transactions/src/utils/validation/validation.js
+++ b/packages/lisk-transactions/src/utils/validation/validation.js
@@ -83,6 +83,8 @@ export const validateAddress = address => {
 	return true;
 };
 
+export const isGreaterThanZero = amount => amount > 0;
+
 export const isGreaterThanMaxTransactionAmount = amount =>
 	amount.cmp(MAX_TRANSACTION_AMOUNT) > 0;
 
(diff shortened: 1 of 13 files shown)


packages/lisk-transactions/src/utils/validation/validation.js
```diff
@@ -83,6 +83,8 @@ export const validateAddress = address => {
 	return true;
 };
 
+export const isGreaterThanZero = amount => amount > 0;
```
shuse2: This would be `bignum` instance so it should compare using the bignum function

packages/lisk-transactions/test/utils/validation/validate_transaction.js
```diff
@@ -54,7 +54,7 @@ describe('validateTransaction', () => {
 		describe('when the transaction contains invalid data in merged schema', () => {
 			const invalidTransaction = {
 				type: 0,
-				amount: '0',
+				amount: '1',
```
shuse2: Probably it's better to add `transferAmount` test here too, because it's also invalid merged schema pattern

packages/lisk-transactions/test/utils/validation/validator.js
```diff
@@ -189,11 +189,55 @@ describe('validator', () => {
 		});
 
 		it('should validate to true when valid amount is provided', () => {
-			return expect(validate({ target: '190105310' })).to.be.true;
+			return expect(validate({ target: '0' })).to.be.true;
 		});
 
 		it('should validate to true when valid amount with leading zeros is provided', () => {
-			return expect(validate({ target: '00190105310' })).to.be.true;
+			return expect(validate({ target: '000000' })).to.be.true;
```
shuse2: I think target number should be non-zero value with leading zero

packages/lisk-transactions/test/utils/validation/validator.js
```diff
@@ -189,11 +189,55 @@ describe('validator', () => {
...
+		});
+
+		it('should validate to true when valid amount is provided', () => {
+			return expect(validate({ target: '100' })).to.be.true;
 		});
 
 		it('should validate to true when valid amount with leading zeros is provided', () => {
-			return expect(validate({ target: '00190105310' })).to.be.true;
+			return expect(validate({ target: '000000100' })).to.be.true;
 		});
```
shuse2: Lets add `validate({ target: '0' }).to.be.false;`?

PR #830, opened 2018-10-23
Move all dev dependencies to the root level - Closes #805

### LINKED ISSUE #805: Consider package duplication between root and child packages
### Description
Currently we use `types` folder in root directory to solve `chai` extra types.
However, while converting the packages, I found out it will sometime require type definition on root `package.json`, too. 
Therefore, we should consider making it not duplicate

shuse2:
### What was the problem?
A lot of duplicate devDependencies were in each packages, and it took so long to install, and often failing on some part of npm

### How did I fix it?
Move all the devDependencies to the root level, and add `--hoist` flag to the `bootstrap` command

### How to test it?
```
npm run clean
npm run clean:node_modules
rm -rf node_modules
npm i
npm run build
npm test
```

### Review checklist

* The PR resolves #805
* Linting passes
* Tests pass


package.json
```diff
@@ -42,13 +42,44 @@
 		"prepush": "npm run lint && npm test"
 	},
 	"devDependencies": {
+		"@types/chai": "4.1.5",
+		"@types/expect": "1.20.3",
+		"@types/jquery": "3.3.9",
+		"@types/mocha": "5.2.5",
+		"@types/node": "10.10.1",
```
mitsujutsu: Also include "types/chai-as-promised":"7.1.0", "types/sinon":"5.0.5", and "types/sinon-chai":"3.2.0"

mitsuaki-u (changes requested):
additional devDependencies in root

package.json
ishantiw: Also, we need `"@types/bip39": "2.4.0"` for passphrase

ishantiw (changes requested):

package.json
shuse2: Are these already exists in `development` branch?

mitsujutsu: Not yet but in near future PRs for typescript. Add now or later?

shuse2: I think it make sense to add them when it requires.
When we the new PR are merged, anyway we need to clean up the `package.json` anyway.

Also, if the another PR goes in first, then we can do opposite

mitsuaki-u (dismissed):

PR #840, opened 2018-11-12
Remove outdated zero transfer amount test - Closes #837

### LINKED ISSUE #837: Update test to check reject amount zero for transfer transaction
### Expected behavior

Type 0 transactinos with amount 0 should be rejected to sign, and the test also should be updated

https://github.com/LiskHQ/lisk-elements/blob/development/packages/lisk-transactions/test/utils/get_transaction_bytes.js#L110

### Actual behavior
Amount zero is accepted

1.0-

mitsujutsu:
### What was the problem?

`getTransactionBytes` test suite had outdated test to check for allowing 0 transfer transactions.

### How did I fix it?

Removed outdated test

### How to test it?

npm run test

### Review checklist

* The PR resolves #837 
* Linting passes
* Tests pass

### DIFF
--- a/packages/lisk-transactions/test/utils/get_transaction_bytes.ts
+++ b/packages/lisk-transactions/test/utils/get_transaction_bytes.ts
@@ -125,17 +125,6 @@ describe('getTransactionBytes module', () => {
 				);
 			});
 
-			it('should not throw on type 0 with an amount that is 0', () => {
-				const amount = 0;
-				const transaction = {
-					...defaultTransaction,
-					amount,
-				};
-				return expect(
-					getTransactionBytes.bind(null, transaction),
-				).not.to.throw();
-			});
-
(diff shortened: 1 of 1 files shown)


packages/lisk-transactions/test/utils/get_transaction_bytes.ts
```diff
@@ -125,17 +125,6 @@ describe('getTransactionBytes module', () => {
 				);
 			});
 
-			it('should not throw on type 0 with an amount that is 0', () => {
```
shuse2: should we check type 0 with amount 0 throws an error?

mitsujutsu: I think it's not necessary since this check already exists in 0_transfer tests, and is outside the scope of this func.

PR #846, opened 2018-11-13
Add custom error with errno in api-client - Closes #791

### LINKED ISSUE #791: Add Error.code for API response error
### Expected behavior
API response error should also return `Error.code` and `Error.errno` to be handled by interface

### Actual behavior
It does not have the error code

1.0-

mitsujutsu:
### What was the problem?

APIClient was not returning status code in errors.

### How did I fix it?

Created new class CustomError which extends VError lib. New class has property `errno`

### How to test it?

npm test

### Review checklist

* The PR resolves #791
* Linting passes
* Tests pass


packages/lisk-api-client/src/errors.ts
```diff
@@ -0,0 +1,9 @@
+import { VError } from 'verror';
+
+export class CustomError extends VError {
```
shuse2: maybe `APIError` or `APIResponseError`?

packages/lisk-api-client/src/errors.ts
```diff
@@ -0,0 +1,9 @@
...
+	public errno: number;
+	public constructor(message: string, errno: number) {
```
shuse2: would it be better to make it optional and if not present add `500`?

packages/lisk-api-client/test/api_resource.ts
```diff
@@ -134,7 +133,8 @@ describe('API resource module', () => {
 						data: sendRequestResult,
 					},
 				});
-				return resource.request(defaultRequest, true).catch(err => {
+
+				return resource.request(defaultRequest, false).catch(err => {
```
shuse2: why change the repeat flag? (also below)

mitsujutsu: with the repeat=true it was not actually reaching the catch block and thus err.message was not really being tested.

packages/lisk-api-client/src/errors.ts
```diff
@@ -0,0 +1,11 @@
+import { VError } from 'verror';
+
+const defaultErrorNo = 500;
+
+export class APIError extends VError {
```
shuse2: lets add simple test for this `APIError`

packages/lisk-api-client/test/errors.ts
```diff
@@ -0,0 +1,58 @@
...
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ *
+ */
+import { expect } from 'chai';
+import { APIError } from '../src/errors';
+
+let apiError: APIError;
```
shuse2: these declarations should be inside of `describe`

packages/lisk-api-client/src/errors.ts
```diff
@@ -0,0 +1,11 @@
...
+	public errno?: number;
+	public constructor(message?: string, errno?: number) {
+		super(message || '');
+		this.errno = errno || defaultErrorNo;
+	}
```
ishantiw: We should also override `name` property to something like `this.name = 'API Error'`. So, in this case, it will give out stack something like, `{ API Error: An unknown error occurred...` instead of `{ VError: An unknown error occurred...`

PR #865, opened 2018-11-20
Create base transaction class - Closes #874

### LINKED ISSUE #874: Add BaseTransaction implementation
### Description
Currently in elements it only supports creation of the transaction, but it should add all the features related to the transaction.
Also, developer should be able to easily extend the BaseTransaction to create new type of transaction.

- validate
- verify
- apply
- undo
- create
- verifyAgainstTransactions

### Child issues
- [x] create draft implementation
- [x] Add Base Transaction `checkSchema` #941 
- [x] Add Base Transaction `checkTypes` #944 
- [x] Add Base Transaction `validate` #942
- [x] Add Base Transaction `apply` #945 
- [x] Add Base Transaction `undo` #946 
- [x] Add Base Transaction `isExpired` #947

mitsujutsu:
### Description

Implementation of base transaction class for lisk-transactions library based on [LIP-0006](https://github.com/LiskHQ/lips-staging/blob/master/proposals/lip-0006.md)

* The PR resolves #874 #947 #946 #945 #944 #943 #942 #941 
* Tests pass


packages/lisk-transactions/src/base_transaction.ts
```diff
@@ -0,0 +1,136 @@
...
+	public senderPublicKey: string;
+	public signature: string;
+	public signatures?: ReadonlyArray<string>;
+	public signSignature?: string;
+	public timestamp: number;
+	public transactionJSON: ITransactionJSON;
+	public type: number;
+
+	public constructor(rawTransaction: ITransactionJSON) {
+		this.amount = new BigNum(rawTransaction.amount);
```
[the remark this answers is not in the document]
mitsujutsu: after discussion with @SargeKhan, we will allow valid string and bignum for amount and fee inputs.

packages/lisk-transactions/src/base_transaction.ts
```diff
@@ -0,0 +1,263 @@
...
+		const signedTransaction = {
+			...singleSignedTransaction,
+			signSignature: this.signSignature,
+		};
+
+		return signedTransaction;
+	}
+
+	public getBytes(): Buffer {
+		const { signature, signSignature } = this.toJSON();
```
shuse2: why signature and signSignature needs to call `toJSON`?

packages/lisk-transactions/src/base_transaction.ts
```diff
@@ -0,0 +1,263 @@
...
+			? cryptography.hexToBuffer(signSignature)
+			: Buffer.alloc(0);
+
+		return Buffer.concat([
+			transactionType,
+			transactionTimestamp,
+			transactionSenderPublicKey,
+			transactionRecipientID,
+			transactionAmount,
+			transactionSignature,
```
shuse2: I think you shouldn't include `signature` and `signSignature` here because in between you need to put asset bytes

packages/lisk-transactions/src/base_transaction.ts
```diff
@@ -0,0 +1,263 @@
...
+			transactionSecondSignature,
+		]);
+	}
+
+	public abstract containsUniqueData(): boolean;
+
+	public validate(): ValidateReturn {
+		const transaction = this.toJSON();
+
+		if (!TRANSACTION_TYPES.includes(transaction.type)) {
```
shuse2: if we validate type by known ID, I think it will limit us not being able to add custom transaction.
Maybe type check should be done in extended class?

packages/lisk-transactions/src/base_transaction.ts
```diff
@@ -0,0 +1,263 @@
...
+
+		return {
+			validated: valid && verified,
+			errors: transactionErrors,
+		};
+	}
+
+	public getRequiredAttributes(): object {
+		return {
+			ACCOUNTS: [cryptography.getAddressFromPublicKey(this.senderPublicKey)],
```
shuse2: this is TBD right?

packages/lisk-transactions/src/base_transaction.ts
```diff
@@ -0,0 +1,263 @@
...
+		};
+	}
+
+	public verifyAgainstState(sender: Account): VerifyReturn {
+		// Check sender balance
+		const { verified: balanceVerified, errors } = checkBalance(
+			sender,
+			this.fee,
```
shuse2: this should be fee + amount?

packages/lisk-transactions/src/base_transaction.ts
```diff
@@ -0,0 +1,263 @@
...
+		);
+
+		// Check multisig
+		const multisigVerified = sender.secondPublicKey
```
shuse2: this is checking second-signature, not multi-signature.
But, also it should check multi-signature

packages/lisk-transactions/src/base_transaction.ts
```diff
@@ -0,0 +1,263 @@
...
+		};
+	}
+
+	public getRequiredAttributes(): object {
+		return {
+			ACCOUNTS: [cryptography.getAddressFromPublicKey(this.senderPublicKey)],
+		};
+	}
+
+	public verifyAgainstState(sender: Account): VerifyReturn {
```
shuse2: maybe it's better to have name `verify`?

packages/lisk-transactions/src/base_transaction.ts
```diff
@@ -0,0 +1,263 @@
...
+	): VerifyReturn;
+
+	public apply(sender: Account): StateReturn {
+		if (this.applied) {
+			return {
+				sender,
+			};
+		}
+
+		const updatedBalance = new BigNum(sender.balance).sub(this.fee);
```
shuse2: this.fee + this.amount

packages/lisk-transactions/src/base_transaction.ts
```diff
@@ -0,0 +1,263 @@
...
+			errors,
+		};
+	}
+
+	public abstract verifyAgainstOtherTransactions(
+		transactions: ReadonlyArray<TransactionJSON>,
+	): VerifyReturn;
+
+	public apply(sender: Account): StateReturn {
+		if (this.applied) {
```
shuse2: why do we need `this.applied`?

packages/lisk-transactions/src/base_transaction.ts
```diff
@@ -0,0 +1,263 @@
...
+		};
+	}
+
+	public undo(sender: Account): StateReturn {
+		if (!this.applied) {
+			return {
+				sender,
+			};
+		}
+		const updatedBalance = new BigNum(sender.balance).add(this.fee);
```
shuse2: this.fee + this.amount

packages/lisk-transactions/src/base_transaction.ts
```diff
@@ -0,0 +1,263 @@
...
+			senderPublicKey: this.senderPublicKey,
+			senderId: this.senderId,
+			recipientId: this.recipientId,
+			recipientPublicKey: this.recipientPublicKey,
+			fee: this.fee.toString(),
+			signatures: this.signatures,
+			asset: this.asset,
+		};
+
+		if (!(typeof this.signature === 'string' && this.signature.length > 0)) {
```
shuse2: what is the reason to check in this way?

packages/lisk-transactions/src/base_transaction.ts
```diff
@@ -0,0 +1,263 @@
...
+			return transaction;
+		}
+
+		const singleSignedTransaction = {
+			...transaction,
+			signature: this.signature,
+		};
+
+		if (
+			!(typeof this.signSignature === 'string' && this.signSignature.length > 0)
```
shuse2: same as above

packages/lisk-transactions/src/base_transaction.ts
```diff
@@ -0,0 +1,263 @@
...
+			amount: this.amount.toString(),
+			type: this.type,
+			timestamp: this.timestamp,
+			senderPublicKey: this.senderPublicKey,
+			senderId: this.senderId,
+			recipientId: this.recipientId,
+			recipientPublicKey: this.recipientPublicKey,
+			fee: this.fee.toString(),
+			signatures: this.signatures,
+			asset: this.asset,
```
shuse2: how do we deserialize this if it contains Bignum etc?

packages/lisk-transactions/src/base_transaction.ts
```diff
@@ -0,0 +1,263 @@
...
+							`'${error.dataPath}' ${error.message}`,
+							transaction.id,
+							error.dataPath,
+						),
+			  )
+			: undefined;
+
+		// Single signature validation
+		if (!transaction.signature) {
+			return {
```
shuse2: why are we returning single error here?

packages/lisk-transactions/src/transaction_types.ts
```diff
@@ -29,9 +54,24 @@ export interface BaseTransaction {
...
+	readonly errors?: ReadonlyArray<TransactionError>;
+	readonly validated: boolean;
+}
+
+export interface StateReturn {
+	readonly recipient?: Account;
+	readonly sender: Account;
+}
+
 type Partial<T> = { [P in keyof T]?: T[P] };
```
shuse2: it's not directly related to this PR, but we can remove this.
`Partial` is actually defined in typescript itself

packages/lisk-transactions/src/utils/check_balance.ts
```diff
@@ -0,0 +1,43 @@
...
+	return {
+		verified: !exceeded,
+		errors: exceeded
+			? [
+					new TransactionError(
+						`Account does not have enough LSK: ${
+							account.address
+						} balance: ${new BigNum(account.balance.toString() || '0').div(
+							// tslint:disable-next-line:no-magic-numbers
+							Math.pow(10, 8),
```
shuse2: this value was in constants

packages/lisk-transactions/src/utils/check_balance.ts
```diff
@@ -0,0 +1,43 @@
...
+	const exceeded = new BigNum(account.balance).lt(amount);
+
+	return {
+		verified: !exceeded,
+		errors: exceeded
+			? [
+					new TransactionError(
+						`Account does not have enough LSK: ${
+							account.address
+						} balance: ${new BigNum(account.balance.toString() || '0').div(
```
shuse2: when is `account.balance.toString()` false?

packages/lisk-transactions/src/utils/validation/validation.ts
```diff
@@ -136,3 +138,39 @@ export const validateFee = (data: string) =>
 
 export const isValidInteger = (num: unknown) =>
 	typeof num === 'number' ? Math.floor(num) === num : false;
+
+export const normalizeInput = (rawTransaction: TransactionJSON): void => {
```
shuse2: isn't this checked in validateTransaction?
also, it doesn't seem like this function is `normalizing` input

packages/lisk-transactions/src/base_transaction.ts
```diff
@@ -0,0 +1,263 @@
...
+		this.senderId = rawTransaction.senderId;
+		this.senderPublicKey = rawTransaction.senderPublicKey;
+		this.signature = rawTransaction.signature;
+		this.signatures = rawTransaction.signatures;
+		this.signSignature = rawTransaction.signSignature;
+		this.timestamp = rawTransaction.timestamp;
+		this.type = rawTransaction.type;
+	}
+
+	public abstract prepareTransaction(
```
shuse2: I think it's ok to remove this.
we just need to implement `public static create(xxx): TransactionJSON` function for each type of transaction i think

packages/lisk-transactions/src/utils/validation/validation.ts
```diff
@@ -136,3 +138,39 @@ export const validateFee = (data: string) =>
...
+			!((value as unknown) instanceof BigNum) &&
+			(!isNumberString(value) || !Number.isSafeInteger(parseInt(value, 10)))
+		) {
+			throw new TransactionError(
+				`\`${key}\` must be a valid string or BigNum.`,
+			);
+		}
+	});
+
+	Object.entries(strippedTransaction).forEach(field => {
```
shuse2: you can do `forEach([key, value] => {`

packages/lisk-transactions/src/base_transaction.ts
mitsujutsu: Bignum is converted to string as seen in `amount: this.amount.toString()` and `this.fee.toString()`

packages/lisk-transactions/src/base_transaction.ts
mitsujutsu: transactions in transit may have empty string `''` for signature and signSignature. this is checking for empty strings and only adding to result json if it's an actual signature.

packages/lisk-transactions/src/base_transaction.ts
mitsujutsu: @SargeKhan this is ok for now?

packages/lisk-transactions/src/base_transaction.ts
mitsujutsu: in base transaction it should be just fee, with amount in specific type classes?

packages/lisk-transactions/src/base_transaction.ts
mitsujutsu: so it can only be called once, until `undo` is called.

packages/lisk-transactions/src/base_transaction.ts
mitsujutsu: amount should be added in Transfer 1 class?

packages/lisk-transactions/src/base_transaction.ts
mitsujutsu: should we check only second-sig here and multi-sig in subclasses?

packages/lisk-transactions/src/utils/validation/validation.ts
mitsujutsu: in order to set `amount` and `fee` to `Bignum` upon initialization, we need to check if type is valid at this stage. for name how about `checkInput`?

packages/lisk-transactions/src/base_transaction.ts
```diff
@@ -0,0 +1,263 @@
...
+ * Unless otherwise agreed in a custom licensing agreement with the Lisk Foundation,
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ *
+ */
+// tslint:disable-next-line no-reference
+/// <reference path="../../../types/browserify-bignum/index.d.ts" />
```
SargeKhan: Is the reference path here correct? I thought #912 means that each package will reference it's own interface declaration file. Isn't that the case?

packages/lisk-transactions/src/base_transaction.ts
```diff
@@ -0,0 +1,263 @@
...
+			return {
+				validated: false,
+				errors: [
+					new TransactionError('Invalid transaction type.', transaction.id),
+				],
+			};
+		}
+
+		// Schema validation
+		const { valid, errors } = validateTransaction(transaction);
```
SargeKhan: I think common checks for all transactions in validateTransaction should be moved here, whereas checks specific for a particular transaction should be moved to their subclasses. I think having validateTransaction in utils defeats the purpose of creating separate classes.

packages/lisk-transactions/src/base_transaction.ts
```diff
@@ -0,0 +1,263 @@
...
+				errors: [
+					new TransactionError(
+						'Cannot validate transaction without signature.',
+						transaction.id,
+					),
+				],
+			};
+		}
+
+		const verified = verifyTransaction(transaction);
```
SargeKhan: I have the same opinion for `verifyTransaction` as for `validateTransaction`. Furthermore, we should start using core terminology for `verify`, `validate` functions. In core, when we are just performing schema checks, we call the process `objectNormalize`, `validate` means that we are performing signature validations, and `verify` means that we are performing database related checks.

packages/lisk-transactions/src/base_transaction.ts
SargeKhan: Yes, it needs to be discussed further before final decision.

packages/lisk-transactions/src/base_transaction.ts
```diff
@@ -0,0 +1,263 @@
...
+		const { verified: balanceVerified, errors } = checkBalance(
+			sender,
+			this.fee,
+		);
+
+		// Check secondPublicKey
+		const verified = sender.secondPublicKey
+			? verifyTransaction(this.toJSON(), sender.secondPublicKey)
+			: true;
+
```
SargeKhan: There are many checks in core /logic/transaction.js.. Don't we need them here as well?

packages/lisk-transactions/src/utils/get_transaction_bytes.ts
```diff
@@ -178,7 +177,7 @@ const transactionTypeAssetGetBytesMap: {
 	7: getAssetDataForTransferOutOfDappTransaction,
 };
 
-export const getAssetBytes = (transaction: BaseTransaction): Buffer =>
+export const getAssetBytes = (transaction: TransactionJSON): Buffer =>
```
SargeKhan: I think this function should be removed from here and moved to each transaction type.

SargeKhan: This whole file, actually.

packages/lisk-transactions/src/base_transaction.ts
mitsujutsu: this is the same path used in the [PR](https://github.com/LiskHQ/lisk-elements/commit/2714828036d50ad7f03fc0563b65f29308285ab9) for the issue.

packages/lisk-transactions/src/utils/get_transaction_bytes.ts
mitsujutsu: `getAsset` functions in this file will be used in each type class as they are implemented. as for removing them, I believe we want to keep `util` functions as-is for client-side for the time being? @shuse2

packages/lisk-transactions/src/base_transaction.ts
shuse2: Multi-sig transaction is registering multi signature transaction, and all the transactions after registration needs to be checked if sender is multi-sig account, it needs correct number of signatures.
it is similar to the case of second passphrase registration

packages/lisk-transactions/src/base_transaction.ts
shuse2: I feel like that is more of user's responsibility. what do you think @SargeKhan

packages/lisk-transactions/src/utils/get_transaction_bytes.ts
shuse2: yes, to not to make breaking change now, we should keep it for now, but not use it in the BaseTransaction etc.
Maybe add a comment to deprecate?

packages/lisk-transactions/src/base_transaction.ts
mitsujutsu: @shuse2 updated to not use `utils/verifyTransaction` and to return all errors at the end

packages/lisk-transactions/src/base_transaction.ts
```diff
@@ -0,0 +1,497 @@
...
+			type: this.type,
+			timestamp: this.timestamp,
+			senderPublicKey: this.senderPublicKey,
+			senderId: this.senderId,
+			recipientId: this.recipientId,
+			recipientPublicKey: this.recipientPublicKey,
+			fee: this.fee.toString(),
+			signatures: this.signatures,
+			asset: this.assetToJSON(this.asset),
+			receivedAt: this.receivedAt,
```
shuse2:
```
signature: this.signature ? this.signature : undefined,
signSignature: this.signSignature ? this.signSignature : undefined,
}
```
could work too?

packages/lisk-transactions/src/base_transaction.ts
```diff
@@ -0,0 +1,497 @@
...
+						bufferFromFirstEntriesReversed,
+					);
+					if (value !== transactionId) {
+						return [
+							...errorArray,
+							new TransactionError('Invalid transaction id', this.id, '.id'),
+						];
+					}
+				}
+				if (key === 'senderId') {
```
shuse2: this could be part of `checkShema`?
I think conclusion was
constructor => basic type check
checkSchema => detailed field check (formatting etc) and cross field check (senderID = senderPublicKey)
validation => signature check
verification => state check

packages/lisk-transactions/src/base_transaction.ts
```diff
@@ -0,0 +1,497 @@
...
+		const transaction = this.toJSON();
+		const transactionHash = cryptography.hash(this.getBasicBytes());
+
+		const validateErrors = Object.entries(transaction).reduce(
+			(
+				errorArray: ReadonlyArray<TransactionError>,
+				property: ReadonlyArray<string & ReadonlyArray<string>>,
+			): ReadonlyArray<TransactionError> => {
+				const [key, value] = property;
+				if (key === 'id') {
```
shuse2: This also could be part of check schema?

packages/lisk-transactions/src/base_transaction.ts
```diff
@@ -0,0 +1,497 @@
...
+
+		// Signature verifications
+
+		const transactionHash = cryptography.hash(this.getBasicBytes());
+
+		// Verify signSignature
+		const signSignatureError =
+			sender.secondPublicKey &&
+			!cryptography.verifyData(
+				transactionHash,
```
shuse2: To check second signature, I think first signature needs to be included in the hash. Please check the core

packages/lisk-transactions/src/utils/validate_multisignatures.ts
```diff
@@ -0,0 +1,54 @@
+import * as cryptography from '@liskhq/lisk-cryptography';
+import { Account, TransactionJSON } from '../transaction_types';
+
+export const validateMultisignatures = (
+	sender: Account,
+	transaction: TransactionJSON,
```
shuse2: maybe just get the signatures as input is fine?

packages/lisk-transactions/src/utils/validate_multisignatures.ts
```diff
@@ -0,0 +1,54 @@
...
+	transactionHash: Buffer,
+): ReadonlyArray<string> => {
+	const minimumValidations = sender.multimin || 0;
+	const transactionSignatures = transaction.signatures as ReadonlyArray<string>;
+	const senderPublicKeys = sender.multisignatures || [];
+	// tslint:disable-next-line no-let
+	let validatedSignaturesCount = 0;
+	// tslint:disable-next-line no-let
+	let checkedPublicKeys: ReadonlyArray<string> = [];
```
shuse2: having this with reduce together seems a bit weird.
If it makes easier to understand safely, not using reduce is also an option.

packages/lisk-transactions/src/base_transaction.ts
```diff
@@ -0,0 +1,497 @@
...
+			id: this.id,
+			status: Status.OK,
+			state: [updatedAccount],
+			errors: [],
+		};
+	}
+
+	public isExpired(): boolean {
+		// tslint:disable-next-line no-magic-numbers
+		const timeNow = Math.floor(Date.now() / 1000);
```
shuse2: is it important to divide by 1000?
also, @SargeKhan wanted to make it like
`isExpired(now: Date = Date.now()): boolean`?

packages/lisk-transactions/src/base_transaction.ts
```diff
@@ -0,0 +1,497 @@
...
+		// Balance verification
+		const { errors: balanceError } = checkBalance(sender, this.fee);
+
+		// Check transaction against account
+		const transactionErrors = Object.entries(sender).reduce(
+			(
+				errorArray: ReadonlyArray<TransactionError>,
+				property: ReadonlyArray<string | number>,
+			): ReadonlyArray<TransactionError> => {
+				const [key, value] = property;
```
shuse2: i'm not sure abt this usage of `reduce`.
what do u think @SargeKhan ?

packages/lisk-transactions/src/utils/validate_multisignatures.ts
```diff
@@ -0,0 +1,54 @@
...
+						return [signature, ...collectedSignatures];
+					},
+					[],
+				),
+			];
+		},
+		[],
+	);
+
+	return validatedSignaturesCount >= minimumValidations
```
SargeKhan: This function will never return an error? Even if we have some invalid signatures in the transaction.signatures array. Is that the expected behavior?

packages/lisk-transactions/src/utils/validation/schema.ts
```diff
@@ -121,6 +184,9 @@ export const signatureTransaction = {
 		source: { $ref: 'lisk/base-transaction' },
 		with: {
 			properties: {
+				amount: {
+					format: 'nonTransferAmount',
```
SargeKhan: why validateNonTransferAmount? The amount should be equal to zero. nonTransferAmount will return false for negative values.

packages/lisk-transactions/src/utils/validation/validation.ts
```diff
@@ -121,7 +121,7 @@ export const isNumberString = (str: string) => {
 	return /^[0-9]+$/g.test(str);
 };
 
-export const validateAmount = (data: string) =>
+export const validateNonTransferAmount = (data: string) =>
 	isNumberString(data) && !isGreaterThanZero(new BigNum(data));
```
SargeKhan: should be isEqualToZero.

packages/lisk-transactions/src/utils/validation/schema.ts
```diff
@@ -12,6 +12,69 @@
...
+			type: 'string',
+		},
+		signatures: {
+			type: 'array',
+		},
+		asset: {
+			type: 'object',
+		},
+		receivedAt: {
+			type: 'object',
```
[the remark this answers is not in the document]
mitsujutsu: it will be a javascript `Date` object. ajv options for `date`, `date-time` fail. can add format in baseTransaction schema to ensure instanceof Date.

packages/lisk-transactions/src/base_transaction.ts
SargeKhan: Yes, I think the function should accept the date parameter.

packages/lisk-transactions/src/base_transaction.ts
SargeKhan: Yes, I am not sure either. If we want to follow functional approach, we can define separate functions in the `verify` and call them all together at the end..  something like:
```
const verifySenderPublicKey = (sender) => {
if (key === 'publicKey' && value !== this.senderPublicKey) {
				  return new TransactionError(
						'Invalid sender publicKey',
						this.id,
						'.senderPublicKey',
					);
				}
}
const verifySecondPublicKey = (account) => {...}
...
...
[verifyPublicKey, verifySecondPublicKey,...].map(verify => verify(account)).filter(error => error instanceof TransactionError);
```
Another option would be to just push items in the errors array and that should be work as well.

packages/lisk-transactions/src/base_transaction.ts
mitsujutsu: would it be better not having `undefined` fields in result JSON?

packages/lisk-transactions/src/base_transaction.ts
mitsujutsu: between these two, I think I would prefer simply pushing. should we apply whatever is decided here everywhere? @shuse2 @SargeKhan?

shuse2: I think having small functions like (verifySenderPublicKey, verifySecondPublicKey) as @SargeKhan said would be nice for readability and testability.


How to handle those errors, I dont have strong opinion abt simple pushing or mapping, both sounds good to me =)

packages/lisk-transactions/src/utils/validate_multisignatures.ts
mitsujutsu: I can return errors directly from here. not sure if we should also return errors if there were more valid signatures than the minimum amount.

packages/lisk-transactions/src/utils/check_balance.ts
```diff
@@ -0,0 +1,43 @@
...
+	readonly errors: ReadonlyArray<TransactionError>;
+	readonly verified: boolean;
+} => {
+	const exceeded = new BigNum(account.balance).lt(amount);
+
+	return {
+		verified: !exceeded,
+		errors: exceeded
+			? [
+					new TransactionError(
```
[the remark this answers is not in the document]
mitsujutsu: in this function we don't have access to transaction id. for `dataPath` the amount could be fee or amount or both, therefore we don't know the `dataPath`.

packages/lisk-transactions/src/transaction_types.ts
```diff
@@ -12,26 +12,54 @@
  * Removal or modification of this copyright notice is prohibited.
  *
  */
-export interface BaseTransaction {
+export enum Status {
+	OK = 1,
+	FAIL,
+}
+export interface Account {
```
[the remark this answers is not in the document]
mitsujutsu: i think it's related to this `transactions library` so it can be here. maybe later if there are more types we can create a separate type file?

packages/lisk-transactions/src/base_transaction.ts
mitsujutsu: @shuse2 `/  1000` follows the logic in core `transaction_pool:isExpired`

packages/lisk-transactions/src/transactions/base.ts
```diff
@@ -0,0 +1,365 @@
...
+
+	public getRequiredAttributes(): object {
+		return {
+			ACCOUNTS: [cryptography.getAddressFromPublicKey(this.senderPublicKey)],
+		};
+	}
+
+	public verify(sender: Account): TransactionResponse {
+		// tslint:disable-next-line no-let
+		let errors: TransactionError[] = [];
```
shuse2: `const` should be ok

packages/lisk-transactions/src/transactions/helpers/get_bytes.ts
```diff
@@ -0,0 +1,72 @@
...
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ *
+ */
+import * as cryptography from '@liskhq/lisk-cryptography';
+import BigNum from 'browserify-bignum';
+import { BYTESIZES } from '../../constants';
+import { TransactionJSON } from '../../transaction_types';
+
+export const getBytes = (
```
shuse2: im not sure if this function make sense.
Asset bytes always need to be included in the base bytes.
only in case of Transfer transaction, it's could be empty

packages/lisk-transactions/src/transactions/helpers/get_id.ts
```diff
@@ -0,0 +1,30 @@
...
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ *
+ */
+import * as cryptography from '@liskhq/lisk-cryptography';
+import { TransactionJSON } from '../../transaction_types';
+import { getTransactionBytes } from '../../utils';
+
+export const getId = (transaction: TransactionJSON): string => {
```
shuse2: we have the same function in `utils/get_transaction_id.ts`

packages/lisk-transactions/src/transactions/helpers/index.ts
```diff
@@ -0,0 +1,29 @@
+/*
```
shuse2: whats the difference between utils and helpers?

packages/lisk-transactions/src/transactions/helpers/verify_multisignatures.ts
```diff
@@ -0,0 +1,90 @@
...
+import { verifySignature } from './verify_signature';
+
+interface VerifyReturn {
+	readonly verified: boolean;
+	readonly errors: ReadonlyArray<TransactionError>;
+}
+
+export const verifyMultisignatures = (
+	memberPublicKeys: ReadonlyArray<string> = [],
+	minimumValidations: number = 0,
```
shuse2: assigning minimum would be dangerous here. if we put undefined, i will always valid

packages/lisk-transactions/src/transactions/helpers/verify_multisignatures.ts
```diff
@@ -0,0 +1,90 @@
...
+				new TransactionError(
+					'Missing signatures',
+					transaction.id,
+					'.signatures',
+				),
+			],
+		};
+	}
+
+	// tslint:disable-next-line no-unnecessary-type-assertion
```
shuse2: why do we need this?

packages/lisk-transactions/src/transactions/helpers/verify_signature.ts
```diff
@@ -0,0 +1,61 @@
...
+	transaction: TransactionJSON,
+	isSecondSignature: boolean = false,
+): VerifyReturn => {
+	const {
+		signature: removedSignature,
+		signSignature,
+		...strippedTransaction
+	} = transaction;
+
+	// If transaction includes asset data, include those bytes
```
shuse2: it should always include asset bytes.

shuse2: also similar function exists in `sign_and_verify.ts`

packages/lisk-transactions/src/transactions/helpers/verify_signature.ts
```diff
@@ -0,0 +1,61 @@
...
+	const transactionBytes =
+		transaction.asset && Object.keys(transaction.asset).length
+			? getTransactionBytes(strippedTransaction as TransactionJSON)
+			: getBytes(transaction, !isSecondSignature);
```
shuse2: this transction could include signSignature?

packages/lisk-transactions/src/transactions/helpers/index.ts
mitsujutsu: The existing util functions such as `sign_and_verify/verifyTransaction` and `utils/get_transaction_id` uses `getTransactionBytes` which uses getAssetBytes.  As per previous discussion this may negate the point of classes by doing this. If we are to stick to handling assets in each class, we need new functions to fit this approach.  How about `helpers/verifySignature(key, signature, txBytes, isSecondSignature)` where `txBytes` will come from each type classes' `getBytes` method (and remove the `helpers/getBytes` and put back previous `getBasicBytes` method which will be used in `getBytes` method).

packages/lisk-transactions/src/transactions/helpers/get_id.ts
mitsujutsu: see below answer

packages/lisk-transactions/src/transactions/helpers/verify_signature.ts
mitsujutsu: see above answer re: `sign_and_verify.ts`

packages/lisk-transactions/src/transactions/helpers/verify_multisignatures.ts
mitsujutsu: `transaction.signatures.forEach` below throws tslint `possible undefined`, so casting in assignment of `transactionSignatures`. however this is turn shows `no-unnecessary-type-assertion` lint error so disabling that here.

shuse2: In this case, definition of signatures should not be optional or should it be checked if signatures is undefined?

packages/lisk-transactions/src/transactions/helpers/index.ts
shuse2: helpers and utils are essentially the same imo.
We can name function differently, and put deprecated comments in one of them if it will not be used in the future.
as per similar functions, i think we can put it in the same file in the utils folder.
maybe in the base class we can have
```
getBytes => (return whole bytes including singature/signSignature) uses abstract getAssetBytes
getBasicBytes => (return bytes without signature) uses abstract getAssetBytes
abstract getAssetBytes as it is now
```
I would prefer not to use `skip` flag.
also `verifySignature(publicKey, signature, txBytes)` looks good. and depends on where it's called `txBytes` changes

packages/lisk-transactions/src/transactions/helpers/verify_multisignatures.ts
shuse2: mmm strange ='(

packages/lisk-transactions/src/transactions/base.ts
```diff
@@ -0,0 +1,402 @@
...
+				}
+			}
+		}
+
+		// Verify multisignatures
+		if (
+			Array.isArray(sender.multisignatures) &&
+			sender.multisignatures.length > 0 &&
+			sender.multimin &&
+			this.signatures
```
SargeKhan: deafult value of signatures is `[]`, so I think we don't need the check here.

packages/lisk-transactions/src/utils/sign_and_verify.ts
```diff
@@ -36,8 +45,109 @@ export const multiSignTransaction = (
...
+			errors: [new TransactionError('Missing signatures', id, '.signatures')],
+		};
+	}
+
+	const verifiedSignatures: string[] = [];
+	const checkedPublicKeys: string[] = [];
+
+	memberPublicKeys.forEach(publicKey => {
+		if (checkedPublicKeys.includes(publicKey)) {
+			return;
```
SargeKhan: this condition will never be true unless memberPublicKeys are not unique

packages/lisk-transactions/src/utils/sign_and_verify.ts
```diff
@@ -36,8 +45,109 @@ export const multiSignTransaction = (
...
+			);
+			if (signatureVerified) {
+				checkedPublicKeys.push(publicKey);
+				verifiedSignatures.push(signature);
+			}
+		});
+	});
+
+	const unverifiedSignatures = signatures.filter(
+		e => !verifiedSignatures.find(a => e === a),
```
SargeKhan: more descriptive variable names?

packages/lisk-transactions/src/utils/sign_and_verify.ts
```diff
@@ -36,8 +45,109 @@ export const multiSignTransaction = (
...
+	);
+
+	return {
+		verified: verifiedSignatures.length >= minimumValidations,
```
SargeKhan: Could be that verified is true while this function still returns errors.

packages/lisk-transactions/src/utils/sign_and_verify.ts
mitsujutsu: it can be true as per [issue #2540](https://github.com/LiskHQ/lisk/issues/2540). in core logic this check is also done in the same way.

packages/lisk-transactions/src/utils/sign_and_verify.ts
mitsujutsu: in `base` it's checking the `errors` content as opposed to `verified` boolean, so it would fail with errors anyway. in some cases it could be helpful to know it passed multimin validations even if there were invalid signatures. what do you think?

packages/lisk-transactions/test/transactions/base.ts
```diff
@@ -0,0 +1,746 @@
...
+					fee: '0000',
+					recipientId: '',
+					senderPublicKey: '11111111',
+					senderId: '11111111',
+					timestamp: 79289378,
+					asset: {},
+					signature: '1111111111',
+					id: '1',
+				};
+				baseTransaction = new TestTransaction(invalidTransaction as any);
```
SargeKhan: We have nested `beforeEach` functions where we are creating multiple new Transactions in each level. This will make tests a lot harder to understand and reason about. So, maybe it's better to not have so many nested levels, or only perform actions in the beforeEach blocks which are shared in all the tests.

packages/lisk-transactions/test/transactions/base.ts
```diff
@@ -0,0 +1,746 @@
...
+			});
+
+			describe('when given unmatching senderId and senderPublicKey', () => {
+				beforeEach(() => {
+					const invalidIdTransaction = {
+						...defaultTransaction,
+						senderId: defaultTransaction.senderId.replace('1', '0'),
+					};
+
+					baseTransaction = new TestTransaction(invalidIdTransaction as any);
```
SargeKhan: Same goes here as well. I think for this test, the `new Transaction` will be called 5 times.

packages/lisk-transactions/test/utils/sign_and_verify.ts
```diff
@@ -156,15 +162,235 @@ describe('signAndVerify module', () => {
...
+				signature:
+					'3357658f70b9bece24bd42769b984b3e7b9be0b2982f82e6eef7ffbd841598d5868acd45f8b1e2f8ab5ccc8c47a245fe9d8e3dc32fc311a13cc95cc851337e01',
+				signSignature:
+					'11f77b8596df14400f5dd5cf9ef9bd2a20f66a48863455a163cabc0c220ea235d8b98dec684bd86f62b312615e7f64b23d7b8699775e7c15dad0aef0abd4f503',
+				id: '11638517642515821734',
+				receivedAt: new Date(),
+			};
+
+			beforeEach(() => {
+				cryptoVerifyDataStub.restore();
```
SargeKhan: Why do we need to restoring a stub in `beforeEach`? Since each stub created via sandbox is restored in the afterEach block anyway?

packages/lisk-transactions/test/utils/sign_and_verify.ts
```diff
@@ -156,15 +162,235 @@ describe('signAndVerify module', () => {
...
+					'36d5c7da5f54007e22609105570fad04597f4f2b00d46baba603c213eaed8de55e9f3e5d0f39789dbc396330b2d9d4da46b7d67187075e86220bc0341c3f7802',
+				],
+				asset: {
+					data: 'the real test',
+				},
+				receivedAt: new Date(),
+			};
+
+			beforeEach(() => {
+				cryptoVerifyDataStub.restore();
```
SargeKhan: Same here.

packages/lisk-transactions/test/utils/sign_and_verify.ts
mitsujutsu: It was not restoring and returning true each time :/ I will re-organize in more comprehensive testing PR

packages/lisk-transactions/test/transactions/base.ts
mitsujutsu: I can remove some new Transaction calls (the ones with `defaultTransaction`) but not all of them as it is necessary to instantiate with specific transactions in order to test methods.

PR #866, opened 2018-11-21
Add valid number of words to validate passphrase

### LINKED ISSUE #862: Lisk Element should allow BIP39 standard passphrase with up to 24 words
### Expected behavior
 - Since the BIP39 standard allows to generate a passphrase with up to 24 words and since we use mnemonic which implements this standard, we should not allow users to use passphrases longer than 12 words. This includes generation and validation. 
One option would be to let people pass the number of words they want to get from the passphrase generator module.

### Actual behavior
Currently, the passphrase is limited to 12 words. This seems to not be a technical restriction, but more of a business decision.

### Which version(s) does this effect? (Environment, OS, etc...)
 - Version 1.1.2 and before

shuse2:
### What was the problem?
Validating passphrase was limited to exactly 12 words.

### How did I fix it?
Add another input field to adjust expected words count

### How to test it?
`npm t`
or
```
passphrase.getPassphraseValidationErrors('post dumb recycle buddy round normal scrap better people corn crystal again never shrimp kidney') // this should be error
passphrase.getPassphraseValidationErrors('post dumb recycle buddy round normal scrap better people corn crystal again never shrimp kidney', undefined, 15)
```

### Review checklist

* The PR resolves #862 
* Linting passes
* Tests pass

### DIFF
--- a/packages/lisk-passphrase/src/validation.ts
+++ b/packages/lisk-passphrase/src/validation.ts
@@ -117,9 +117,9 @@ export const locateConsecutiveWhitespaces = (
 export const getPassphraseValidationErrors = (
 	passphrase: string,
 	wordlists?: ReadonlyArray<string>,
+	expectedWords: number = 12,
 ): ReadonlyArray<PassphraseError> => {
-	const expectedWords = 12;
-	const expectedWhitespaces = 11;
+	const expectedWhitespaces = expectedWords - 1;
 	const expectedUppercaseCharacterCount = 0;
 	const wordsInPassphrase = countPassphraseWords(passphrase);
 	const whiteSpacesInPassphrase = countPassphraseWhitespaces(passphrase);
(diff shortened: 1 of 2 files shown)


packages/lisk-passphrase/src/validation.ts
```diff
@@ -117,9 +117,9 @@ export const locateConsecutiveWhitespaces = (
 export const getPassphraseValidationErrors = (
 	passphrase: string,
 	wordlists?: ReadonlyArray<string>,
+	expectedWords: number = 12,
```
mitsujutsu: should this be optional: `expectedWords?: number = 12`

shuse2: once we have default value, it doesn't allow to use optional. Instead it will automatically read it as optional

PR #869, opened 2018-11-22
:bug: Add babel-runtime to the dependencies

### LINKED ISSUE #868: Failed to require lisk-elements package
### Expected behavior
After installing `lisk-elements` I should be able to require the package.
### Actual behavior
The package is failed to require with following error message
```
Error: Cannot find module 'babel-runtime/helpers/toConsumableArray'
Should we have babel-runtime/helpers/toConsumableArray?
We should have every package on npm. Sometimes this error is caused by a typo in your require path, but if you think we are actually missing this package, please confirm below and we'll do our best to fix it as soon as possible!
```
### Steps to reproduce
1) `npm i lisk-elements`
2) `const elements = require('lisk-elements')`
`>1.0.0`

shuse2:
### What was the problem?
When not used in the babel environment in node-js, it was filing to use `babel-runtime` feature

### How did I fix it?
Add `babel-runtime` to dependencies for each packages.

### Review checklist

* The PR resolves #868 
* Linting passes
* Tests pass

### DIFF
--- a/packages/lisk-api-client/package.json
+++ b/packages/lisk-api-client/package.json
@@ -58,7 +58,8 @@
 	},
 	"dependencies": {
 		"@liskhq/lisk-constants": "1.1.1",
-		"axios": "0.18.0"
+		"axios": "0.18.0",
+		"babel-runtime": "6.26.0"
 	},
 	"devDependencies": {
 		"babel-cli": "6.26.0",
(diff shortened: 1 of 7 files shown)


packages/lisk-api-client/package.json
```diff
@@ -58,7 +58,8 @@
 	},
 	"dependencies": {
 		"@liskhq/lisk-constants": "1.1.1",
-		"axios": "0.18.0"
+		"axios": "0.18.0",
+		"babel-runtime": "6.26.0"
```
ManuGowda: @shuse2 I am not that familiar with `lerna`, however, I read somewhere that if the dependency is common across all packages we can add it root, so why are we not considering that here?

shuse2: It is only for `devDependencies` which doesn't use the ./bin folder, so it doesn't apply here

PR #898, opened 2018-11-22
Add PeerPool and prepare integration tests

jondubois:
### Description

Add PeerPool and set up integration tests.

### DIFF
--- a/packages/lisk-p2p/package.json
+++ b/packages/lisk-p2p/package.json
@@ -42,20 +42,20 @@
 		"build": "npm run build:node",
 		"build:check": "node -e \"require('./dist-node')\"",
 		"prepublishOnly": "npm run lint && npm test && npm run build && npm run build:check"
-  },
-  "dependencies": {
-    "socketcluster-client": "14.2.1",
-    "socketcluster-server": "14.3.1"
-  },
+	},
+	"dependencies": {
+		"socketcluster-client": "14.2.1",
+		"socketcluster-server": "14.3.1"
+	},
 	"devDependencies": {
-    "@types/chai": "4.1.5",
-    "@types/component-emitter": "1.2.7",
+		"@types/chai": "4.1.5",
+		"@types/component-emitter": "1.2.7",
 		"@types/expect": "1.20.3",
-    "@types/jquery": "3.3.9",
-    "@types/jsonwebtoken": "8.3.0",
+		"@types/jquery": "3.3.9",
+		"@types/jsonwebtoken": "8.3.0",
 		"@types/mocha": "5.2.5",
-    "@types/node": "10.10.1",
-    "@types/ws": "6.0.1",
+		"@types/node": "10.10.1",
+		"@types/ws": "6.0.1",
 		"browserify": "16.2.2",
 		"chai": "4.1.2",
 		"mocha": "5.2.0",
--- a/packages/lisk-p2p/package.json
+++ b/packages/lisk-p2p/package.json
@@ -42,13 +42,20 @@
 		"build": "npm run build:node",
 		"build:check": "node -e \"require('./dist-node')\"",
 		"prepublishOnly": "npm run lint && npm test && npm run build && npm run build:check"
-	},
+  },
+  "dependencies": {
+    "socketcluster-client": "14.2.1",
+    "socketcluster-server": "14.3.1"
+  },
 	"devDependencies": {
-		"@types/chai": "4.1.5",
+    "@types/chai": "4.1.5",
+    "@types/component-emitter": "1.2.7",
 		"@types/expect": "1.20.3",
-		"@types/jquery": "3.3.9",
+    "@types/jquery": "3.3.9",
+    "@types/jsonwebtoken": "8.3.0",
 		"@types/mocha": "5.2.5",
-		"@types/node": "10.10.1",
+    "@types/node": "10.10.1",
+    "@types/ws": "6.0.1",
 		"browserify": "16.2.2",
 		"chai": "4.1.2",
 		"mocha": "5.2.0",
--- a/packages/lisk-p2p/src/errors/p2p_error.ts
+++ b/packages/lisk-p2p/src/errors/p2p_error.ts
@@ -1,12 +1,6 @@
 export class P2PError extends Error {
-	private readonly _name: string;
-
-	public constructor(message: string, errorName?: string) {
+	public constructor(message: string) {
 		super(message);
-		this._name = errorName ? errorName : 'P2PError';
-	}
-
-	public get name(): string {
-		return this._name;
+		this.name = 'P2PError';
 	}
 }
--- a/packages/lisk-p2p/src/errors/p2p_error.ts
+++ b/packages/lisk-p2p/src/errors/p2p_error.ts
@@ -1,0 +1,12 @@
+export class P2PError extends Error {
+	private readonly _name: string;
+
+	public constructor(message: string, errorName?: string) {
+		super(message);
+		this._name = errorName ? errorName : 'P2PError';
+	}
+
+	public get name(): string {
+		return this._name;
+	}
+}
--- a/packages/lisk-p2p/src/errors/peer_transport_error.ts
+++ b/packages/lisk-p2p/src/errors/peer_transport_error.ts
@@ -1,14 +1,9 @@
-import { P2PError } from './p2p_error';
-
-export class PeerTransportError extends P2PError {
-	private readonly _peerId: string;
+export class PeerTransportError extends Error {
+	public peerId: string;
 
 	public constructor(message: string, peerId: string) {
-		super(message, 'PeerTransportError');
-		this._peerId = peerId;
-	}
-
-	public get peerId(): string {
-		return this._peerId;
+		super(message);
+		this.name = 'PeerTransportError';
+		this.peerId = peerId;
 	}
 }
--- a/packages/lisk-p2p/src/errors/peer_transport_error.ts
+++ b/packages/lisk-p2p/src/errors/peer_transport_error.ts
@@ -1,0 +1,14 @@
+import { P2PError } from './p2p_error';
+
+export class PeerTransportError extends P2PError {
+	private readonly _peerId: string;
+
+	public constructor(message: string, peerId: string) {
+		super(message, 'PeerTransportError');
+		this._peerId = peerId;
+	}
+
+	public get peerId(): string {
+		return this._peerId;
+	}
+}
--- a/packages/lisk-p2p/src/p2p_types.ts
+++ b/packages/lisk-p2p/src/p2p_types.ts
@@ -25,8 +25,10 @@ export interface IP2PNodeStatus {}
 export interface P2PConfig {
 	readonly blacklistedPeers: ReadonlyArray<string>;
 	readonly connectTimeout: number;
+	readonly ipAddress?: string;
 	readonly seedPeers: ReadonlyArray<string>;
-	readonly wsEngine: string;
+	readonly wsEngine?: string;
+	readonly wsPort: number;
 }
 
 export interface IP2PPenalty {}
--- a/packages/lisk-p2p/src/p2p_types.ts
+++ b/packages/lisk-p2p/src/p2p_types.ts
@@ -13,6 +13,7 @@
  *
  */
 /* tslint:disable:interface-name no-empty-interface */
+
 export interface IP2PMessagePacket {}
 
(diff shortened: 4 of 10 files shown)


packages/lisk-p2p/package.json
```diff
@@ -42,13 +42,20 @@
 		"build": "npm run build:node",
 		"build:check": "node -e \"require('./dist-node')\"",
 		"prepublishOnly": "npm run lint && npm test && npm run build && npm run build:check"
-	},
+  },
+  "dependencies": {
```
shuse2: formatting...?

packages/lisk-p2p/src/p2p_types.ts
```diff
@@ -21,6 +22,46 @@ export interface IP2PResponsePacket {}
...
+
+export interface P2PConfigPeers {
+	readonly access: P2PConfigPeersAccess;
+	readonly enabled: boolean;
+	readonly list: ReadonlyArray<string>;
+	readonly options: P2PConfigPeersOptions;
+}
+
+export interface P2PConfig {
+	readonly logger?: Logger;
```
shuse2: do we need to pass logger?
I'm not sure if library should support expecting logger, socket-cluster have possibility to show logs but not inserting logger for example
we could use https://www.npmjs.com/package/debug?

packages/lisk-p2p/src/p2p_types.ts
```diff
@@ -21,6 +22,46 @@ export interface IP2PResponsePacket {}
...
+export interface P2PConfigPeersOptions {
+	readonly broadhashConsensusCalculationInterval: number;
+	readonly timeout: number;
+	readonly wsEngine: string;
+}
+
+export interface P2PConfigPeers {
+	readonly access: P2PConfigPeersAccess;
+	readonly enabled: boolean;
+	readonly list: ReadonlyArray<string>;
```
shuse2: maybe better name? like `seeds`?

packages/lisk-p2p/src/p2p_types.ts
```diff
@@ -21,6 +22,46 @@ export interface IP2PResponsePacket {}
...
+
+export interface P2PConfigPeersOptions {
+	readonly broadhashConsensusCalculationInterval: number;
+	readonly timeout: number;
+	readonly wsEngine: string;
+}
+
+export interface P2PConfigPeers {
+	readonly access: P2PConfigPeersAccess;
+	readonly enabled: boolean;
```
shuse2: do we need `enable` on config here?
I feel like it should be handled on application

packages/lisk-p2p/src/p2p_types.ts
```diff
@@ -21,6 +22,46 @@ export interface IP2PResponsePacket {}
...
+	readonly list: ReadonlyArray<string>;
+	readonly options: P2PConfigPeersOptions;
+}
+
+export interface P2PConfig {
+	readonly logger?: Logger;
+	readonly peers: P2PConfigPeers;
```
shuse2: `peers` are confusing name for config. maybe options?
Also, important setting should be on the root level

packages/lisk-p2p/src/peer.ts
```diff
@@ -13,45 +13,65 @@
...
-	readonly state?: number;
 	readonly version?: string;
 	readonly wsPort: number;
 }
 
 export class Peer {
 	private height: number;
-	private ip: string;
-	private wsPort: number;
+	private readonly id: string;
```
shuse2: let's discuss how to name private members (ie: _id vs id)

packages/lisk-p2p/src/peer.ts
```diff
@@ -13,45 +13,65 @@
  *
  */
 /* tslint:disable:interface-name */
+import { Logger, PeerState } from './p2p_types';
+
+// TODO: Use to create outbound socket connection inside peer object.
+// TODO: const socketClusterClient = require('socketcluster-client');
 
 export interface IPeerConfig {
 	readonly clock?: Date;
```
shuse2: do we use `clock`...?

packages/lisk-p2p/src/peer.ts
```diff
@@ -13,45 +13,65 @@
...
 	readonly height: number;
 	readonly httpPort?: number;
```
shuse2: `httpPort` probably not needed?

packages/lisk-p2p/test/integration/p2p.ts
```diff
@@ -0,0 +1,47 @@
+import { P2P } from '../../src/index';
+
+const NETWORK_PEER_COUNT: Number = 10;
```
shuse2: this should be defined inside of `describe`

packages/lisk-p2p/test/integration/p2p.ts
```diff
@@ -0,0 +1,47 @@
...
+
+describe('Integration tests for BlockchainP2P', () => {
```
shuse2: need to update the naming

packages/lisk-p2p/src/p2p_types.ts
jondubois: We need to account for multiple processes. This means that we need some way to aggregate the logs from different libraries and modules. I don't know if `debug` support this. It depends on the controller. Also there is https://github.com/trentm/node-bunyan and https://github.com/winstonjs/winston to consider.

packages/lisk-p2p/src/peer.ts
jondubois: We may want to allow nodes to report their HTTP ports to each other but I guess this should be part of the nodeStatus object (along with the height).

The nodeStatus is meant to be the alternative to peer headers so the API module could tell us what the httpPort should be to include as part of the nodeStatus; so we can report it to other peers that way.

packages/lisk-p2p/src/peer.ts
ishantiw: I added it because it was present in the core, but we can always remove or ignore it as its an optional field

packages/lisk-p2p/src/peer.ts
ishantiw: should we discuss these fields and how it's going to be used?

packages/lisk-p2p/src/peer_pool.ts
```diff
@@ -0,0 +1,108 @@
...
+	readonly blacklistedPeers?: ReadonlyArray<string>;
+	readonly logger: Logger;
+	readonly seedPeers: ReadonlyArray<string>;
+}
+
+export class PeerPool extends EventEmitter {
+	public httpServer: Server;
+	public logger: Logger;
+	public newPeers: Map<string, Peer>;
+	public scServer: any;
```
ishantiw: `scServer` doesn't have the type from the library?

packages/lisk-p2p/test/integration/p2p.ts
ishantiw: shouldn't it be lowercase `number`

packages/lisk-p2p/src/peer_pool.ts
jondubois: No, the type definitions from the community were not usable.

jondubois:
@shuse2 Please review again.

packages/lisk-p2p/src/errors/p2p_error.ts
```diff
@@ -0,0 +1,12 @@
+export class P2PError extends Error {
+	private readonly _name: string;
```
shuse2: `name` can be public and exactly same as the class name?

packages/lisk-p2p/src/errors/peer_transport_error.ts
```diff
@@ -0,0 +1,14 @@
+import { P2PError } from './p2p_error';
+
+export class PeerTransportError extends P2PError {
```
shuse2: maybe we can create `errors.ts` and put all the errors there?

packages/lisk-p2p/src/p2p_types.ts
shuse2: We have decided to use even emitting for the non-functional actions, and if needed, we will use `debug` internally, but not exposing the logs to the users

packages/lisk-p2p/src/peer.ts
shuse2: Since this is out of scope of this PR, lets fix it in another one

packages/lisk-p2p/src/peer.ts
shuse2: this should reflect decision to use `_id`?

packages/lisk-p2p/src/peer_pool.ts
```diff
@@ -0,0 +1,112 @@
...
+				`Received inbound connection from peer ${peerId} which is already in our triedPeers map.`,
+				peerId,
+			);
+			this.emit('inboundPeerFail', error);
+		} else if (this._newPeers.has(peerId)) {
+			const error: PeerTransportError = new PeerTransportError(
+				`Received inbound connection from peer ${peerId} which is already in our newPeers map.`,
+				peerId,
+			);
+			this.emit('inboundPeerFail', error);
```
shuse2: maybe create event name constants and use that?
also, is it usually camel case or capital with underscore?

packages/lisk-p2p/src/errors/peer_transport_error.ts
shuse2: ah ok then let's keep it that way then

packages/lisk-p2p/src/errors/p2p_error.ts
jondubois: Yes, it works and the getter overrides the default name property. It's a tricky case because the `name` property is public on the `Error` class but at the same time we want to discourage public properties.

That approach solves that problem and prevents users from overwriting the name property but I guess we can decide to allow public `name` property in our case for the sake of consistency with JavaScript.

packages/lisk-p2p/src/peer_pool.ts
jondubois: We need to decide what conventions we should use for events. Should we use a pattern like: `peerPool.on(peerPool.EVENT_INBOUND_PEER_FAIL, handler)` or just `peerPool.on(peerPool.INBOUND_PEER_FAIL, handler)`?

packages/lisk-p2p/src/peer_pool.ts
```diff
@@ -0,0 +1,120 @@
...
+
+import { EventEmitter } from 'events';
+import http, { Server } from 'http';
+import querystring from 'querystring';
+import { Peer } from './peer';
+import { PeerTransportError } from './errors';
+
+import socketClusterServer from 'socketcluster-server';
+
+const EVENT_INBOUND_PEER_FAIL = 'inboundPeerFail';
```
shuse2: you could just say
`export const EVENT_INBOUND_PEER_FAIL = 'inboundPeerFail';`

packages/lisk-p2p/src/peer.ts
```diff
@@ -13,45 +13,63 @@
...
+		this._inboundSocket = peerConfig.inboundSocket;
+		this._height = peerConfig.height;
 	}
 
 	public getHeight(): number {
-		return this.height;
+		return this._height;
+	}
+
+	public getId(): string {
```
shuse2: i'm not sure if this is scope of  Ishan's PR or this PR, but it should be 
`public get id(): string {`
@ishantiw

packages/lisk-p2p/src/peer.ts
```diff
@@ -13,45 +13,63 @@
...
+		return this._id;
 	}
+
 	public getIp(): string {
```
shuse2: same as above

packages/lisk-p2p/src/peer.ts
```diff
@@ -13,45 +13,63 @@
...
-		this.ip = peerConfig.ip;
-		this.wsPort = peerConfig.wsPort;
+		this._id = peerConfig.id;
+		this._ipAddress = peerConfig.ipAddress;
+		this._wsPort = peerConfig.wsPort;
+		this._inboundSocket = peerConfig.inboundSocket;
+		this._height = peerConfig.height;
 	}
 
 	public getHeight(): number {
```
shuse2: same as below

packages/lisk-p2p/src/peer.ts
ishantiw: I think we can ignore it here as it will be covered by my PR

PR #909, opened 2018-11-23
Adding p2p peer selection algorithm - Closes#891

### LINKED ISSUE #891: Add peer selection algorithm for lisk-p2p
### Description
Add basic peer selection mechanism for lisk-p2p, which will be used to select peers to communicate to. This algorithm is aimed to return n number of good peers from the peer list based on good peer algorithm.

- Takes peerlist, options and number of peers to be returned,  as an arguments.
- Returns an object (options and good peerlist) where `options` is an object with parameters needed for calculation of good peers, in our case its block height.
- Fix getter/setter for Peer class
- Add missing tests

ishantiw:
### Description

Add basic peer selection mechanism for lisk-p2p, which will be used to select peers to communicate to. This algorithm is aimed to return n number of good peers from the peer list based on good peer algorithm.

- Takes peerlist, options and number of peers to be returned, as an arguments.
- Returns an object (options and good peerlist) where options is an object with parameters needed for calculation of good peers, in our case its block height.

### Review checklist

* The PR resolves #891
* Linting passes
* Tests pass


packages/lisk-p2p/src/p2p.ts
```diff
@@ -12,48 +12,149 @@
...
+			throw new NotEnoughPeersError(
+				`Requested no. of peers: '${numOfPeers}' is more than the available no. of good peers: '${
+					processedPeers.length
+				}'`,
+			);
+		}
+		if (numOfPeers === processedPeers.length) {
+			return { options, peers: processedPeers };
+		}
+		if (numOfPeers === 1) {
```
shuse2: isn't this case same as below?

packages/lisk-p2p/test/unit/error.ts
```diff
@@ -0,0 +1,60 @@
...
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ *
+ */
+import { expect } from 'chai';
+import { NotEnoughPeersError } from '../../src';
+
+describe('#not enough peers errors module', () => {
```
shuse2: Top title should be the module name.
Something like `Error Module` please check other tests?

packages/lisk-p2p/test/unit/error.ts
```diff
@@ -0,0 +1,60 @@
...
+	let notEnoughPeersError: NotEnoughPeersError;
+	const defaultMessage =
+		'Requested number of peers is greater than available good peers';
+
+	beforeEach(() => {
```
shuse2: for newer ones, maybe it's good to start to use `async` rather than return ing promise

packages/lisk-p2p/test/unit/error.ts
shuse2: Probably you need one more outer `describe`

packages/lisk-p2p/test/unit/p2p.ts
```diff
@@ -1,10 +1,70 @@
...
+	};
+	const peer1 = new Peer(peerOption1);
+	const peer2 = new Peer(peerOption2);
+	const peer3 = new Peer(peerOption3);
+	const peer4 = new Peer(peerOption4);
+
+	return [peer1, peer2, peer3, peer4];
+};
 
 describe('#Good peers test', () => {
```
shuse2: same as above, it's good to have top level `describe`

packages/lisk-p2p/test/unit/peer.ts
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ *
+ */
+/* tslint:disable:no-console */
+import { expect } from 'chai';
+
+import { PeerConfig, Peer } from '../../src';
+
+describe('#Peer test', () => {
```
shuse2: `#` referes to method until now, so I think it's good to say `Peer module` or `Peer entity` please check other tests

packages/lisk-p2p/test/unit/peer.ts
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * Unless otherwise agreed in a custom licensing agreement with the Lisk Foundation,
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ *
+ */
+/* tslint:disable:no-console */
```
shuse2: I think you don't need this anymore

packages/lisk-p2p/src/p2p.ts
```diff
@@ -12,48 +12,149 @@
...
-	IP2PResponsePacket,
+	NetworkStatus,
+	P2PMessagePacket,
+	P2PNodeStatus,
+	P2PRequestPacket,
+	P2PResponsePacket,
 } from './p2p_types';
 import { Peer } from './peer';
-/* tslint:disable:interface-name readonly-keyword no-empty-interface no-let no-unused-expression */
+/* tslint:disable: no-unused-expression */
```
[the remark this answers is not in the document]
ishantiw: Still, methods of P2P are not implemented and so I kept it there, will remove once it's done.

packages/lisk-p2p/src/p2p.ts
```diff
@@ -12,48 +12,149 @@
...
 
-export interface IPeerReturnType {
-	readonly options: IPeerOptions;
+export interface PeerReturnType {
+	readonly options: PeerOptions;
```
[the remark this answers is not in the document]
ishantiw: In core, we are returning `{blockHeight, peers}`. since we are using options for holding block-height, so we are returning options too.

packages/lisk-p2p/src/errors.ts
```diff
@@ -0,0 +1,27 @@
...
+import { VError } from 'verror';
+
+export class NotEnoughPeersError extends VError {
+	private readonly code = 'NOT_ENOUGH_PEERS';
+
+	public constructor(message?: string) {
+		super(message || '');
+		this.name = 'Not Enough Peers Error';
+	}
+	public get Code(): string {
```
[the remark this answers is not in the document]
ishantiw: Actually, I made a getter for code property because I wanted to make it read-only and private. Also, by default, this error will not have `code` property so I thought it will be a good idea to have `code`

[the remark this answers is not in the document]
ishantiw: I think I will remove this method and make the code property public in order to avoid confusion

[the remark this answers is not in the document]
ishantiw: Since it's an error only used in this method, I'm simply throwing `VError` with its custom name field which will act similar to what we are doing now

packages/lisk-p2p/src/helpers.ts
```diff
@@ -0,0 +1,128 @@
...
+import { Peer } from './peer';
+
+export interface PeerReturnType {
+	readonly options: PeerOptions;
+	readonly peers: ReadonlyArray<Peer>;
+}
+export interface PeerOptions {
+	readonly [key: string]: string | number;
+}
+/* tslint:disable: readonly-keyword*/
```
shuse2: `// tslint:disable-next-line` should work better

packages/lisk-p2p/src/peer.ts
```diff
@@ -26,32 +25,29 @@ export interface IPeerConfig {
 }
 
 export class Peer {
-	private height: number;
-	private ip: string;
-	private wsPort: number;
+	private _height: number;
+	private readonly _ip: string;
+	private _wsPort: number;
```
shuse2: why wsPort is not readonly?

packages/lisk-p2p/src/peer.ts
```diff
@@ -27,31 +26,28 @@ export interface IPeerConfig {
...
 
-	public constructor(peerConfig: IPeerConfig) {
+	public constructor(peerConfig: PeerConfig) {
 		this.height = peerConfig.height;
 		this.ip = peerConfig.ip;
 		this.wsPort = peerConfig.wsPort;
 	}
 
-	public getHeight(): number {
+	public get Height(): number {
```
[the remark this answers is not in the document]
shuse2: this should be `height()`

packages/lisk-p2p/src/peer.ts
```diff
@@ -26,32 +25,29 @@ export interface IPeerConfig {
...
-		return this.height;
-	}
-	public getIp(): string {
-		return this.ip;
+	public get Height(): number {
+		return this._height;
 	}
-	public getWsPort(): number {
-		return this.wsPort;
+	public set Height(height: number) {
```
shuse2: if it has setter/getter without logic, why not public?

packages/lisk-p2p/src/peer.ts
```diff
@@ -26,32 +25,29 @@ export interface IPeerConfig {
...
+		this._height = height;
 	}
-	public setHeight(height: number): void {
-		this.height = height;
+	public get Ip(): string {
```
shuse2: this should be `ip`

packages/lisk-p2p/src/peer.ts
```diff
@@ -26,32 +25,29 @@ export interface IPeerConfig {
...
+		return this._ip;
 	}
-	public setIp(ip: string): void {
-		this.ip = ip;
+	public get WsPort(): number {
```
shuse2: same as `Height`, if it has setter/getter without logic, i think public is fine

packages/lisk-p2p/src/helpers.ts
```diff
@@ -0,0 +1,128 @@
...
+		if (numOfPeers === 1) {
+			const goodPeer: ReadonlyArray<Peer> = [
+				processedPeers[Math.floor(Math.random() * processedPeers.length)],
+			];
+
+			return { options, peers: goodPeer };
+		}
+
+		const selectedPeersList = Array(numOfPeers)
+			.fill({})
```
shuse2: if it's just for looping, u can fill with `0`

packages/lisk-p2p/src/helpers.ts
```diff
@@ -0,0 +1,128 @@
+/*
+ * Copyright © 2018 Lisk Foundation
+ *
```
shuse2: maybe better naming for this? something like `peer_selector`?

packages/lisk-p2p/src/helpers.ts
```diff
@@ -0,0 +1,128 @@
...
+}
+interface HistogramValues {
+	height: number;
+	histogram: Histogram;
+	max: number;
+}
+/* tslint:enable: readonly-keyword */
+export const selectPeers = (
+	peers: ReadonlyArray<Peer>,
+	options: PeerOptions = { blockHeight: 0 },
```
shuse2: maybe options should be the last argument?

packages/lisk-p2p/src/helpers.ts
```diff
@@ -0,0 +1,128 @@
...
+			peer &&
+			Math.abs(
+				(calculatedHistogramValues ? calculatedHistogramValues.height : 0) -
+					peer.Height,
+			) <
+				aggregation + 1,
+	);
+
+	// Select n number of peers
+	if (numOfPeers) {
```
shuse2: can we avoid this big `if`?

packages/lisk-p2p/src/helpers.ts
```diff
@@ -0,0 +1,128 @@
...
+		if (numOfPeers > processedPeers.length) {
+			throw new VError(
```
shuse2: if it still thrown an error, I think use default error or create custom error with extending Verror would be better.

packages/lisk-p2p/test/unit/helpers.ts
```diff
@@ -0,0 +1,106 @@
...
+				_ip: '18.28.48.1',
+				_wsPort: 5008,
+				_height: 645980,
+			},
+		];
+		describe('get list of n number of good peers', () => {
+			beforeEach(async () => {
+				peerList = initializePeerList();
+			});
+			it('should return an object', () => {
```
shuse2: space in between each `it` and `beforeEach`

packages/lisk-p2p/test/unit/peer.ts
```diff
@@ -0,0 +1,56 @@
...
+describe('peer', () => {
+	const defaultPeerConfig: PeerConfig = {
+		ip: '12.12.12.12',
+		wsPort: 5001,
+		height: 545776,
+	};
+
+	const defaultPeer = new Peer(defaultPeerConfig);
+	describe('#constructor', () => {
+		it('should be an object', () => {
```
shuse2: space between `it`

packages/lisk-p2p/src/helpers.ts
```diff
@@ -0,0 +1,128 @@
...
+		const optionsTemp = { ...options, blockHeight: 0 };
+
+		return { options: optionsTemp, peers: [] };
+	}
+
+	// Order peers by descending height
+	const sortedPeers = filteredPeers.sort((a, b) => b.Height - a.Height);
+
+	const aggregation = 2;
+	const defaultValue: HistogramValues = { height: 0, histogram: {}, max: -1 };
```
shuse2: here also, no need to define here?

packages/lisk-p2p/src/helpers.ts
```diff
@@ -0,0 +1,128 @@
...
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ *
+ */
+import { VError } from 'verror';
+import { Peer } from './peer';
+
+export interface PeerReturnType {
+	readonly options: PeerOptions;
```
shuse2: in core, I think it's returning calculated network height here.
options would be misleading, and returning the input doesn't make much sense.
we don't need to follow core exactly, and maybe network height where peers are selected from might be useful information.

packages/lisk-p2p/src/p2p.ts
```diff
@@ -12,48 +12,149 @@
...
+			peer &&
+			Math.abs(
+				(calculatedHistogramValues ? calculatedHistogramValues.height : 0) -
+					peer.Height,
+			) <
+				aggregation + 1,
+	);
+
+	// Select n number of peers
+	if (numOfPeers) {
```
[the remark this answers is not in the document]
shuse2: so if we don't specify `numOfPeers`, possibly returns big amount of peers?

packages/lisk-p2p/src/helpers.ts
ishantiw: `HistogramValues` also needs readonly keyword disabled, so I thought putting them in the block will look better as we will need multiple lines of `tslint:disable-next-line`

```
/* tslint:disable: readonly-keyword*/
interface Histogram {
	[key: number]: number;
}
interface HistogramValues {
	height: number;
	histogram: Histogram;
	max: number;
}
/* tslint:enable: readonly-keyword */
```

packages/lisk-p2p/src/peer.ts
ishantiw: I think I need to discuss with Jon about which fields should be kept readonly, as last time he only mentioned `ip`

packages/lisk-p2p/src/helpers.ts
ishantiw: `options` is actually not optional value whereas `numOfPeers` is, so I kept it at the end

packages/lisk-p2p/src/helpers.ts
ishantiw: earlier it was giving me lint error without defining it, restarted TS and it works now so I will remove

packages/lisk-p2p/src/helpers.ts
ishantiw: Actually, this simple condition at the top avoids unnecessary check of conditions if `numOfPeers` is `undefined` else we will have to check for `undefined` value in all conditions related to `numOfPeers`

packages/lisk-p2p/src/peer_selector.ts
```diff
@@ -0,0 +1,125 @@
...
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ *
+ */
+import { NotEnoughPeersError } from './errors';
+import { Peer } from './peer';
+
+// For future extension of return type
+export interface PeerReturnType {
```
jondubois: Do we need `PeerReturnType`? Maybe we should return `ReadonlyArray<Peer>` from the selectPeers method instead to keep it simple.

packages/lisk-p2p/src/peer.ts
```diff
@@ -30,46 +31,25 @@ export interface IPeerConfig {
 }
 
 export class Peer {
-	private _height: number;
-	private readonly _id: string;
-	private readonly _inboundSocket: any;
```
shuse2: aren't these private variable non-modifiable after constructor?
I think it was ok to be private

packages/lisk-p2p/src/peer_selector.ts
```diff
@@ -0,0 +1,125 @@
...
+	height: number;
+	histogram: Histogram;
+	max: number;
+}
+/* tslint:enable: readonly-keyword */
+export const selectPeers = (
+	peers: ReadonlyArray<Peer>,
+	selectionParams: PeerOptions = { lastBlockHeight: 0 },
+	numOfPeers?: number,
+): PeerReturnType => {
```
shuse2: if we are returning just one key from type, why not just use `ReadonlyArray<Peer>` as return value?

packages/lisk-p2p/src/peer_selector.ts
```diff
@@ -0,0 +1,125 @@
...
+			peer &&
+			Math.abs(
+				(calculatedHistogramValues ? calculatedHistogramValues.height : 0) -
+					peer.height,
+			) <
+				aggregation + 1,
+	);
+
+	// Select n number of peers
+	if (numOfPeers) {
```
shuse2: prefer fast return. maybe like this?
```
if (!numOfPeers) {
  return xxx;
}
```

packages/lisk-p2p/src/peer_selector.ts
```diff
@@ -0,0 +1,125 @@
...
+		if (numOfPeers > processedPeers.length) {
+			throw new NotEnoughPeersError(
+				`Requested no. of peers: '${numOfPeers}' is more than the available no. of good peers: '${
```
shuse2: prefer verbose `no. => number`

packages/lisk-p2p/src/peer_selector.ts
```diff
@@ -0,0 +1,125 @@
...
+	if (filteredPeers.length === 0) {
+		return { peers: [] };
+	}
+
+	// Order peers by descending height
+	const sortedPeers = filteredPeers.sort((a, b) => b.height - a.height);
+
+	const aggregation = 2;
+
+	const calculatedHistogramValues = sortedPeers.reduce(
```
jondubois: I know that this is partly because of the rules about avoiding `for` loops but I find this hard to read.
Is it possible to simplify this a bit? I'll approve anyway if it's too tricky.

packages/lisk-p2p/src/peer_selector.ts
```diff
@@ -0,0 +1,125 @@
...
+	// Perform histogram cut of peers too far from histogram maximum
+	const processedPeers = sortedPeers.filter(
+		peer =>
+			peer &&
+			Math.abs(
+				(calculatedHistogramValues ? calculatedHistogramValues.height : 0) -
+					peer.height,
+			) <
+				aggregation + 1,
+	);
```
jondubois: Spacing and indentation looks wrong here.

packages/lisk-p2p/src/index.ts
```diff
@@ -12,5 +12,8 @@
  * Removal or modification of this copyright notice is prohibited.
  *
  */
-export { P2P, IPeerReturnType, IPeerOptions } from './p2p';
-export { IPeerConfig, Peer } from './peer';
+export * from './errors';
+export * from './p2p';
+export * from './peer';
```
shuse2: I just realized, but probably in here only thing it needs to export is `p2p` and maybe `error` for library user

packages/lisk-p2p/test/unit/errors.ts
```diff
@@ -0,0 +1,86 @@
...
+				return Promise.resolve();
+			});
+
+			it('should set error message when passed an argument', () => {
+				return expect(notEnoughPeersError.message).to.eql(defaultMessage);
+			});
+		});
+	});
+
+	describe('#P2PError', () => {
```
shuse2: title should be PeerTransportError

packages/lisk-p2p/src/peer_selector.ts
ishantiw: It was done by prettify, but let me check if I can avoid prettify for this expression

packages/lisk-p2p/src/peer_selector.ts
ishantiw: I thought it will be easier to add any other field if we want to return something else from this function that can be added in future but I guess it could be a simple `ReadonlyArray<Peer>`

packages/lisk-p2p/src/peer_selector.ts
ishantiw: @shuse2 What do you think about it? should I change to a `forLoop` but that will also make me to use `let` keyword

packages/lisk-p2p/src/peer_selector.ts
ishantiw: I thought it will be easier to add any other field if we want to return something else from this function that can be added in future but I guess it could be a simple `ReadonlyArray<Peer>`

packages/lisk-p2p/src/peer.ts
ishantiw: If we don't make it public then we won't be able to read it from outside without a getter, also, if I change it to private then it will throw lint error `'id' is declared but its value is never read.`

shuse2: yes, originally it had private _id with getter id, so i think that was okay =0

packages/lisk-p2p/src/peer_selector.ts
ishantiw: @jondubois discussed with Shu and we removed some unnecessary initialization, and reduce will look a bit simpler now

packages/lisk-p2p/src/peer.ts
```diff
@@ -30,46 +31,45 @@ export interface IPeerConfig {
...
-			return PeerState.CONNECTED;
-		}
-
-		return PeerState.DISCONNECTED;
+	public get ipAddress(): any {
+		return this._ipAddress;
 	}
 
-	public get wsPort(): number {
+	public get wsPort(): any {
```
shuse2: `any`!!!

packages/lisk-p2p/test/unit/peer_selector.ts
```diff
@@ -0,0 +1,126 @@
...
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ *
+ */
+import { expect } from 'chai';
+import { initializePeerList } from '../utils/peers';
+import { PeerOptions, selectPeers } from '../../src/peer_selector';
+
+describe('helpers', () => {
```
shuse2: `peer selector`?

packages/lisk-p2p/src/peer_selector.ts
shuse2: 😢

PR #914, opened 2018-11-27
Fix not to expose custom external library definition - Closes #912

### LINKED ISSUE #912: Could not find a declaration file for module 'browserify-bignum'.
### Expected behavior
External user should be able to build using typescript.

### Actual behavior
```
../lisk-elements/packages/lisk-transactions/dist-node/utils/validation/validation.d.ts:1:20 - error TS7016: Could not find a declaration file for module 'browserify-bignum'. '/Users/shuse2/Documents/10_lisk/lisk-elements/packages/lisk-transactions/node_modules/browserify-bignum/bignumber.js' implicitly has an 'any' type.
  Try `npm install @types/browserify-bignum` if it exists or add a new declaration (.d.ts) file containing `declare module 'browserify-bignum';`
```

### Steps to reproduce
Try to use and build @liskhq/lisk-transactions from TypeScript environment

2.0-

shuse2:
### What was the problem?
Custom external library definition that we have was used in the input of exposed function.
Since those custom definition doesn't get compiled, it will fail on compilation

### How did I fix it?
Move the production `types` into each folder, and it will be packaged to the npm repository.
Added `/// <reference path="..." />`, which is not recommended in the docs. 
https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html

However, it seems like it's the only way to reference the custom external library definition

### How to test it?
`npm run build`
and check if the `validation.d.ts` and includes the reference tag

### Review checklist

* The PR resolves #912 
* Linting passes
* Tests pass


packages/lisk-transactions/package.json
```diff
@@ -57,6 +57,7 @@
 		"prepublishOnly": "npm run lint && npm test && npm run build && npm run build:check"
 	},
 	"dependencies": {
+		"@types/node": "10.10.1",
```
ishantiw: it should come after `@liskhq`

PR #917, opened 2018-11-28
Define listeners for transaction pool related events - Closes #873

### LINKED ISSUE #873: Define listeners for transaction pool related events
### Description

Transaction pool needs to define listeners for
- new blocks
- deleted blocks
- reverted rounds
which will affect the queues in each stage.

SargeKhan:
Add blockchain state change event listeners for the transaction pool.

* The PR resolves #873
* Linting passes
* Tests pass


packages/lisk-transaction-pool/src/transaction_pool.ts
```diff
@@ -40,62 +40,136 @@ interface Queues {
...
 
-	public onNewBlock(block: Block): void {}
+	public getProcessableTransactions(limit: number): ReadonlyArray<Transaction> {
+		return this._queues.ready.dequeueUntil(
+			queueCheckers.returnTrueUntilLimit(limit),
+		);
+	}
 
-	public onRoundRollback(delegates: ReadonlyArray<string>): void {}
+	public onDeleteBlock(block: Block): void {
```
nazarhussain: `TransactionPool` as a library should have nothing to do with other entities. It should just expose interfaces to be used based on business logic. And that business logic is not part of the library itself. Otherwise creating a library for transaction-pool does not make sense, everything could be just appear into Lisk Core Module. 

Being library It should expose interfaces like: 

1. moveTransactions(transactions, queue1, queue2)
2. resetTransactions(transactions) 
3. emptyQueue(queue)
4. resetPool

Additionally current approach is making `TransactionPool` to be tightly bounded to `Block` and other structures, which is completely out of its context. `TransactionPool` library should only deal with transactions and queues.

SargeKhan: Thanks for the review. In the transaction pool library, on certain events in the blockchain,  the transactions are moved from one queue to the other based on the behaviour of different types of transactions described in the specification of the LIP. The behavior of transactions is relevant to transactions in general and doesn't necessarily have to do with transactions in core. For-example, if a new block is received which contains transactions which have unique data, then transactions in the transaction pool which also have unique data should be reverified again, therefore they are moved to a different queue.

The scenario I just described is independent of the business logic of Lisk-core, it is based on how the transactions which require unique data behave in the blockchain. An important thing to note here is that transaction pool does not decide by itself which transaction has unique data, and which don't, it acquires this information by looking at the transaction definition. 

The business logic relevant to lisk-core will be supplied to the library, for-example, the functions for validating, verifying or applying transactions will be written in Core and transaction pool library will use these functions to move the transactions to the relevant queues.

The exposed interface you described looks like a queue library and not a transaction pool library. I think the behavior of transaction pool when a block is added/removed is relevant to core as well as sidechains, therefore, it should be incorporated in the library. 

As for the last point, I somewhat agree with your point, the transaction pool shouldn't care about the block structure, we can update these events, and use `onDeletedTransactions` etc. And hence, the transaction pool library will expect an interface of transactions.

packages/lisk-transaction-pool/test/transaction_pool.ts
```diff
@@ -1,27 +1,185 @@
...
+			),
+		};
+
+		transactionPool = new TransactionPool();
+		Object.keys(transactionPool.queues).forEach(queueName => {
+			sandbox
+				.stub((transactionPool as any)._queues, queueName)
+				.value(sinon.createStubInstance(Queue));
+		});
+		return;
```
shuse2: it should consistently make it async or return Promise

packages/lisk-transaction-pool/test/transaction_pool.ts
```diff
@@ -1,27 +1,185 @@
...
+			transactionPool.onNewBlock(block);
+			return expect(transactionPool.queues.validated.enqueueMany).to.be.calledWith([
+				...removedTransactions,
+				...removedTransactions,
+			]);
+		});
+	});
+
+	describe('#onRoundRollback', () => {
+		const roundDelegateAddresses = addresses;
```
shuse2: why do we need to reassign this?

packages/lisk-transaction-pool/src/transaction_pool.ts
```diff
@@ -40,62 +40,136 @@ interface Queues {
...
-	private validateTransactions: processTransactions;
-	private verifyTransactions: processTransactions;
-
-	public constructor(
-		validateTransactions: processTransactions,
-		verifyTransactions: processTransactions,
-		applyTransactions: processTransactions,
-	) {
-		this.queues = {
+	// tslint:disable-next-line variable-name
```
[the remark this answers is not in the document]
SargeKhan: It is throwing a linting error even after rebase. I don't see this "variable-name" rule in tslint.json file for development or transaction_improvement_experiment branch. Are you sure you added this rule?

packages/lisk-transaction-pool/src/transaction_pool.ts
```diff
@@ -40,62 +40,136 @@ interface Queues {
...
+				transaction.containsUniqueData && transaction.containsUniqueData(),
+		);
+		const transactionsOfTypesWithUniqueData = this.removeTransactionsFromQueues(
+			otherQueues,
+			queueCheckers.checkTransactionForTypes(blockTransactionsWithUniqueData),
+		);
+
+		// Add transactions which need to be reverified to the validated queue
+		this._queues.validated.enqueueMany([
+			...transactionsFromAffectedAccounts,
```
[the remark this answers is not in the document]
SargeKhan: Nice question! But I think these 2 arrays will always contain unique entries, this is because of the fact that `removeTransactionsFromQueues` *removes* transactions from the queues and not copies/clones them. Which means that if there is a transaction from an affected senderPublicKey and contains unique data, it will be removed from the queues by the `checkTransactionForSenderPublicKey` filter (which is executed first), and by the time `checkTransactionForTypes` filter is executed, it will not be in the queues.

packages/lisk-transaction-pool/src/transaction_pool.ts
```diff
@@ -40,62 +40,136 @@ interface Queues {
...
 
-	private processVerifiedTransactions(): void {}
+	public onRoundRollback(delegates: ReadonlyArray<string>): void {
+		// Move transactions from the verified, pending and ready queues to the validated queue which were sent from delegate accounts
+		const { received, validated, ...otherQueues } = this._queues;
+		const senderProperty: queueCheckers.TransactionFilterableKeys =
+			'senderPublicKey';
+		const transactionsFromAffectedAccounts = this.removeTransactionsFromQueues(
+			otherQueues,
+			queueCheckers.checkTransactionPropertyForValues(
```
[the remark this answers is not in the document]
SargeKhan: Because checkTransactionForSenderPublicKey expects transaction objects to be passed, where as here we don't  have transaction objects, and only a list of publicKeys.

packages/lisk-transaction-pool/src/transaction_pool.ts
```diff
@@ -40,62 +40,136 @@ interface Queues {
...
+					...this.queues.verified.transactions,
+			  ])
+			: true;
+	}
 
-	private verifyValidatedTransactions(
-		transactions: ReadonlyArray<Transaction>,
-	): void {}
+	private removeTransactionsFromQueues(
+		queues: Queues,
```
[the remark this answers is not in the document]
SargeKhan: Is it confusing because i am internally using this._queues? I can see that I don't need to do that, and I can just do: `.map(queueName => queues[queueName].removeFor(condition))`. Passing in queueName as input would just make the function calls a bit dirty, I think.

packages/lisk-transaction-pool/src/transaction_pool.ts
nazarhussain: Meaning of a library means which can be used by irrespective of the core. If our library can only be used in Lisk Core then there is no need of splitting up transaction pool into module and library. 

So transaction pool library is just provide a Queue interface specialized for transctions. So along with transaction library anyone should be able to use that library in any project. 

So rather than exposing events specified by our blockchain it should just expose interfaces based on behavior. e.g. instead of ‘onDeleteTransaction’ it should expose interface ‘deleteTransactions’ and leave the business logic to implementation. If we want to delete on new block or some other condition. 

And that’s just an example to straighten up the ideology of any library in lisk-elements.

nazarhussain: In short, instead of exposing event handlers, just expose interfaces. And yes trigger events that are strictly associated to library. e.g. if I want to know when a transaction changed its queu I can listen to an event of transaction pool.

shuse2: I don't think it's better to have
```
moveTransactions(transactions, queue1, queue2)
resetTransactions(transactions)
emptyQueue(queue)
resetPool
```
these interfaces. Since transaction pool should manage it's internal state.
but I agree that instead of implementing handler, it should provide more generic functions.
so instead of `onNewBlock` => `deleteTransactions` sounds good to me

packages/lisk-transaction-pool/src/transaction_pool.ts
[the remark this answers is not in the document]
shuse2: i thought I did, but it seems not ='(

packages/lisk-transaction-pool/src/transaction_pool.ts
```diff
@@ -13,89 +13,159 @@
...
+	}
 
-	public verifyTransaction(): void {}
+	public get queues(): Queues {
+		return this._queues;
+	}
 
-	private existsInTransactionPool(transaction: Transaction): boolean {}
+	public getProcessableTransactions(limit: number): ReadonlyArray<Transaction> {
+		return this._queues.ready.dequeueUntil(
```
shuse2: Wasnt the idea to keep it in the queue until we `removeConfirmedTransaction`?

SargeKhan: This function is not just the basic implementation, I'll update this function I implement this feature.

packages/lisk-transaction-pool/src/transaction_pool.ts
```diff
@@ -13,89 +13,159 @@
...
 	readonly senderPublicKey: string;
 	readonly type: number;
-	receivedAt?: Date;
 }
 
 export interface TransactionFunctions {
-	verifyTransactionAgainstOtherTransactions?(
-		transaction: Transaction,
+	containsUniqueData(): boolean;
+	verifyTransactionAgainstOtherTransactions(
```
mitsujutsu: Leaving a comment here for future naming revision: `verifyTransactionAgainstOtherTransactions` -> `verifyAgainstOtherTransactions`

PR #919, opened 2018-11-28
Extend APIError to have errors from core

### LINKED ISSUE #916: Include all high level error response in APIError
### Expected behavior
APIError should include error types to detect what error exactly it happened.

### Actual behavior
It only returns API status code

### Remarks
Required to check if Core API Error responses are consistent across the endpoints.

2.0-

shuse2:
### What was the problem?
APIError did not have enough information to show proper error on the user side.

### How did I fix it?
Add `errors` from APIResponse if exists in the error.

### How to test it?
```
cd packages/lisk-client
npm start
client = lisk.APIClient.createTestnetAPIClient()
client.votes.get({ "random": 1 }).catch((err) => console.log(err.message))
client.votes.get({ "random": 1 }).catch((err) => console.log(err.errno))
client.votes.get({ "random": 1 }).catch((err) => console.log(err.errors))

```

### Review checklist
* ~~Merge After #914~~
* The PR resolves #916 
* Linting passes
* Tests pass

### DIFF
--- a/packages/lisk-api-client/src/api_resource.ts
+++ b/packages/lisk-api-client/src/api_resource.ts
@@ -14,7 +14,7 @@
  */
 import Axios, { AxiosError, AxiosRequestConfig, AxiosResponse } from 'axios';
 import { APIClient } from './api_client';
-import { ApiResponse, HashMap } from './api_types';
+import { APIErrorResponse, APIResponse, HashMap } from './api_types';
 import { APIError } from './errors';
 
 const API_RECONNECT_MAX_RETRY_COUNT = 3;
@@ -42,12 +42,12 @@ export class APIResource {
 		error: Error,
 		req: AxiosRequestConfig,
 		retryCount: number,
-	): Promise<ApiResponse> {
+	): Promise<APIResponse> {
 		if (this.apiClient.hasAvailableNodes()) {
-			return new Promise<ApiResponse>(resolve =>
+			return new Promise<APIResponse>(resolve =>
 				setTimeout(resolve, REQUEST_RETRY_TIMEOUT),
 			).then(
-				async (): Promise<ApiResponse> => {
+				async (): Promise<APIResponse> => {
 					if (retryCount > API_RECONNECT_MAX_RETRY_COUNT) {
 						throw error;
 					}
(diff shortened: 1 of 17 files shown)


packages/lisk-api-client/src/api_resource.ts
```diff
@@ -67,26 +67,26 @@ export class APIResource {
...
 			.then((res: AxiosResponse) => res.data)
 			.catch(
-				(error: AxiosError): Error => {
+				(error: AxiosError): void => {
 					if (error.response) {
-						if (error.response.data && error.response.data.message) {
+						const { status } = error.response;
+						if (error.response.data) {
+							const {
+								error: errorString,
```
SargeKhan: where is errorString defined?

shuse2: this is the definition of `errorString`.
It's destructure with giving different name

packages/lisk-api-client/test/errors.ts
```diff
@@ -13,16 +13,15 @@
...
-import { APIError } from '../src/errors';
+import { APIError, APIErrorData } from '../src/errors';
 
 describe('api errors module', () => {
 	let apiError: APIError;
 	const defaultMessage = 'this is an error';
 	const defaultErrno = 401;
 
-	beforeEach(() => {
+	beforeEach(async () => {
```
mitsujutsu: why `async`?

shuse2: I will put it back, we should at some point move everything to async

packages/lisk-api-client/test/api_resource.ts
```diff
@@ -125,6 +125,34 @@ describe('API resource module', () => {
...
+				const statusCode = 300;
+				requestStub.rejects({
+					response: {
+						status: statusCode,
+						data: undefined,
+					},
+				});
+
+				return resource.request(defaultRequest, false).catch(err => {
+					return expect(err.message).to.equal('An unknown error has occurred.');
```
SargeKhan: This expectation does not match the test description.

packages/lisk-api-client/test/api_resource.ts
```diff
@@ -125,6 +125,34 @@ describe('API resource module', () => {
...
+				});
+			});
+
+			it('should reject with "An unknown error has occured." message if there is no data is supplied', () => {
```
SargeKhan: Is this test case any useful? Because in the following test, the request is rejected with a defined data property, and the expectation for `err.message` is the same.

shuse2: I think it's useful, because data exists but no message, and no data is treated in the different way in the code

PR #924, opened 2018-11-29
Add peer discovery algorithm - Closes#893

### LINKED ISSUE #893: Add basic discovery algorithm for lisk-p2p
### Description

In order to start developing the p2p network, we need a basic discovery system.

Add peer discovery algorithm that picks all the relevant peers from the network and returns inbound and outbound peers to establish a connection.

ishantiw:
### Description

Add peer discovery algorithm that picks all the relevant peers from the network and returns inbound and outbound peers to establish the connection.

-  A pure function returning the desired list of inbound/outbound list of peers
-  Add related tests

### Review checklist

* The PR resolves #893 #954 
* Linting passes
* Tests pass


packages/lisk-p2p/src/peer_discovery.ts
```diff
@@ -0,0 +1,53 @@
...
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ *
+ */
+import { Peer } from './peer';
+
+export interface DiscoverPeersReturntype {
```
shuse2: prefer `ReturnType`
but I think we should decide on naming convention, i used Input/Output, but ReturnType seems valid too.
or use more generic name?

jondubois: For return types, we should try to have more meaningful names. If another method elsewhere in the code also needs to return an object in the form `{inbound: [], outbound: []}` (now or in the future), it would be better if they had the same type; no need to couple to a specific method.

Maybe we can call this `PeerAdjacencyInfo`, `PeerConnectionSchema`, `PeerRelationInfo`, `PeerRelationSchema` or something like this.

packages/lisk-p2p/src/peer_discovery.ts
```diff
@@ -0,0 +1,53 @@
...
+	readonly inbound: ReadonlyArray<Peer>;
+	readonly outbound: ReadonlyArray<Peer>;
+}
+
+export interface PeersListType {
```
jondubois: Same here about naming semantics. Maybe `PeersListType` can be called `PeerListInfo` or even `PeerPeerList` or `PeerListOfPeer` something like this. More descriptive is better.

packages/lisk-p2p/src/peer_discovery.ts
```diff
@@ -0,0 +1,53 @@
...
+	readonly peer: Peer;
+	readonly peerList: ReadonlyArray<Peer>;
+}
+
+export const discoverPeers = (
+	peersList: ReadonlyArray<PeersListType>,
+): DiscoverPeersReturntype => {
+	// Make a list of all the peers
+	const allPeers = peersList.reduce<ReadonlyArray<Peer>>(
+		(allPeersArray, peer) => [...allPeersArray, ...peer.peerList],
```
jondubois: `peer` name is a bit confusing here since it's not a peer and it already has a `peer` property (`peer.peer` would be strange). Maybe it should be called `peerInfo` or `peerListInfo` (or whatever; depending on the type name).

packages/lisk-p2p/src/peer_discovery.ts
ishantiw: I like `PeerConnectionSchema`, sounds appropriate

packages/lisk-p2p/src/peer_discovery.ts
ishantiw: `PeerListOfPeer`

packages/lisk-p2p/src/peer_discovery.ts
ishantiw: @jondubois Do you think we can improve the argument type? instead of 
```
ReadonlyArray<{
	readonly peer: Peer;
	readonly peerList: ReadonlyArray<Peer>;
}>
```
We could also use,
`Map<string,  ReadonlyArray<Peer>>`
where string is the peer's `id`

packages/lisk-p2p/test/unit/peer_discovery.ts
```diff
@@ -0,0 +1,83 @@
...
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ *
+ */
+import { expect } from 'chai';
+import { PeerConfig, Peer } from '../../src/peer';
+import { initializePeerList } from '../utils/peers';
+import { discoverPeers } from '../../src/peer_discovery';
+
+describe('peer disocovery', () => {
```
jondubois: Typo `disocovery`.

packages/lisk-p2p/test/unit/errors.ts
```diff
@@ -83,4 +87,74 @@ describe('errors', () => {
...
+				return Promise.resolve();
+			});
+
+			it('should set error property info when passed as an argument', () => {
+				return expect(rpcGetPeersFailed)
+					.and.to.have.property('peerId')
+					.which.is.eql(peerId);
+			});
+
+			it('should set error property info when passed as an argument', () => {
```
jondubois: These two test cases look the same (including the assertions) `should set error property info when passed as an argument`.

packages/lisk-p2p/test/unit/errors.ts
```diff
@@ -83,4 +87,74 @@ describe('errors', () => {
 			});
 		});
 	});
+
+	describe('#RPCGetPeersFailed', () => {
```
jondubois: Don't worry about it for now but maybe we shouldn't be so detailed with the tests at this early stage in the project? Makes it harder to change things.

Testing `cause` property might be overkill since this functionality is related to the VError object and should already be covered by unit tests inside the VError library.

packages/lisk-p2p/test/unit/errors.ts
```diff
@@ -83,4 +87,74 @@ describe('errors', () => {
...
+
+			it('should set error property info when passed as an argument', () => {
+				expect(rpcGetPeersFailed)
+					.and.to.have.property('cause')
+					.and.to.be.a('function');
+
+				return expect(rpcGetPeersFailed.cause()).is.instanceOf(Error);
+			});
+
+			it('should set error property info when passed as an argument', () => {
```
jondubois: The two test cases have some overlap in coverage. The previous test case already checked that the rpcGetPeersFailed has a cause property which is a function. Also, it already checked that rpcGetPeersFailed.cause() is an instanceOf(Error). Maybe we should remove the first (less detailed) test case and only keep the more detailed one.

The test suite has to be maintained and is prone to change so at this stage of the project it's better to have fewer test cases which cover more application logic; that way we get more value out every test case.

packages/lisk-p2p/test/unit/peer_selector.ts
```diff
@@ -52,54 +52,60 @@ describe('peer selector', () => {
...
 			it('should return an array having all good peers', () => {
-				return expect(selectPeers(peerList, option, 0))
+				return expect(selectPeers(peerList, selectionParams, 0))
 					.and.be.an('array')
 					.and.of.length(3);
 			});
 
 			it('should return an array having all good peers ignoring requested negative number of peers', () => {
-				return expect(selectPeers(peerList, option, -1))
+				return expect(selectPeers(peerList, selectionParams, -1))
```
jondubois: This test case might be overkill. The caller should never pass a negative number; if they do, then we have a much bigger problem in another part of the code. We can never test every input so we should focus on inputs that are more relevant to program behaviour.

packages/lisk-p2p/src/peer_discovery.ts
```diff
@@ -0,0 +1,80 @@
...
+						error,
+						peer.id,
+					);
+				}),
+		),
+	);
+
+export const discoverPeers = async (
+	peers: ReadonlyArray<Peer>,
+	options: Options = { blacklistIds: [] },
```
jondubois: `Options` type - The word `Options` is a bit too generic?

packages/lisk-p2p/test/unit/errors.ts
ishantiw: Oh, I forgot to change the messages here :P

ishantiw: Initially, I added the whole VError cause and info object and removed it later but forgot to remove from the tests

packages/lisk-p2p/test/unit/errors.ts
ishantiw: You are right that we should not make very detailed test cases at this stage, I will remove the unnecessary ones. Regarding `cause`, I'm not testing VError functionality exactly but testing if our custom error `RPCGetPeersFailed ` is able to set cause property of `VError`

packages/lisk-p2p/test/unit/errors.ts
ishantiw: addressed

packages/lisk-p2p/src/errors.ts
```diff
@@ -31,3 +31,13 @@ export class NotEnoughPeersError extends VError {
 		this.name = 'NotEnoughPeersError';
 	}
 }
+
+export class RPCGetPeersFailedError extends VError {
```
shuse2: maybe this error is too specific.
maybe something like `RPCResponseError`?

packages/lisk-p2p/src/peer.ts
```diff
@@ -72,4 +80,8 @@ export class Peer {
 	public get wsPort(): number {
 		return this._wsPort;
 	}
+	/* tslint:disable:promise-function-async prefer-function-over-method no-unused-variable */
+	public async request(data: RPCRequest<any>): Promise<RPCResponse<any>> {
```
shuse2: maybe make it like this? also, we need procedure as first param?
public async request<T,K>(procedure: string, data?: T): Promise<RPCResponse<K>>

packages/lisk-p2p/src/peer_discovery.ts
```diff
@@ -0,0 +1,80 @@
...
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ *
+ */
+import { RPCGetPeersFailedError } from './errors';
+import { Peer, RPCRequest, RPCResponse } from './peer';
+
+export interface Options {
+	readonly blacklistIds: ReadonlyArray<string>;
```
shuse2: maybe `blacklist` are ok?

packages/lisk-p2p/src/peer_discovery.ts
```diff
@@ -0,0 +1,80 @@
...
+	);
+
+export const discoverPeers = async (
+	peers: ReadonlyArray<Peer>,
+	options: Options = { blacklistIds: [] },
+): Promise<ReadonlyArray<Peer>> => {
+	const rpcRequest: RPCRequest<void> = {
+		procedure: 'getPeers',
+	};
+	const peersOfPeer = await rpcRequestHandler(peers, rpcRequest);
```
shuse2: i think `peersOfPeer` is not really `peer` type.
for peer list and status we need type conversion?

packages/lisk-p2p/src/peer_discovery.ts
```diff
@@ -0,0 +1,80 @@
...
+			...peersList,
+		],
+		[],
+	);
+
+	const discoveredPeers = peersOfPeerFlat.reduce(
+		(uniquePeersArray: ReadonlyArray<Peer>, peer: Peer) => {
+			const found = uniquePeersArray.find(findPeer => findPeer.id === peer.id);
+
+			if (found) {
```
shuse2: in this simple case maybe something like
`return found ? uniquePeersArray :  [...uniquePeersArray, peer]`

packages/lisk-p2p/src/peer_discovery.ts
```diff
@@ -0,0 +1,80 @@
...
+				return uniquePeersArray;
+			}
+
+			return [...uniquePeersArray, peer];
+		},
+		[],
+	);
+
+	if (options.blacklistIds.length === 0) {
+		return Promise.resolve(discoveredPeers);
```
shuse2: i dont think you need to use Promise on async function

packages/lisk-p2p/src/peer_discovery.ts
ishantiw: Oh I think you reviewed old version of the code, now I'm simply returning list of peers as return

ishantiw:
Need validation for unknown types #940

ishantiw:
Rebased and addressed comments

packages/lisk-p2p/src/peer.ts
```diff
@@ -30,6 +30,7 @@ export interface PeerConfig {
 	readonly height?: number;
 	readonly inboundSocket?: any; // TODO: Type SCServerSocket
 	readonly os?: string;
+	readonly outboundSocket?: any; // TODO: Type SCServerSocket
```
jondubois: The `Peer` creates its own outboundSocket internally so it doesn't need one in the PeerConfig.

packages/lisk-p2p/test/unit/rpc_handler.ts
```diff
@@ -0,0 +1,181 @@
...
+				.and.of.length(3);
+		});
+
+		it('should return an array of instantiated peers', () => {
+			return expect(peersRPCHandler)
+				.to.be.an('array')
+				.eql(newlyCreatedPeers);
+		});
+
+		it('should return an array of instantiated peers', () => {
```
jondubois: Same name as test case above.

ishantiw:
Based on the discussion and suggestion from @jondubois I will move RPC calls within the Peer class to be more clear. So now, it will be simple `Peer.fetchPeers()` or `Peer.fetchStatus()` that will use the `Peer.request(procedureName, data)`

packages/lisk-p2p/src/peer.ts
```diff
@@ -127,6 +132,25 @@ export class Peer {
...
+	public async fetchPeers(): Promise<ReadonlyArray<PeerConfig>> {
+		try {
+			const response: P2PResponsePacket = await this.request<void>({
+				procedure: GET_ALL_PEERS_LIST_RPC,
+			});
+
+			return processPeerListFromResponse(response.data);
+		} catch (error) {
+			throw new RPCResponseError(
+				`Error when fetching peerlist of peer with peer ip ${
```
shuse2: isn't ip and port duplicate information?

packages/lisk-p2p/src/peer_discovery.ts
```diff
@@ -0,0 +1,67 @@
...
+		}
+		// Remove blacklist ids
+		const discoveredPeersFiltered = discoveredPeers.filter(
+			(peer: PeerConfig) =>
+				!filterPeerOptions.blacklist.includes(peer.ipAddress),
+		);
+
+		return discoveredPeersFiltered;
+	} catch (error) {
+		throw new DiscoveryError(
```
shuse2: should this error catch here?
if this function throws error, it is obvious enough that it's discovery error, and it doesn't give additional information

packages/lisk-p2p/src/response_handler.ts
```diff
@@ -69,10 +68,35 @@ export const instantiatePeerFromResponse = (peer: unknown): PeerConfig => {
...
 		os,
 		version,
 	};
 
 	return peerConfig;
 };
+
+interface RPCPeerListResponse {
+	readonly peers: ReadonlyArray<object>;
+	readonly success?: boolean; // Could be used in future
```
shuse2: isn't this current protocol schema which is used now?

packages/lisk-p2p/src/response_handler.ts
```diff
@@ -69,10 +68,35 @@ export const instantiatePeerFromResponse = (peer: unknown): PeerConfig => {
...
+}
+
+export const processPeerListFromResponse = (
+	response: unknown,
+): ReadonlyArray<PeerConfig> => {
+	if (!response) {
+		return [];
```
shuse2: if response is undefined, shouldnt' this throw error?
empty response should never happen, and it should be handle by developer?

packages/lisk-p2p/src/response_handler.ts
```diff
@@ -69,10 +68,35 @@ export const instantiatePeerFromResponse = (peer: unknown): PeerConfig => {
...
+		if (Array.isArray(peers)) {
+			const peerList = peers.map<PeerConfig>(instantiatePeerFromResponse);
+			return peerList;
+		}
+	} catch (error) {
+		throw error;
+	}
+
+	// Ignores any other value of response that is not an array
+	return [];
```
shuse2: same as above

packages/lisk-p2p/src/peer_discovery.ts
ishantiw: I thought to catch it here for logging but I guess we don't need it here

packages/lisk-p2p/src/response_handler.ts
ishantiw: yes, as we want it to be compatible with the current protocol. Especially for `list` RPC because it has different response type than others
`{success: boolean, peers: Array<Peers>}`

ishantiw:
Addressed

packages/lisk-p2p/src/peer.ts
```diff
@@ -44,17 +47,19 @@ export interface PeerConnectionState {
 	readonly outbound: ConnectionState;
 }
 
+const GET_ALL_PEERS_LIST_RPC = 'list';
+
 export class Peer {
 	private readonly _id: string;
-	private readonly _peerConfig: PeerConfig;
+	private readonly _peerConfig: PeerInfo;
```
jondubois: Rename _peerConfig -> _peerInfo?

packages/lisk-p2p/src/peer.ts
```diff
@@ -127,14 +132,31 @@ export class Peer {
...
+
 	public send<T>(packet: P2PMessagePacket<T>): void {
 		this._createOutboundSocket();
 		this._outboundSocket.emit(packet.event, {
 			data: packet.data,
 		});
 	}
 
-	public get peerConfig(): PeerConfig {
+	public get peerConfig(): PeerInfo {
```
jondubois: peerConfig -> peerInfo

packages/lisk-p2p/src/response_handler.ts
```diff
@@ -59,20 +64,35 @@ export const instantiatePeerFromResponse = (peer: unknown): PeerConfig => {
...
 	const wsPort = +rawPeer.wsPort;
 	const os = rawPeer.os && isAlpha(rawPeer.os.toString()) ? rawPeer.os : '';
-	const id = `${rawPeer.ip}:${rawPeer.wsPort}`;
 	const height =
 		rawPeer.height && isNumeric(rawPeer.height.toString())
 			? +rawPeer.height
 			: 0;
 
-	const peerConfig: PeerConfig = {
+	const peerConfig: PeerInfo = {
```
jondubois: peerInfo as the name of the constant.

packages/lisk-p2p/test/unit/peer.ts
```diff
@@ -14,14 +14,16 @@
  */
 import { expect } from 'chai';
 
-import { PeerConfig, Peer } from '../../src/peer';
+import { PeerInfo, Peer } from '../../src/peer';
 
 describe('peer', () => {
-	const defaultPeerConfig: PeerConfig = {
+	const defaultPeerConfig: PeerInfo = {
```
jondubois: defaultPeerConfig -> defaultPeerInfo

packages/lisk-p2p/test/utils/peers.ts
```diff
@@ -12,48 +12,54 @@
...
-	const peer4 = new Peer(peerOption4);
-	const peer5 = new Peer(peerOption5);
-
-	return [peer1, peer2, peer3, peer4, peer5];
+	return [peerOption1, peerOption2, peerOption3, peerOption4, peerOption5];
 };
+
+export const initializePeerList = (): ReadonlyArray<Peer> =>
+	initializePeerConfigList().map(
+		(peerConfig: PeerInfo) => new Peer(peerConfig),
```
jondubois: peerConfig -> peerInfo

ishantiw:
@jondubois addressed

ishantiw:
Merging this branch with the base as its already reviewed many times and all the comments have been resolved.

PR #928, opened 2018-12-04
Add functionality for expiring transactions - Closes #875

### LINKED ISSUE #875: Add functionality to expire transactions in transaction pool
### Description
Transaction pool needs to handle expiring the transaction, which does not get included in the transaction pool.

SargeKhan:
This PR adds a job for expiring transactions

* The PR resolves #875
* Linting passes
* Tests pass


packages/lisk-transaction-pool/src/job.ts
```diff
@@ -0,0 +1,22 @@
...
+	private readonly _job: () => Promise<ReadonlyArray<Transaction>>;
+
+	public constructor(
+		context: object,
+		job: () => Promise<ReadonlyArray<Transaction>>,
+		interval: number,
+	) {
+		this._interval = interval;
+		this._job = job.bind(context);
+		this.start();
```
shuse2: should this `start` called here?

packages/lisk-transaction-pool/src/job.ts
```diff
@@ -0,0 +1,22 @@
...
+	}
+
+	public start(): void {
```
shuse2: if we have start, shouldn't we be able to stop too?

packages/lisk-transaction-pool/src/transaction_pool.ts
```diff
@@ -20,16 +21,22 @@ export interface TransactionObject {
 	receivedAt?: Date;
 	readonly recipientId: string;
 	readonly senderPublicKey: string;
+	signatures?: ReadonlyArray<string>;
 	readonly type: number;
 }
 
 export interface TransactionFunctions {
 	containsUniqueData?(): boolean;
+	isExpired(date: Date): boolean;
```
shuse2: do we need to insert `date` here?

packages/lisk-transaction-pool/src/transaction_pool.ts
```diff
@@ -20,16 +21,22 @@ export interface TransactionObject {
...
 	verifyTransactionAgainstOtherTransactions?(
 		otherTransactions: ReadonlyArray<Transaction>,
 	): boolean;
 }
 
+interface TransactionPoolOptions {
+	readonly EXPIRE_TRANSACTIONS_JOB: number;
```
shuse2: this is not constant, as a option, so it should be camel case

packages/lisk-transaction-pool/test/queue_checkers.ts
```diff
@@ -1,16 +1,17 @@
 import { expect } from 'chai';
 import { Transaction } from '../src/transaction_pool';
 import * as queueCheckers from '../src/queue_checkers';
-import transactions from '../fixtures/transactions.json';
+import TransactionObjects from '../fixtures/transactions.json';
```
shuse2: it's better to keep it camel `transactionObjects`

packages/lisk-transaction-pool/test/transaction_pool.ts
```diff
@@ -181,4 +189,20 @@ describe('transaction pool', () => {
...
+		let removeTransactionsFromQueuesStub: sinon.SinonStub;
+		let expireTransactions: () => Promise<ReadonlyArray<Transaction>>;
+
+		beforeEach(() => {
+			removeTransactionsFromQueuesStub = sandbox.stub(transactionPool as any, 'removeTransactionsFromQueues');
+			return expireTransactions = (transactionPool as any)['expireTransactions'].bind(transactionPool);
+		});
+
+		it('should call removeTransactionsFromQueues once', () => {
+			return expireTransactions().then(() =>
```
shuse2: maybe here it's good to use eventually from chai-as-promised?

packages/lisk-transaction-pool/src/transaction_pool.ts
```diff
@@ -43,15 +50,20 @@ interface Queues {
 export class TransactionPool {
 	// tslint:disable-next-line variable-name
 	private readonly _queues: Queues;
+	private readonly EXPIRE_TRANSACTIONS_JOB: number;
```
shuse2: since this is not constant, it should be camelCase too

packages/lisk-transaction-pool/src/transaction_pool.ts
```diff
@@ -43,15 +50,20 @@ interface Queues {
...
 			received: new Queue(),
 			validated: new Queue(),
 			verified: new Queue(),
 			pending: new Queue(),
 			ready: new Queue(),
 		};
+		this.EXPIRE_TRANSACTIONS_JOB = EXPIRE_TRANSACTIONS_JOB;
+
+		// tslint:disable-next-line
+		new Job(this, this.expireTransactions, this.EXPIRE_TRANSACTIONS_JOB)
```
shuse2: probably it's better to manage the instance

packages/lisk-transaction-pool/src/transaction_pool.ts
```diff
@@ -155,6 +167,10 @@ export class TransactionPool {
 			: true;
 	}
 
+	private expireTransactions(): Promise<ReadonlyArray<Transaction>> {
```
shuse2: don't we need `async` here?

packages/lisk-transaction-pool/test/transaction_pool.ts
```diff
@@ -181,4 +189,20 @@ describe('transaction pool', () => {
...
 		});
 	});
-});
+
+	describe('#expireTransactions', () => {
+		let removeTransactionsFromQueuesStub: sinon.SinonStub;
+		let expireTransactions: () => Promise<ReadonlyArray<Transaction>>;
+
+		beforeEach(() => {
+			removeTransactionsFromQueuesStub = sandbox.stub(transactionPool as any, 'removeTransactionsFromQueues');
```
shuse2: do we need `as any` here?

packages/lisk-transaction-pool/src/transaction_pool.ts
```diff
@@ -43,15 +50,20 @@ interface Queues {
 export class TransactionPool {
 	// tslint:disable-next-line variable-name
 	private readonly _queues: Queues;
+	private readonly EXPIRE_TRANSACTIONS_JOB: number;
 
-	public constructor() {
+	public constructor({EXPIRE_TRANSACTIONS_JOB}: TransactionPoolOptions) {
```
shuse2: maybe we need default value?

packages/lisk-transaction-pool/test/transaction_pool.ts
```diff
@@ -181,4 +189,20 @@ describe('transaction pool', () => {
...
+			return expireTransactions = (transactionPool as any)['expireTransactions'].bind(transactionPool);
```
shuse2: prefer not to return and assign together

packages/lisk-transaction-pool/test/transaction_pool.ts
```diff
@@ -181,4 +189,20 @@ describe('transaction pool', () => {
...
+		});
+
+		it('should call removeTransactionsFromQueues once', () => {
```
shuse2: maybe it's good to use fakeTimer feature here to test the job

packages/lisk-transaction-pool/src/transaction_pool.ts
SargeKhan: I want to supply the current time (or the time to compare against) to this function so that every transaction is compared against the same time. Do you think it's unnecessary?

packages/lisk-transaction-pool/test/transaction_pool.ts
SargeKhan: I didn't know about it, i'll check it out.

packages/lisk-transaction-pool/src/transaction_pool.ts
SargeKhan: this is the interval of the expireTransaction job. Which will be passed to the constructor, shouldn't it be a constant?

packages/lisk-transaction-pool/test/transaction_pool.ts
SargeKhan: yes, because `removeTransactionsFromQueues` is a private function.

packages/lisk-transaction-pool/src/transaction_pool.ts
shuse2: if we pass from constructor, i think it's not a constant =p

packages/lisk-transaction-pool/src/transaction_pool.ts
shuse2: I think comparing against the same time is a good point.
maybe make it as optional like `isExpired(date?: Date)`? so if date doesn't exist, transaction will create new Date by themself ?

packages/lisk-transaction-pool/src/job.ts
shuse2: I think `job` should be cleanable in safe way, so keep the setInterval id and when stop is called, clean up?

packages/lisk-transaction-pool/src/transaction_pool.ts
SargeKhan: These interfaces are rudimentary, I think their names and parameters can be finalised after base Transaction PR is merged.

packages/lisk-transaction-pool/src/job.ts
```diff
@@ -0,0 +1,30 @@
...
+		job: () => Promise<T>,
+		interval: number,
+	) {
+		this._interval = interval;
+		this._job = job.bind(context);
+	}
+
+	public start(): void {
+		if (typeof this._id === 'undefined') {
+			this._id = setInterval(this._job, this._interval);
```
MaciejBaj: What if a job didn't finish yet? Does the function suppose to start jobs regardless of the results?
 
Let's say if a function execution lasts for 200ms and the interval is set to 100ms, the call stack will be exceeded sooner or later. 

There is a file of identical purpose in Lisk Core - `jobs_queue.js` which uses setTimeout. 

It would be better to keep one file and share it across the projects, rather than creating two of similar behaviours.

SargeKhan: I agree that maybe it's important to use setTimeout instead of setInterval while I think it's important to have metrics for the execution time of jobs. It's important because the transaction pool should fire events at certain times before forging blocks etc.

In lisk-elements, we aren't using jobs anywhere else as of yet. So, there is no code duplication inside the project. I don't know how we can share functionality between Core and SDK. We can develop a strategy, later on, to share common utility functions/classes between projects. But until then, I think it's okay to have it here. What do you think?

SargeKhan: @shuse2, I agree that we should devise a strategy for calculating metrics. I've created a new issue: #937. In this issue, we can discuss and finalize the approach we want to follow to calculate the performance of the new transaction pool as a whole.

SargeKhan:
I experimented with AsyncIterators to implement Timeout iterator, but after spending a day on trying to make it work, I found that we need to either use Node v.10 or add a new shim for Symbol.asyncIterator. Therefore, I gave up on using an iterator and simply used timeout wrapped in promise.

packages/lisk-transaction-pool/src/job.ts
```diff
@@ -0,0 +1,50 @@
...
+	private readonly _job: () => Promise<T>;
+	private _running = false;
+
+	public constructor(
+		context: object,
+		job: () => Promise<T>,
+		interval: number,
+	) {
+		this._interval = interval;
+		this._job = job.bind(context);
```
shuse2: should we bind it for them, or it's caller's responsibility?

packages/lisk-transaction-pool/src/transaction_pool.ts
```diff
@@ -151,6 +178,15 @@ export class TransactionPool {
 			: true;
 	}
 
+	private async expireTransactions(): Promise<ReadonlyArray<Transaction>> {
+		return Promise.resolve(
```
shuse2: If it's `async`, u don't need `Promise.resolve`?

packages/lisk-transaction-pool/src/job.ts
SargeKhan: I thought it's better to pass context to constructor function so we don't have to do it for every new job. Do you prefer the other way around?

shuse2: I think it's more clear to do it outside so it's easier to see what the context is when registering the job

packages/lisk-transaction-pool/src/job.ts
```diff
@@ -0,0 +1,50 @@
...
+	private async run(): Promise<void> {
+		// Base case for recursive function
+		if (!this._running) {
+			return;
+		}
+		await this.callJobAfterTimeout();
+
+		return this.run();
+	}
+}
```
MaciejBaj: Don't you have an agreement of empty line at the end of files?

packages/lisk-transaction-pool/src/job.ts
```diff
@@ -0,0 +1,50 @@
...
+				await this._job();
+
+				return resolve();
+			}, this._interval);
+		});
+	}
+
+	private async run(): Promise<void> {
+		// Base case for recursive function
+		if (!this._running) {
```
MaciejBaj: Not the other way around? This condition means I can never run a job if it's not running.

packages/lisk-transaction-pool/src/job.ts
```diff
@@ -0,0 +1,50 @@
...
+		interval: number,
+	) {
+		this._interval = interval;
+		this._job = job;
+	}
+
+	public async start(): Promise<void> {
+		if (!this._running) {
+			this._running = true;
+
```
MaciejBaj: remove empty line

packages/lisk-transaction-pool/src/job.ts
MaciejBaj: I guess a standard, as you see GitHub is marking it as an error by default. Lisk Core has it in every file.

packages/lisk-transaction-pool/src/job.ts
SargeKhan: We have a lint rule that there should be an empty line before a return statement.

packages/lisk-transaction-pool/src/job.ts
```diff
@@ -0,0 +1,50 @@
+export class Job<T> {
+	private _id: NodeJS.Timer | undefined;
+	private readonly _interval: number;
+	private readonly _job: () => Promise<T>;
+	private _running = false;
```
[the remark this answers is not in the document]
SargeKhan: I like `_active`. I'll update the name.

packages/lisk-transaction-pool/src/job.ts
shuse2: `npm run format` should solve the problem.
Prettier is doing it automatically

packages/lisk-transaction-pool/src/job.ts
MaciejBaj: Once you call `start` at least once, the run function will be checking `if (!this._running) ` condition indefinitely. Calling `stop` doesn't revert the Job to the initial state. You could achieve it by getting rid of recursion in favour of while:

```suggestion
		while (this._running) { // I would suggest _started or _active
		    await this.callJobAfterTimeout();
		} 
```
Lmk what do you think.

packages/lisk-transaction-pool/src/job.ts
MaciejBaj: Kk.

PR #935, opened 2018-12-10
Disable esModuleInterop setting - Closes #993

### LINKED ISSUE #933: Consider disabling esModuleInterop
### Description
Currently `esModuleInterop` is `true` at the root level of tsconfig.
As a library, this will limit TypeScript User to forcing them to use the same setting, too.

Therefore, it would be better if it's disabled.

### Version
2.0-

shuse2:
### What was the problem?
If esModuleInterop is true, there was no option to disable it on the client side.

### How did I fix it?
Turn off esModuleInterop on root level tsconfig, and fix the build and test error

### How to test it?
`npm t`

### Review checklist

* The PR resolves #933 
* Linting passes
* Tests pass

### DIFF
--- a/packages/lisk-transactions/src/utils/validation/validator.ts
+++ b/packages/lisk-transactions/src/utils/validation/validator.ts
@@ -12,9 +12,10 @@
  * Removal or modification of this copyright notice is prohibited.
  *
  */
-import Ajv from 'ajv';
-import addKeywords from 'ajv-merge-patch';
-import BigNum from 'browserify-bignum';
+import * as Ajv from 'ajv';
+// tslint:disable-next-line no-require-imports
+import addKeywords = require('ajv-merge-patch');
+import * as BigNum from 'browserify-bignum';
 import * as schemas from './schema';
 import {
 	isGreaterThanMaxTransactionId,
--- a/packages/lisk-api-client/src/resources/index.ts
+++ b/packages/lisk-api-client/src/resources/index.ts
@@ -12,13 +12,13 @@
  * Removal or modification of this copyright notice is prohibited.
  *
  */
-export { default as AccountsResource } from './accounts';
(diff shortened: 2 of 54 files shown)


packages/lisk-transactions/src/utils/validation/validator.ts
```diff
@@ -12,9 +12,10 @@
  * Removal or modification of this copyright notice is prohibited.
  *
  */
-import Ajv from 'ajv';
-import addKeywords from 'ajv-merge-patch';
-import BigNum from 'browserify-bignum';
+import * as Ajv from 'ajv';
+// tslint:disable-next-line no-require-imports
+import addKeywords = require('ajv-merge-patch');
```
SargeKhan, line 17: why do we need to use require here?

shuse2: when library is exposed only function like `module.exports = function (xx) {}`, it cannot `import *` and not as default either

shuse2:
Since we were using `--esModuleInterop ` and module as `common.js, `allowSyntheticDefaultImports` was true by default.

I think from the elements side, --esModuleInterop was working, but main reason is it enforces the users of elements to use the same setting

SargeKhan:
@shuse2, I was thinking that if we set `--allowSyntheticDefaultImports` to true and `--esModuleInterop` to false then maybe we can use default imports from within the application without enforcing it for clients. (Since the documentation says that `--esModuleInterop` emit __importStar and __importDefault, these emits are not fired on `--allowSyntheticDefaultImports`.)

shuse2:
@SargeKhan I see, I will try it out.
My first thought about `--allowSyntheticDefaultImports to true and --esModuleInterop to false `  was that will break the build file coz it does't have the Star or Default

shuse2:
Suspended until new transaction module is merged back

Issue #895, opened 2018-11-22
Add initialization function, constructor, start() and stop()


shuse2:
### Description

lisk-p2p needs to have functions to  and instantiate, start and stop the library.

on start, it should do 
- the seed connection
- initial discovery
- peer selection (for phase 1, it returns all peers)
- and connect them

on end, it should disconnect all peers safely

PR #936, opened 2018-12-11
Initialization start(), stop() and peer management - Addresses #894 and #895 

### LINKED ISSUE #894: Add peer management mechanism
### Description

lisk-p2p needs a mechanism to manage all the known peers.

The main purpose of the pool is to hold the connected peer objects, but it should contain
- get() using selection function similar to getGoodPeers
- addConnection(connection) which takes raw connection and assign/create peer
- addPeer(peer) which insert to pool
- handle disconnect and remove from pool
- disconnect all the peers

jondubois:
### What was the problem?

P2P node needs to be able start, connect to seed peers and initiate discovery.
As part of this, basic peer management functionality is required.

### How did I fix it?

Implemented:

- Basic node start functionality.
- Basic peer management.
- Initiate peer discovery through seed nodes.
- Basic peer connection.
- Peer request() and send() methods.
- Cleanup of existing code.

### How to test it?

`npm test`

### Review checklist

* The PR partially resolves #894 and #895 
* Linting passes
* Tests pass

### DIFF
--- a/packages/lisk-p2p/src/p2p.ts
+++ b/packages/lisk-p2p/src/p2p.ts
@@ -12,59 +12,197 @@
  * Removal or modification of this copyright notice is prohibited.
  *
  */
-/* tslint:disable:no-unused-expression */
+
+import { EventEmitter } from 'events';
+import http, { Server } from 'http';
+import os from 'os';
+import querystring from 'querystring';
+import socketClusterServer from 'socketcluster-server';
+
 import {
 	NetworkStatus,
-	P2PConfig,
 	P2PMessagePacket,
 	P2PNodeStatus,
-	P2PPenality,
+	P2PPenalty,
 	P2PRequestPacket,
 	P2PResponsePacket,
+	P2PConfig,
+	PeerInfo,
+	RPCResponsePeerList,
+	RPCResponsePeerInfo,
+	NodeInfo,
 } from './p2p_types';
 
 import { PeerPool } from './peer_pool';
 
-export class P2P {
-	private readonly peerPool: PeerPool;
+export const EVENT_NEW_INBOUND_PEER = 'newInboundPeer';
+export const EVENT_FAILED_TO_ADD_INBOUND_PEER = 'failedToAddInboundPeer';
+export const EVENT_NEW_PEER = 'newPeer';
+
+export interface IPeerReturnType {
+	readonly options: IPeerOptions;
+	readonly peers: ReadonlyArray<Peer>;
+}
+export interface IPeerOptions {
+	readonly [key: string]: string | number;
+}
+
+const BASE_10_RADIX = 10;
+
+export class P2P extends EventEmitter {
+	private readonly _config: P2PConfig;
+	private readonly _peerPool: PeerPool;
+	private readonly _httpServer: Server;
+	private readonly _scServer: any;
+	private readonly _newPeers: Set<string>;
+	private readonly _triedPeers: Set<string>;
+	private readonly _nodeInfo: NodeInfo;
 
 	public constructor(config: P2PConfig) {
-		this.peerPool = new PeerPool({
-			blacklistedPeers: config.blacklistedPeers,
-			connectTimeout: config.connectTimeout,
-			ipAddress: config.ipAddress,
-			seedPeers: config.seedPeers,
-			wsEngine: config.wsEngine,
+		super();
+		this._config = config;
+		this._nodeInfo = {
 			wsPort: config.wsPort,
-		});
+			os: os.platform(),
+			version: config.version,
+		};
+		this._newPeers = new Set();
+		this._triedPeers = new Set();
+
+		this._peerPool = new PeerPool();
+		this._httpServer = http.createServer();
+		this._scServer = socketClusterServer.attach(this._httpServer);
 	}
 
-	public applyPenalty = (penalty: P2PPenality): void => {
+	public applyPenalty = (penalty: P2PPenalty): void => {
 		penalty;
 	};
 	// TODO
 	public getNetworkStatus = (): NetworkStatus => true;
 	// TODO
+
 	public getNodeStatus = (): P2PNodeStatus => true;
-	// TODO
-	public request = async (
-		packet: P2PRequestPacket,
-	): Promise<P2PResponsePacket> => {
-		const response = packet;
+	// TODO: Implement
 
-		return Promise.resolve(response);
-	};
+	public async request<T>(
+		packet: P2PRequestPacket<T>,
+	): Promise<P2PResponsePacket> {
+		return Promise.resolve({ data: packet });
+	}
 
-	public send = (message: P2PMessagePacket): void => {
+	public send<T>(message: P2PMessagePacket<T>): void {
 		message;
 		// TODO
-	};
+	}
 	public setNodeStatus = (nodeStatus: P2PNodeStatus): void => {
 		nodeStatus;
 		// TODO
 	};
-	// TODO
-	public start = async (): Promise<void> => this.peerPool.start();
-	// TODO
-	public stop = async (): Promise<void> => this.peerPool.stop();
+
+	private async _startPeerServer(): Promise<void> {
+		this._scServer.on(
+			'connection',
+			(socket: any): void => {
+				const queryObject = querystring.parse(socket.request.url);
+				if (
+					typeof queryObject.wsPort !== 'string' ||
+					typeof queryObject.os !== 'string' ||
+					typeof queryObject.version !== 'string'
+				) {
+					super.emit(EVENT_FAILED_TO_ADD_INBOUND_PEER);
+				} else {
+					const wsPort: number = parseInt(queryObject.wsPort, BASE_10_RADIX);
+					const peerId = P2P.generatePeerIdFromPeerInfo({
+						ipAddress: socket.remoteAddress,
+						wsPort,
+					});
+					const existingPeer = this._peerPool.getPeer(peerId);
+					if (existingPeer === undefined) {
+						const peer = new Peer({
+							height: 0,
+							id: peerId,
+							inboundSocket: socket,
+							ipAddress: socket.remoteAddress,
+							os: queryObject.os,
+							version: queryObject.version,
+							wsPort,
+							nodeInfo: this._nodeInfo,
+						});
+						this._peerPool.addPeer(peer);
+						super.emit(EVENT_NEW_INBOUND_PEER, peer);
+						super.emit(EVENT_NEW_PEER, peer);
+					} else {
+						existingPeer.inboundSocket = socket;
+					}
+					this._triedPeers.add(peerId);
+				}
+			},
+		);
+		this._httpServer.listen(this._config.wsPort);
+
+		return new Promise<void>(resolve => {
+			this._scServer.once('ready', () => {
+				resolve();
+			});
+		});
+	}
+
+	public async _loadListOfPeerListsFromSeeds(
+		seedList: ReadonlyArray<PeerInfo>,
+	): Promise<ReadonlyArray<ReadonlyArray<Peer>>> {
+		return Promise.all(
+			seedList.map(async (seedPeer: PeerInfo) => {
+				const seedPeerId = P2P.generatePeerIdFromPeerInfo(seedPeer);
+				const peer = new Peer({
+					id: seedPeerId,
+					ipAddress: seedPeer.ipAddress,
+					wsPort: seedPeer.wsPort,
+					nodeInfo: this._nodeInfo,
+				});
+				/**
+				 * TODO LATER: For the LIP phase, we shouldn't add the seed peers to our
+				 * _peerPool and we should disconnect from them as soon as we've loaded their peer lists.
+				 */
+				this._newPeers.add(seedPeerId);
+				this._peerPool.addPeer(peer);
+				// TODO LATER: For the LIP phase, the 'list' request will need to be renamed.
+				const seedNodePeerListResponse = await peer.request<void>({
+					procedure: 'list',
+				});
+				const peerListResponse = seedNodePeerListResponse.data as RPCResponsePeerList;
+				// TODO ASAP: Validate the response before returning. Check that seedNodePeerListResponse.data.peers exists.
+
+				return peerListResponse.peers.map((peerObject: RPCResponsePeerInfo) => {
+					const peerOfPeerId = P2P.generatePeerIdFromPeerInfo({
+						ipAddress: peerObject.ip,
+						wsPort: peerObject.wsPort,
+					});
+
+					return new Peer({
+						id: peerOfPeerId,
+						ipAddress: peerObject.ip,
+						wsPort: peerObject.wsPort, // TODO ASAP: Add more properties
+						nodeInfo: this._nodeInfo,
+					});
+				});
+			}),
+		);
+	}
+
+	public start = async (): Promise<void> => {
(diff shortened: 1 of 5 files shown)


packages/lisk-p2p/src/peer.ts
```diff
@@ -12,59 +12,167 @@
...
+	readonly nodeInfo: NodeInfo;
+}
+
+export enum ConnectionState {
+	CONNECTING = 0,
+	CONNECTED = 1,
+	DISCONNECTED = 2,
+}
+
+export interface PeerState {
```
ishantiw: there are so many names revolving around Peer, maybe we need a descriptive name, maybe `PeerConnnectionState`?

packages/lisk-p2p/src/peer.ts
```diff
@@ -12,59 +12,167 @@
...
-// TODO: Use to create outbound socket connection inside peer object.
-// TODO: const socketClusterClient = require('socketcluster-client');
+import socketClusterClient from 'socketcluster-client';
 
 export interface PeerConfig {
 	readonly clock?: Date;
-	readonly height: number;
+	readonly height?: number;
 	readonly id: string;
 	/* tslint:disable:next-line: no-any */
```
ishantiw: shouldn't we also have `outboundSocket` in `PeerConfig`?

packages/lisk-p2p/src/peer.ts
```diff
@@ -12,59 +12,167 @@
...
-			return PeerState.CONNECTED;
-		}
 
-		return PeerState.DISCONNECTED;
+	public connect(): void {
+		const nodeInfo = this._nodeInfo;
+		this._outboundSocket = socketClusterClient.create({
+			hostname: this._ipAddress,
+			port: this._wsPort,
+			query: {
```
ishantiw: `query: this._nodeInfo`? since they have the same property names? or, maybe its good for clarity?

packages/lisk-p2p/src/peer.ts
```diff
@@ -12,59 +12,167 @@
...
+				wsPort: nodeInfo.wsPort,
+				os: nodeInfo.os,
+				version: nodeInfo.version, // TODO LATER: Provide nonce for compatibility with current protocol.
+			},
+		});
 	}
 
-	public get height(): number {
-		return this._height;
+	public disconnect(): void {
```
ishantiw: Should this be covered in this PR?

packages/lisk-p2p/src/peer.ts
```diff
@@ -12,59 +12,167 @@
...
+		return;
+	}
+
+	public async request<T>(
+		packet: P2PRequestPacket<T>,
+	): Promise<P2PResponsePacket> {
+		return new Promise<P2PResponsePacket>(
+			(resolve: (result: any) => void, reject: (result: any) => void): void => {
```
ishantiw: can't we write `resolve: (result: RawResponseBody) =>, reject: (result: Error | undefined) => void)` ?

packages/lisk-p2p/src/p2p.ts
```diff
@@ -12,59 +12,197 @@
...
+			});
+		});
+	}
+
+	public async _loadListOfPeerListsFromSeeds(
+		seedList: ReadonlyArray<PeerInfo>,
+	): Promise<ReadonlyArray<ReadonlyArray<Peer>>> {
+		return Promise.all(
+			seedList.map(async (seedPeer: PeerInfo) => {
+				const seedPeerId = P2P.generatePeerIdFromPeerInfo(seedPeer);
```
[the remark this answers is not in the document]
jondubois: Yes, I'll wait until your PR is merged.

packages/lisk-p2p/src/p2p.ts
```diff
@@ -12,59 +12,197 @@
...
+					id: seedPeerId,
+					ipAddress: seedPeer.ipAddress,
+					wsPort: seedPeer.wsPort,
+					nodeInfo: this._nodeInfo,
+				});
+				/**
+				 * TODO LATER: For the LIP phase, we shouldn't add the seed peers to our
+				 * _peerPool and we should disconnect from them as soon as we've loaded their peer lists.
+				 */
+				this._newPeers.add(seedPeerId);
```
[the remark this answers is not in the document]
jondubois: At that point, the seed peer has been created but has not been connected yet. It will be added to the triedPeers list after it connects (after request is called).

packages/lisk-p2p/src/p2p_types.ts
```diff
@@ -13,28 +13,58 @@
...
+	readonly event: string;
+	readonly data: T;
+}
+
+export interface PeerInfo {
+	readonly ipAddress: string;
+	readonly wsPort: number;
+}
+
+export interface NodeInfo {
```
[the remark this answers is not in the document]
jondubois: I renamed to `P2PNodeStatus`; it can be the same type as the argument passed to the `P2P` module's `setNodeStatus(nodeStatus: P2PNodeStatus): void` method.

packages/lisk-p2p/src/peer.ts
jondubois: Ah yep forgot to update that :p

packages/lisk-p2p/src/p2p.ts
```diff
@@ -12,59 +12,191 @@
...
+		return Promise.resolve({ data: packet });
+	}
 
-	public send = (message: P2PMessagePacket): void => {
+	public send<T>(message: P2PMessagePacket<T>): void {
 		message;
 		// TODO
+	}
+
+	public getNodeStatus(): P2PNodeStatus {
```
SargeKhan: Isn't it better to use getter and setter functions?

jondubois: I also had this thought I but wasn't sure because it deviates from the P2P library API design document https://docs.google.com/document/d/10_LmHqgf7EkPX-U3QIMyyWGBcHcEZGYp82cyfqodQNs/edit# but I guess it is inconsistent with our internal use of getters and setters so I'll make the chance and we can update the document at some point.

packages/lisk-p2p/src/p2p.ts
```diff
@@ -12,59 +12,191 @@
...
+
+	public start = async (): Promise<void> => {
+		await this._startPeerServer();
+		await this._loadListOfPeerListsFromSeeds(this._config.seedPeers);
 	};
-	public setNodeStatus = (nodeStatus: P2PNodeStatus): void => {
-		nodeStatus;
-		// TODO
+
+	public stop = async (): Promise<void> => {
```
SargeKhan: Shouldn't we stop listening to the server in here?

packages/lisk-p2p/src/peer.ts
```diff
@@ -12,59 +12,158 @@
...
+
+	public async request<T>(
+		packet: P2PRequestPacket<T>,
+	): Promise<P2PResponsePacket> {
+		return new Promise<P2PResponsePacket>(
+			(
+				resolve: (result: P2PResponsePacket) => void,
+				reject: (result: Error) => void,
+			): void => {
+				// TODO LATER: Change to LIP protocol format.
```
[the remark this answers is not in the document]
jondubois: The request *might* be different (would be nice to get rid of unnecessary JSON wrapper) but the response format might be the same though.

packages/lisk-p2p/src/peer.ts
```diff
@@ -12,59 +12,158 @@
...
+						data: packet.params,
+					},
+					(err: Error | undefined, responseData: unknown) => {
+						if (err) {
+							reject(err);
+
+							return;
+						}
+
+						if (responseData) {
```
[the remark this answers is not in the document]
jondubois: Yes, but I want to wait for the PR from @ishantiw because it adds some validation logic and I can add my validation function to the same file.

packages/lisk-p2p/src/peer.ts
```diff
@@ -12,59 +12,158 @@
...
-			return PeerState.CONNECTED;
-		}
 
-		return PeerState.DISCONNECTED;
+	public connect(): void {
+		const nodeStatus = this._nodeStatus;
+		this._outboundSocket = socketClusterClient.create({
+			hostname: this._ipAddress,
+			port: this._wsPort,
+			query: nodeStatus,
```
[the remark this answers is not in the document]
jondubois: No, socketcluster-client accepts an object and formats it into a string.

packages/lisk-p2p/src/p2p.ts
```diff
@@ -12,59 +12,191 @@
...
+		// This method resolves immediately.
+
+		return Promise.resolve();
 	};
-	// TODO
-	public start = async (): Promise<void> => this.peerPool.start();
-	// TODO
-	public stop = async (): Promise<void> => this.peerPool.stop();
+
+	public static generatePeerIdFromPeerInfo(peerInfo: PeerInfo): string {
```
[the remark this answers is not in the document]
jondubois: Yes. Would be better to wait for @ishantiw PR since it adds a utils file.

packages/lisk-p2p/src/p2p.ts
```diff
@@ -12,59 +12,191 @@
...
-	public setNodeStatus = (nodeStatus: P2PNodeStatus): void => {
-		nodeStatus;
-		// TODO
+
+	public stop = async (): Promise<void> => {
+		this._peerPool.disconnectAllPeers();
+		// In the current implementation, we don't need to shut down any child processes.
+		// This method resolves immediately.
+
+		return Promise.resolve();
```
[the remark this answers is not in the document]
jondubois: The inner function closure is not async.

packages/lisk-p2p/src/p2p.ts
```diff
@@ -12,59 +12,191 @@
...
 			wsPort: config.wsPort,
-		});
+			os: platform(),
+			version: config.version,
+		};
+		this._newPeers = new Set();
+		// TODO: this._triedPeers = new Set();
+
+		this._peerPool = new PeerPool();
+		this._httpServer = http.createServer();
```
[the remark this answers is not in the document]
jondubois: True, I was thinking of `start` more as 'start peer discovery' but in fact it should start the whole service and `stop` should also stop everything.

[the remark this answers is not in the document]
jondubois: Actually, I had second thoughts about this. That code doesn't actually start the http and WebSocket servers; it just creates them. The line `this._httpServer.listen(this._config.wsPort);` is where it actually starts listening. So I think the current way is better; that way the `this._httpServer` and `this._scServer` are never null or undefined.

packages/lisk-p2p/src/p2p.ts
```diff
@@ -12,59 +12,191 @@
...
+						ipAddress: peerObject.ip,
+						wsPort: peerObject.wsPort, // TODO ASAP: Add more properties
+						nodeStatus: this._nodeStatus,
+					});
+				});
+			}),
+		);
+	}
+
+	public start = async (): Promise<void> => {
```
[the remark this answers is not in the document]
jondubois: I think this answer on Stack Overflow makes a good comparison: https://stackoverflow.com/a/45882417/1181615

packages/lisk-p2p/src/peer.ts
```diff
@@ -12,59 +12,158 @@
...
+	P2PResponsePacket,
+	P2PNodeStatus,
+} from './p2p_types';
+
+import socketClusterClient from 'socketcluster-client';
 
 export interface PeerConfig {
 	readonly clock?: Date;
-	readonly height: number;
+	readonly height?: number;
```
ishantiw: why height is optional now?

packages/lisk-p2p/src/p2p_types.ts
```diff
@@ -13,28 +13,58 @@
...
+	readonly data: unknown;
+}
 
-export interface P2PResponsePacket {}
+export interface P2PMessagePacket<T> {
+	readonly event: string;
+	readonly data: T;
+}
+
+export interface PeerInfo {
```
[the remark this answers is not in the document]
ishantiw: any thoughts regarding this name?

packages/lisk-p2p/src/peer.ts
ishantiw: what about this?

packages/lisk-p2p/src/peer_pool.ts
```diff
@@ -14,107 +14,47 @@
...
-export interface PeerPoolConfig {
-	readonly blacklistedPeers?: ReadonlyArray<string>;
-	readonly connectTimeout: number;
-	readonly ipAddress?: string;
-	readonly seedPeers: ReadonlyArray<string>;
-	readonly wsEngine?: string;
-	readonly wsPort: number;
-}
-
 export class PeerPool extends EventEmitter {
```
ishantiw: does at any point later `EventEmitter` be used here? or it will be used by`P2P` as now it has reference to `PeerPool` instance, maybe we can remove it

packages/lisk-p2p/src/peer.ts
ishantiw: yeah maybe we can default it to one according to the core, I will also change it in my PR.

packages/lisk-p2p/src/peer_pool.ts
jondubois: Based on our last discussion, the current plan was for the peer to emit events directly on the PeerPool instead of making the PeerPool add event handlers to every peer and then manually cleaning from every peer.

jondubois: Based on our new discussion, I'm going for the approach of not passing the PeerPool to the Peer object and will make the PeelPool add and manage the the event handlers on the peer object.

packages/lisk-p2p/src/p2p.ts
```diff
@@ -12,59 +12,185 @@
...
+	private async _stopPeerServer(): Promise<void> {
+		// TODO ASAP: Test this and check for potential failure scenarios.
+		return new Promise<void>(resolve => {
+			this._scServer.close(() => {
+				resolve();
+			});
+		});
+	}
+
+	public async _loadListOfPeerListsFromSeeds(
```
ishantiw: a public method name should start without underscore

packages/lisk-p2p/src/peer.ts
```diff
@@ -12,64 +12,195 @@
...
 
 	public constructor(peerConfig: PeerConfig) {
-		this._id = peerConfig.id;
+		this._peerConfig = peerConfig;
 		this._ipAddress = peerConfig.ipAddress;
 		this._wsPort = peerConfig.wsPort;
+		this._id = Peer.constructPeerId(this._ipAddress, this._wsPort);
 		this._inboundSocket = peerConfig.inboundSocket;
-		this._height = peerConfig.height;
+		this._height = peerConfig.height === undefined ? 0 : peerConfig.height;
```
ishantiw: maybe use `(!peerConfig.height) ` that will cover more invalid values and also, I thought the default is `1`?

packages/lisk-p2p/src/p2p.ts
jondubois: Ah it should be private.

packages/lisk-p2p/src/peer.ts
jondubois: Default is 1 in the current core but 0 is a better default since the chain module may not have processed the genesis block yet. I think we should try to keep it 0 but if it ends up causing protocol compatibility issues then we can switch it to 1.

PR #940, opened 2018-12-11
Add validation and helper for rpc response - Closes#938

### LINKED ISSUE #938: Add validation for unknown types in p2p
### Description

In peer rpc calls, we are receiving the unknown types. We need to validate all the values and schema from these calls and then use it for any purpose. For example, we will only instantiate the peer object if it has valid ip address, port and height values.

ishantiw:
### Description

Add util functions to validate individual incoming values and helper functions to validate peer objects. Add related tests.

Closes #938


packages/lisk-p2p/test/unit/utils.ts
```diff
@@ -0,0 +1,90 @@
...
+ */
+import { expect } from 'chai';
+import { onlyDigits, validateIp, validatePort } from '../../src/utils';
+
+describe('utils', () => {
+	describe('#validateIp', () => {
+		const correctIp = '127.0.0.1';
+		const IpWithCharacters = '127k.0.0.1';
+		const IpWithLessBlocks = '127.0.0';
+		const IpWithSpaces = '127.0.0. 1';
```
jondubois: We don't need so many simple cases (one or two is enough?).

A more useful case would be related to IPv4 vs IPv6 formats.

Another case would be; what if the ip provided is not a string but an object or null.

packages/lisk-p2p/src/utils.ts
```diff
@@ -0,0 +1,73 @@
...
+ *
+ */
+const IP_NUMBER_OF_BLOCKS = 4;
+const IP_MAX_VAL_PER_BLOCK = 256;
+const IP_MIN_VAL_PER_BLOCK = 0;
+const MAX_VAL_PORT = 65535;
+const MIN_VAL_PORT = 1;
+// We disabled no-any because we want to validate the values in response from the network that could be any
+/* tslint:disable: no-any */
+export const validateIp = (ipAddress: any): boolean => {
```
SargeKhan: maybe it's better to use some schema checking library?

ishantiw: @SargeKhan should we use https://www.npmjs.com/package/@types/validator for validator.js?

ishantiw: @jondubois what do you think?

SargeKhan: Are these the only validations which will be required in p2p library or do you see the validators list expanding?

ishantiw: I think there will be a few more validations, but there will not be many. But there are some validations which are tricky, like ipv4/ipv6 which is better handled by the library. I just have validation for ipv4 right now and Jon said to have it for ipv6 too

packages/lisk-p2p/src/helpers.ts
```diff
@@ -0,0 +1,40 @@
+/*
```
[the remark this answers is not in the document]
shuse2: Prefer not to use `helper`.  something like `response_handler.ts`?

ishantiw:
Actually, I am mostly using the approach similar to what you use. It's just that for the validators and utils I felt this approach will look more clear for all input types. But like you said, we need to have consistency here. I can change to your approach for these tests if you guys agree.

packages/lisk-p2p/src/utils.ts
```diff
@@ -0,0 +1,73 @@
...
+		) {
+			return;
+		}
+		isValid = false;
+	});
+
+	return isValid;
+};
+
+export const validatePort = (port: any): boolean => {
```
shuse2: probably this should be `port: unknown`

packages/lisk-p2p/src/utils.ts
```diff
@@ -0,0 +1,73 @@
...
+	const portInt = +port;
+
+	return (
+		portInt >= MIN_VAL_PORT &&
+		portInt <= MAX_VAL_PORT &&
+		port === portInt.toString()
+	);
+};
+
+export const onlyDigits = (input: any): boolean => {
```
shuse2: same as above

packages/lisk-p2p/src/utils.ts
shuse2: For schema check, we use https://github.com/epoberezkin/ajv.
For IP, I think core was using some library, which we can use it too.

packages/lisk-p2p/src/validators.ts
```diff
@@ -0,0 +1,39 @@
...
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ *
+ */
+import { onlyDigits, validateIp, validatePort } from './utils';
+
+interface ResponsePeerObject {
+	readonly height: string | number;
```
shuse2: why is this expecting string or number? shouldn't it be one way or another?

packages/lisk-p2p/src/validators.ts
```diff
@@ -0,0 +1,39 @@
...
+	readonly ip: string;
+	readonly wsPort: string | number;
```
shuse2: same as above

packages/lisk-p2p/src/helpers.ts
```diff
@@ -0,0 +1,42 @@
...
+
+interface RawPeerObject {
+	readonly height: string | number;
+	readonly ip: string;
+	readonly os?: string;
+	readonly version?: string;
+	readonly wsPort: string | number;
+}
+
+export const instantiatePeerFromResponse = (peer: unknown): boolean | Peer => {
```
shuse2: Maybe better not to return two different type

packages/lisk-p2p/src/utils.ts
ishantiw: if I use `unknown`, `const port = +port` will throw error `Object is of type 'unknown'.`

packages/lisk-p2p/src/validators.ts
ishantiw: From the response, we can expect either number or string, and both could be valid, but after we sanitize the response then we instantiate peer object with the port as number type. Or should we only use, one type?

packages/lisk-p2p/src/utils.ts
ishantiw: So, in core, `ip` is used in module and logic and `net` in app.js. which one to use or we should use `validator.js` that covers everything?

ishantiw:
I have moved all the validation stuff of rpc response to `response_handler.ts` and I'm using `validator.js` lib for validation which has validation for` ip4/6` and `port` with a lot of other validations so makes sense to use just one library for all the validations in p2p. Also, `validator.js` is very popular with no dependencies. 

I also refactored test cases according to the general approach we discussed. 

@jondubois @SargeKhan @shuse2 Please review it again.

packages/lisk-p2p/test/unit/response_handler.ts
```diff
@@ -0,0 +1,116 @@
...
+			it('should return true for IPv6', () => {
+				const peer = {
+					ip: '2001:0db8:85a3:0000:0000:8a2e:0370:7334',
+					wsPort: '4001',
+				};
+
+				return expect(checkPeerAddress(peer)).to.be.true;
+			});
+
+			it('should return false', () => {
```
jondubois: We should try to be more descriptive with the names of the test cases. But I think this is a minor thing.

packages/lisk-p2p/src/utils.ts
ishantiw: this the file has been removed now and validation is done with an external lib

packages/lisk-p2p/src/validators.ts
ishantiw: addressed

packages/lisk-p2p/src/response_handler.ts
```diff
@@ -0,0 +1,78 @@
...
+	if ((!isIP(ip, IPV4_NUMBER) && !isIP(ip, IPV6_NUMBER)) || !isPort(wsPort)) {
+		return false;
+	}
+
+	return true;
+};
+
+export const instantiatePeerFromResponse = (
+	peer: unknown,
+): PeerConfig | Error => {
```
shuse2: it's not returning error anywhere?

packages/lisk-p2p/src/response_handler.ts
```diff
@@ -0,0 +1,78 @@
...
+			: '';
+	const id = `${rawPeer.ip}:${rawPeer.wsPort}`;
+	const height =
+		rawPeer.height && isNumeric(rawPeer.height.toString())
+			? +rawPeer.height
+			: 1;
+
+	const peerConfig: PeerConfig = {
+		ipAddress: rawPeer.ip,
+		wsPort: +rawPeer.wsPort,
```
shuse2: we want to handle port as number internally, but return it  as string?
why not make it consistent?

packages/lisk-p2p/src/response_handler.ts
```diff
@@ -0,0 +1,78 @@
...
+	if (!checkPeerAddress(peer)) {
+		return new InvalidPeerAddress(`Invalid Peer Ip or Port`);
+	}
+
+	const rawPeer = peer as RawPeerObject;
+
+	const os = rawPeer.os && isAlpha(rawPeer.os.toString()) ? rawPeer.os : '';
+	const version =
+		rawPeer.version && isAlpha(rawPeer.version.toString())
+			? rawPeer.version
```
shuse2: maybe we should throw if version is empty. (maybe this relates to hand shake).
Please confirm with the core

packages/lisk-p2p/src/response_handler.ts
```diff
@@ -0,0 +1,78 @@
...
+			: '';
+	const id = `${rawPeer.ip}:${rawPeer.wsPort}`;
+	const height =
+		rawPeer.height && isNumeric(rawPeer.height.toString())
+			? +rawPeer.height
+			: 1;
+
+	const peerConfig: PeerConfig = {
```
shuse2: what's the difference between `RawPeer` and `PeerConfig`?

packages/lisk-p2p/src/response_handler.ts
ishantiw: it is returning at [response_handler.ts#L52](https://github.com/LiskHQ/lisk-elements/blob/062056245576654306d35340c94e58722534d334/packages/lisk-p2p/src/response_handler.ts#L52)

packages/lisk-p2p/src/response_handler.ts
ishantiw: but here I'm converting it into integer and returning it

packages/lisk-p2p/src/response_handler.ts
ishantiw: I will check this with Jon and core team

packages/lisk-p2p/src/response_handler.ts
ishantiw: actually, `RawPeer ` will be replaced with `ProtocolPeerInfo` in `p2p_types` which will come from @jondubois  PR. The only difference is that we want a different type for a peer info in response from the network and most of the values will be string, that will be later converted to specific types when creating their corresponding `PeerConfig`

packages/lisk-p2p/src/response_handler.ts
ishantiw: as discussed, I will use `semver` that we use in core to validate version and throw the error if invalid

packages/lisk-p2p/src/response_handler.ts
ishantiw: Discussed with @jondubois to throw an error instead of returning an error to have consistency across the p2p lib

packages/lisk-p2p/src/response_handler.ts
ishantiw: another reason of not using `PeerConfig` is that we are using `ipAddress` instead of `ip` in PeerConfig

jondubois: `PeerConfig` and `ProtocolPeerInfo` have almost the same properties (except for `ip` vs `ipAddress`) but the main reason why we declared them as different types is to decouple internal logic from the external protocol.

If we decide to go with a consistent type for everything (external and internal data), then our internal logic will be more tightly coupled to the current protocol.

packages/lisk-p2p/test/unit/response_handler.ts
```diff
@@ -0,0 +1,170 @@
...
+						wsPort: 5393,
+						os: '',
+						height: 23232,
+						version: '3.4.5-alpha.9',
+					});
+			});
+		});
+
+		describe('for invalid peer response object', () => {
+			it('should return false for invalid peer values', () => {
```
jondubois: You should update the text since it's now throwing instead of returning.

ishantiw: addressed

packages/lisk-p2p/src/response_handler.ts
```diff
@@ -0,0 +1,81 @@
...
+
+interface RawPeerObject {
+	readonly height?: string;
+	readonly ip: string;
+	readonly os?: string;
+	readonly version?: string;
+	readonly wsPort: string;
+}
+
+export const checkForValidPeerAddress = (
```
shuse2: check and valid is similar meaning.
maybe `validPeerAddress` or `validatePeerAddress`?

PR #950, opened 2018-12-17
Need to add rules to tslint.json to account for recent discussion/meeting about rules - Closes #949

### LINKED ISSUE #949: Need to add rules to tslint.json to account for recent discussion/meeting about rules.
### Expected behavior

The TSLint changes which we discussed in the last meeting have not been added to the tslint.json.

### Actual behavior

TSLint is currently too strict in some areas.

### Steps to reproduce

`npm run lint`

Latest lisk-elements branches.

jondubois:
### What was the problem?

Some rules were against what we had decided in an earlier meeting - Others were slowing down development in the lisk-p2p module and need to be disabled temporarily while development hasn't yet settled.

### How did I fix it?

Added rules in tslint.json.
Added comments. See https://github.com/LiskHQ/lisk-elements/issues/934

### How to test it?

`npm run lint`

### Review checklist

* The PR resolves #949
* Linting passes
* Tests pass


jondubois:
I disabled the rule about shorthand arrow notation because it might encourage a dangerous pattern:

It's good for simple one-line expressions like `() => doSomething()` but can be confusing with multi-line expressiones.

For example, it doesn't allow:

```
return peerListResponse.peers.map((peerObject: ProtocolPeerInfo) => {
  return new Peer({
    ipAddress: peerObject.ip,
    wsPort: peerObject.wsPort, // TODO ASAP: Add more properties
    nodeStatus: this._nodeStatus,
  });
});
```

It wants:

```
return peerListResponse.peers.map((peerObject: ProtocolPeerInfo) => new Peer({
    ipAddress: peerObject.ip,
    wsPort: peerObject.wsPort, // TODO ASAP: Add more properties
    nodeStatus: this._nodeStatus,
  }));
```

 If someone is not careful, they might not realise that the parentheses are missing.

tslint.json
```diff
@@ -27,6 +27,16 @@
...
+		// Better to have functionally similar items be close together.
+		"ordered-imports": false,
+		// We need private class properties to start with an underscore.
+		"variable-name": [true, "allow-leading-underscore"],
+		// Better to have functionally similar items be close together.
+		"member-ordering": false,
+		// This rule can sometimes encourage bad practices; better leave this to the PR process.
+		"prefer-function-over-method": false,
+		// The shorthand arrow notation is often more difficult and error-prone to refactor.
+		"arrow-return-shorthand": true
```
shuse2: In general, I think this is just which coding style we choose. Once we get use to it, I dont think it's so confusing.
Also, this rule is applied from `lisk-template` so I would prefer to keep it as all the project suppose to be following.
For example
```
{
// ...
  (peerObject: ProtocolPeerInfo) => new Peer({
      ipAddress: peerObject.ip,
      wsPort: peerObject.wsPort, // TODO ASAP: Add more properties
      nodeStatus: this._nodeStatus,
    });
};
```
or if it's just returning the object
```
{
// ...
  (peerObject: ProtocolPeerInfo) => ({
      ipAddress: peerObject.ip,
      wsPort: peerObject.wsPort, // TODO ASAP: Add more properties
      nodeStatus: this._nodeStatus,
    });
}
```

tslint.json
```diff
@@ -27,6 +27,16 @@
 		"readonly-array": true,
 		"readonly-keyword": [true, "ignore-class"],
 		"strict-boolean-expressions": false,
-		"strict-type-predicates": false
+		"strict-type-predicates": false,
+		// Better to have functionally similar items be close together.
+		"ordered-imports": false,
```
shuse2: I feel like imports can be ordered by this rule mainly for merging problem.

tslint.json
```diff
@@ -27,6 +27,16 @@
...
+		// We need private class properties to start with an underscore.
+		"variable-name": [true, "allow-leading-underscore"],
+		// Better to have functionally similar items be close together.
+		"member-ordering": false,
+		// This rule can sometimes encourage bad practices; better leave this to the PR process.
+		"prefer-function-over-method": false,
```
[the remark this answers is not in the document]
jondubois: There are some cases:
- Abstract methods: Maybe the method just throws an error (so it doesn't use `this`); if it's designed to be overridden.
- The method is not yet fully implemented; it doesn't use `this` right now, but it will soon, but we want to commit the code anyway.
- One method which doesn't use `this` is functionally related to another method which does use `this`. For example `myInstance.startNetwork()` and `myInstance.stopNetwork()`; if one of these methods needs `this` but not the other, it doesn't make sense that one of the methods should be static or standalone: `MyClass.stopNetwork()`. Maybe the first method needs to read `this._config` but the stop method doesn't. 
- The usage of `this` within a method is an implementation detail and should not be a factor when deciding how a function should be exposed.

tslint.json
jondubois: My argument is that this would be better judged on a case-by-case basis than with a hard rule. I don't think that rule adds any value at all.

tslint.json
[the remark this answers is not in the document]
shuse2: > Abstract methods: Maybe the method just throws an error (so it doesn't use this); if it's designed to be overridden.
The method is not yet fully implemented; it doesn't use this right now, but it will soon, but we want to commit the code anyway.

Typescript wise abstract method shouldn't have implementation. Javascript wise, there is no abstract method. If it needs to be used in that way, we could say this is an exception, which shouldn't be the general case.

> One method which doesn't use this is functionally related to another method which does use this. For example myInstance.startNetwork() and myInstance.stopNetwork(); if one of these methods needs this but not the other, it doesn't make sense that one of the methods should be static or standalone: MyClass.stopNetwork(). Maybe the first method needs to read this._config but the stop method doesn't.

this is definitely a valid case against this, but I think we can put this as an exception. 

> The usage of this within a method is an implementation detail and should not be a factor when deciding how a function should be exposed. I don't think that we should let the linter tell us what is the right interface based on our implementation.

I think this is important to prevent meaning less state function, and prefer stateless functions.
Whether to expose function or not is not limited by this rule imo.

tslint.json
shuse2: I think this is more of just for consistency. Rather than case-by-case, I would like to keep it consistent

tslint.json
jondubois: We agreed in the meeting about member ordering, I wasn't sure about import ordering but I thought that the same logic applied.

shuse2: yes, but i don't think we put a lot of meaning into the import order as member ordering

tslint.json
shuse2: https://github.com/airbnb/javascript#arrows--implicit-return
this is the related eslint enabled in lisk-template

tslint.json
ishantiw: I think ordered imports are fine but not member ordering

jondubois:
@shuse2 I removed the controversial rules.

ishantiw:
Should we cover issue #934 in this PR as well, we still have redundant rules which were pointed out in that issue

shuse2:
@ishantiw lets keep them separate =)

PR #957, opened 2018-12-20
Base transaction tests - Closes #956

### LINKED ISSUE #956: Update tests for Base Transaction Class
### Description

Update tests for base transaction

mitsuaki-u:
### Description

Updated tests for base transaction

### Review checklist

* The PR resolves #956
* Linting passes
* Tests pass


packages/lisk-transactions/fixtures/valid_multisignature_transaction.json
```diff
@@ -0,0 +1,18 @@
+{
```
SargeKhan: The file should probably be named as valid_transaction_from_multisignature_account

packages/lisk-transactions/fixtures/valid_second_signature_transaction.json
```diff
@@ -0,0 +1,18 @@
+{
```
SargeKhan: valid_transaction_from_second_signature_account

packages/lisk-transactions/src/transactions/base.ts
```diff
@@ -106,14 +106,22 @@ export abstract class BaseTransaction {
...
 			asset: this.assetToJSON(),
 			receivedAt: this.receivedAt,
 		};
 
-		return transaction;
+		if (!this.signSignature) {
+			return transaction;
+		}
+
+		const signedTransaction = {
```
SargeKhan: shouldn't it be "secondSignedTransaction"?

packages/lisk-transactions/src/transactions/base.ts
```diff
@@ -203,10 +212,8 @@ export abstract class BaseTransaction {
 
 	public validate(): TransactionResponse {
 		const errors: TransactionError[] = [];
-		const transactionBytes = Buffer.concat([
-			this.getBasicBytes(),
-			this.getAssetBytes(),
-		]);
+		const transactionBytes = this.getBasicBytes();
```
SargeKhan, line 225: Why don't we need to get assetBytes now?

packages/lisk-transactions/test/utils/sign_and_verify.ts
```diff
@@ -27,12 +28,194 @@ import { TransactionError } from '../../src/errors';
...
+				defaultSecondSignatureTransaction.senderPublicKey,
+				defaultSecondSignatureTransaction.signature,
+				defaultTransactionBytes,
+			);
+			expect(cryptographyHashStub).to.be.calledOnce;
+		});
+
+		it('should call cryptography verifyData', async () => {
+			const cryptographyVerifyDataStub = sandbox
+				.stub(cryptography, 'hash')
```
SargeKhan: I think you missed changing the function name here.

packages/lisk-transactions/test/utils/sign_and_verify.ts
```diff
@@ -27,12 +28,194 @@ import { TransactionError } from '../../src/errors';
...
+
+	describe('#verifyMultisignatures', () => {
+		const defaultMultisignatureTransaction = addDate(
+			validMultisignatureTransaction,
+		);
+		const defaultTransactionBytes = Buffer.from(
+			'00de46a00424193236b7cbeaf5e6feafbbf7a791095ea64ec73abde8f0470001fee5d39d9d3c9ea25a6b7c648f00e1f50500000000746865207265616c2074657374',
+			'hex',
+		);
+		const memberKeys = [
```
SargeKhan: Shouldn't these be taken from the validMultisignatureAccount?

packages/lisk-transactions/test/utils/sign_and_verify.ts
```diff
@@ -27,12 +28,194 @@ import { TransactionError } from '../../src/errors';
...
+				defaultTransactionBytes,
+			);
+
+			expect(verified).to.be.false;
+			expect((errors as ReadonlyArray<TransactionError>)[0])
+				.to.be.instanceof(TransactionError)
+				.and.have.property('message', `Sender does not have valid multimin`);
+		});
+
+		it('should return a verification fail response with invalid signatures', () => {
```
SargeKhan: This should be an async function

packages/lisk-transactions/test/transactions/base.ts
```diff
@@ -17,647 +17,498 @@ import { BaseTransaction } from '../../src/transactions/base';
...
+			expect(baseTransaction)
 				.to.have.property('isMultisignature')
 				.and.be.a('boolean');
 		});
 
-		describe('when transaction has empty `signatures`', () => {
-			it('should set isMultisignature to false', () => {
-				return expect(baseTransaction.isMultisignature).to.be.false;
-			});
+		it('should set isMultisignature to true with a multisignature transaction', async () => {
```
SargeKhan: Do we set isMultisignature to true even when it doesn't have signatures but the account is a multisignature?

packages/lisk-transactions/test/transactions/base.ts
```diff
@@ -17,647 +17,498 @@ import { BaseTransaction } from '../../src/transactions/base';
...
-			});
-			it('should set isMultisignature to true', () => {
-				return expect(baseTransaction.isMultisignature).to.be.true;
-			});
+		it('should set isMultisignature to false with non-multisignature transaction', async () => {
+			expect(baseTransaction.isMultisignature).to.be.false;
 		});
 
-		describe('when given a transaction with invalid types', () => {
+		it('should throw a transaction multierror with an invalid typed transaction', async () => {
```
SargeKhan: The description is maybe incorrect here.

packages/lisk-transactions/test/transactions/base.ts
```diff
@@ -17,647 +17,498 @@ import { BaseTransaction } from '../../src/transactions/base';
...
 
-				expect(id).to.be.eql(baseTransaction.id);
-				expect(errors).to.be.eql([]);
-				return expect(status).to.eql(Status.OK);
-			});
+			expect(id).to.be.eql(baseTransaction.id);
+			(errors as ReadonlyArray<TransactionError>).forEach(error =>
+				expect(error).to.be.instanceof(TransactionError),
+			);
+			expect(status).to.eql(Status.FAIL);
```
SargeKhan, line 399: Root level beforeEach defines a baseTransaction which we are not using in these tests, maybe it's better to move the baseTransaction creation in the `#constructor` describe block?

SargeKhan (dismissed):

packages/lisk-transactions/test/utils/get_transaction_id.ts
```diff
@@ -14,39 +14,47 @@
...
+	it('should return a valid id', async () => {
+		expect(getId(Buffer.from(defaultTransactionBytes, 'hex'))).to.be.eql(
 			defaultTransaction.id,
 		);
 	});
+
+	it('should call cryptography hash', async () => {
+		const cryptographyHashStub = sandbox
+			.stub(cryptography, 'hash')
+			.callThrough();
```
shuse2: why do we need `callThrough` here?  call through seems to be used if we have conditional stub, but here it's not using any condition.
same as below.

packages/lisk-transactions/fixtures/transactions.json
```diff
@@ -15,6 +15,7 @@
 		},
 		"signature":
 			"69d0c7bc50b82465e2b0885cebc422aa9cd575050dc89905e22a6e2cc88802935c6809a59a2daa04ca99623a6fef76b7d03215ed7f401b74ef5301b12bfe2002",
+		"signatures": [],
```
shuse2: does signature has to be empty array when instantiating or if it's undefined, it will be an empty array?

packages/lisk-transactions/src/transactions/base.ts
shuse2: this getBasic bytes uses `getAssetBytes` inside?

shuse2 (dismissed):

packages/lisk-transactions/src/transactions/base.ts
mitsuaki-u: yes `getBasicBytes` calls `getAssetBytes` internally

packages/lisk-transactions/test/transactions/base.ts
mitsuaki-u: not in `constructor` as we don't have access to account state. however it is a good point as `isMultisignature` will be checked in `isExpired` to increase time until expiry. currently `verify`, `apply` `undo` and `verifyAgainstTransactions` have access to account state. thoughts? @SargeKhan @shuse2

packages/lisk-transactions/test/transactions/base.ts
mitsujutsu: how so? `incorrectly typed` is better?

packages/lisk-transactions/test/transactions/base.ts
mitsuaki-u: I will create new constants for invalid transaction tests, but will keep root level baseTransaction as it is used again in later parts of the test as well.

packages/lisk-transactions/test/utils/get_transaction_id.ts
mitsuaki-u: using `callThrough` here as a means of not having to return specific bytes just to spy, since we are using stubs and not spies in elements. but I can also return the bytes if it's more clear.

packages/lisk-transactions/src/transactions/base.ts
```diff
@@ -61,7 +61,7 @@ export abstract class BaseTransaction {
 	public readonly type: number;
 	public readonly asset: TransactionAsset = {};
 	public readonly receivedAt: Date = new Date();
-	public readonly isMultisignature?: boolean = false;
+	public isMultisignature? = false;
```
shuse2: maybe this is not boolean? it could be unknown, true or false?

packages/lisk-transactions/src/transactions/base.ts
```diff
@@ -106,14 +106,22 @@ export abstract class BaseTransaction {
...
 			asset: this.assetToJSON(),
 			receivedAt: this.receivedAt,
 		};
 
-		return transaction;
+		if (!this.signSignature) {
+			return transaction;
+		}
+
+		const secondSignedTransaction = {
```
shuse2: why not
`signSignature: this.signature`?

packages/lisk-transactions/src/transactions/base.ts
```diff
@@ -332,13 +350,14 @@ export abstract class BaseTransaction {
 			sender.multisignatures.length > 0 &&
 			sender.multimin
 		) {
+			// Ensure boolean is set correctly to calculate valid multisig expiry time
+			this.isMultisignature = true;
 			const transactionBytes = this.signSignature
 				? Buffer.concat([
 						this.getBasicBytes(),
-						this.getAssetBytes(),
 						Buffer.from(this.signature, 'hex'),
```
shuse2, line 355: when converting from hex to buffer, prefer to use cryptography library

shuse2 (dismissed):

packages/lisk-transactions/src/transactions/base.ts
```diff
@@ -376,33 +394,47 @@ export abstract class BaseTransaction {
 	public apply(sender: Account): TransactionResponse {
 		const updatedBalance = new BigNum(sender.balance).sub(this.fee);
 		const updatedAccount = { ...sender, balance: updatedBalance.toString() };
+		const errors = updatedBalance.gte(0)
+			? []
+			: [
+					new TransactionError(
+						`Account does not have enough LSK: ${sender.address}, balance: ${
```
SargeKhan, line 400: Is this a valid error message?

packages/lisk-transactions/src/transactions/base.ts
```diff
@@ -376,33 +394,47 @@ export abstract class BaseTransaction {
...
+					),
+			  ];
 
 		return {
 			id: this.id,
-			status: Status.OK,
-			state: [updatedAccount],
-			errors: [],
+			status: errors.length > 0 ? Status.FAIL : Status.OK,
+			state: errors.length > 0 ? { sender } : { sender: updatedAccount },
```
SargeKhan: We should return the updatedAccount whether irrespective of errors. What do you think @shuse2

packages/lisk-transactions/src/transactions/base.ts
```diff
@@ -376,33 +394,47 @@ export abstract class BaseTransaction {
...
+			? []
+			: [new TransactionError('Invalid balance amount', this.id)];
 
 		return {
 			id: this.id,
-			status: Status.OK,
-			state: [updatedAccount],
-			errors: [],
+			status: errors.length > 0 ? Status.FAIL : Status.OK,
+			state: errors.length > 0 ? { sender } : { sender: updatedAccount },
```
SargeKhan: Same here.

packages/lisk-transactions/src/transactions/base.ts
```diff
@@ -395,33 +394,47 @@ export abstract class BaseTransaction {
...
+					),
+			  ];
 
 		return {
 			id: this.id,
-			status: Status.OK,
-			state: [updatedAccount],
-			errors: [],
+			status: errors.length > 0 ? Status.FAIL : Status.OK,
+			state: errors.length > 0 ? { sender } : { sender: updatedAccount },
```
shuse2: I think u can return the `updatedAccount` in both cases.
Whether to use updatedAccount or not is up to the user

packages/lisk-transactions/src/transactions/base.ts
```diff
@@ -395,33 +394,47 @@ export abstract class BaseTransaction {
...
+			? []
+			: [new TransactionError('Invalid balance amount', this.id)];
 
 		return {
 			id: this.id,
-			status: Status.OK,
-			state: [updatedAccount],
-			errors: [],
+			status: errors.length > 0 ? Status.FAIL : Status.OK,
+			state: errors.length > 0 ? { sender } : { sender: updatedAccount },
```
shuse2: same as above

packages/lisk-transactions/test/helpers/add_fields.ts
```diff
@@ -0,0 +1,7 @@
+export const addFields = (transaction: any) => {
```
shuse2: addFields might be ambiguous...?

shuse2 (dismissed):

shuse2 (approved):

packages/lisk-transactions/src/transactions/base.ts
```diff
@@ -61,7 +72,8 @@ export abstract class BaseTransaction {
 	public readonly type: number;
 	public readonly asset: TransactionAsset = {};
 	public readonly receivedAt: Date = new Date();
-	public readonly isMultisignature?: boolean = false;
+	public readonly containsUniqueData?: boolean;
```
shuse2, line 75: why did this become property instead of abstract func?

mitsuaki-u: discussed with @SargeKhan briefly and decided for all types we know indefinitely if it contains unique data or not based on `type`, therefore no need for func. thoughts?

SargeKhan (approved):
Looks good! Great job!

PR #972, opened 2019-01-07
Use socket cluster types definitions - Closes#969

### LINKED ISSUE #969: Use socket cluster types definitions
### Description

Use socket cluster type definitions which is available for use now at 
[@types/socketcluster](https://www.npmjs.com/package/@types/socketcluster),
[@types/socketcluster-client](https://www.npmjs.com/package/@types/socketcluster-client),
[@types/socketcluster-server](https://www.npmjs.com/package/@types/socketcluster-server)

ishantiw:
### Description

Add socket cluster server and client type definitions from npm.

* The PR resolves #969
* Linting passes
* Tests pass

### DIFF
--- a/packages/lisk-p2p/src/p2p.ts
+++ b/packages/lisk-p2p/src/p2p.ts
@@ -17,7 +17,7 @@ import { EventEmitter } from 'events';
 import http, { Server } from 'http';
 import { platform } from 'os';
 import querystring from 'querystring';
-import { attach, SCServerSocket } from 'socketcluster-server';
+import { attach, SCServer, SCServerSocket } from 'socketcluster-server';
 
 import { Peer, PeerInfo } from './peer';
 
@@ -33,6 +33,12 @@ import {
 	ProtocolPeerList,
 } from './p2p_types';
 
+type SCServerCloseCallback = () => void;
+
+type SCServerUpdated = {
+	readonly close: (sCServerCloseCallback: SCServerCloseCallback) => void;
+} & SCServer;
+
 import { PeerPool } from './peer_pool';
 
 export const EVENT_NEW_INBOUND_PEER = 'newInboundPeer';
@@ -45,7 +51,7 @@ export class P2P extends EventEmitter {
 	private readonly _config: P2PConfig;
 	private readonly _peerPool: PeerPool;
 	private readonly _httpServer: Server;
-	private readonly _scServer: any; // Until we get comeplete definition for SCServer
+	private readonly _scServer: SCServerUpdated; // Until we get comeplete definition for SCServer
 	private readonly _newPeers: Set<PeerInfo>;
 	// TODO ASAP: private readonly _triedPeers: Set<PeerInfo>;
 	private _nodeStatus: P2PNodeStatus;
@@ -64,7 +70,7 @@ export class P2P extends EventEmitter {
 
 		this._peerPool = new PeerPool();
 		this._httpServer = http.createServer();
-		this._scServer = attach(this._httpServer);
+		this._scServer = attach(this._httpServer) as SCServerUpdated;
 	}
 
 	/* tslint:disable:next-line: prefer-function-over-method */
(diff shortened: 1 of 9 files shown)


packages/lisk-p2p/src/p2p_types.ts
```diff
@@ -31,10 +31,7 @@ export interface P2PMessagePacket<T> {
 }
 
 export interface P2PNodeStatus {
-	readonly wsPort: number;
-	readonly os: string;
-	readonly version: string;
-	readonly height: number;
+	readonly [key: string]: string;
```
shuse2: why changing this from strongly typed to key-val?

ishantiw: While creating an outbound socket we are using it as a query [packages/lisk-p2p/src/peer.ts#L71](https://github.com/LiskHQ/lisk-elements/blob/7bf751af4b4eb90d501db4a80d912e96675fae92/packages/lisk-p2p/src/peer.ts#L71).
And the definition of `query` is coming from  [@types/socketcluster-client](https://www.npmjs.com/package/@types/socketcluster-client),
`query?: string | { [key: string]: string };`

shuse2: Probably it's better to check if that's compatible with the current protocol.
Also, please check if every values are `string` it satisfies `[key: string]: string`

ishantiw: Yeah, actually that's why I kept this PR in progress because I wanted to discuss it will @jondubois as well because he wanted to remove `P2PNodeStatus` from here and also from `PeerConfig`. Also, I think we can create a separate type for `query` type and keep the `P2PNodeStatus` as it was before.

shuse2: Maybe, JSON.stringify() will solve the problem? ='(

ishantiw: yes it will solve the problem 👍

packages/lisk-p2p/src/peer.ts
```diff
@@ -150,10 +154,16 @@ export class Peer {
 	}
 
 	public send<T>(packet: P2PMessagePacket<T>): void {
-		this._createOutboundSocket();
-		this._outboundSocket.emit(packet.event, {
-			data: packet.data,
-		});
+		if (this.outboundSocket) {
```
shuse2: Instead of if/else, prefer early return?

packages/lisk-p2p/src/peer.ts
```diff
@@ -176,19 +186,19 @@ export class Peer {
 		return this._nodeStatus;
 	}
 
-	public set inboundSocket(value: any) {
+	public set inboundSocket(value: SCServerSocket | undefined) {
```
shuse2, line 190: why is this possible to insert `undefined`?
same for below

packages/lisk-p2p/src/p2p.ts
```diff
@@ -102,36 +102,42 @@ export class P2P extends EventEmitter {
...
-							height: queryObject.height ? +queryObject.height : 0,
-						});
-						this._peerPool.addPeer(peer);
-						super.emit(EVENT_NEW_INBOUND_PEER, peer);
-						super.emit(EVENT_NEW_PEER, peer);
-						this._newPeers.add(peer.peerInfo);
+			(socket: SCServerSocket): void => {
+				if (socket.request.url) {
+					const queryObject = querystring.parse(socket.request.url);
+					if (
```
shuse2: prefer to make it early return

shuse2 (changes requested):

packages/lisk-p2p/src/peer.ts
ishantiw: Getter and setter need to have the same signature i.e. if the set value is `SCServerSocket` then get should also be `SCServerSocket` and not  `SCServerSocket | undefined` which is not possible as inbound or outbound connection could be `undefined` at some point. 

but this could be avoided if we use a regular function to set the value instead of a setter. If it looks weird then we can change it.

ishantiw: created PR #988  to remove getters and that will solve this issue.

ishantiw:
Addressed

jondubois (approved):

shuse2 (approved):

PR #976, opened 2019-01-07
Add Vote Transaction - Closes #885

### LINKED ISSUE #885: Add VoteTransaction class
### Description

Extend `BaseTransaction` implement all the specific features of vote transactions.

shuse2:
### Description
Add transaction Vote transaction which extends BaseTransaction

### How to test it?
`npm t`

### Review checklist

* The PR resolves #885 
* Linting passes
* Tests pass


packages/lisk-transactions/src/transactions/base.ts
```diff
@@ -369,44 +413,95 @@ export abstract class BaseTransaction {
...
+	}
+
+	public sign(passphrase: string, secondPassphrase?: string): void {
+		this._signature = undefined;
+		this._signSignature = undefined;
+		this._signature = signData(hash(this.getBasicBytes()), passphrase);
+
+		if (secondPassphrase) {
+			this._signSignature = signData(
+				hash(this.getBasicBytes()),
```
mitsuaki-u: need to include signature bytes?

shuse2: yes! it's fixed in this PR https://github.com/LiskHQ/lisk-elements/pull/998

packages/lisk-transactions/src/transactions/3_vote_transaction.ts
```diff
@@ -0,0 +1,423 @@
...
+				format: 'signedPublicKey',
+			},
+		},
+	},
+};
+
+export const voteAssetFormatSchema = {
+	type: 'object',
+	required: ['votes'],
+	properties: {
```
mitsuaki-u, line 76: should it also check `amount` format (make sure it's zero)

packages/lisk-transactions/src/transactions/3_vote_transaction.ts
```diff
@@ -0,0 +1,423 @@
...
+		return {
+			sender,
+			dependentState: {
+				[ENTITY_ACCOUNT]: dependentAccounts,
+			},
+		};
+	}
+
+	public checkSchema(): TransactionResponse {
+		const { status } = super.checkSchema();
```
mitsuaki-u: don't we also need errors from base check?

packages/lisk-transactions/src/transactions/3_vote_transaction.ts
```diff
@@ -0,0 +1,423 @@
...
+			id: this.id,
+			status: errors.length === 0 ? Status.OK : Status.FAIL,
+			errors,
+		};
+	}
+
+	public apply({ sender }: RequiredState): TransactionResponse {
+		const { errors: baseErrors, state } = super.apply({ sender });
+		if (!state) {
+			throw new Error('state is required for applying transaction');
```
mitsuaki-u: not TransactionError with a transaction id? Also prefer Uppercase beginning of message for consistency.

packages/lisk-transactions/src/transactions/3_vote_transaction.ts
shuse2: this part only covers asset, so I will check in validateSchema stage separately

packages/lisk-transactions/src/transactions/3_vote_transaction.ts
shuse2: TransactionError is handleable error (like it could return the message through API) but in this case, this error is more like a developer error, which should never happen. That's why i put it as standard error

packages/lisk-transactions/src/transactions/3_vote_transaction.ts
mitsuaki-u: would it be better to move the existing schema with all field checks in `schema.ts` to here without the `$merge` keyword, and deprecate the unused version in that file. every other class except type `0` will also have 0 amount, and it may be cleaner to validate in one go?

packages/lisk-transactions/src/transactions/3_vote_transaction.ts
```diff
@@ -0,0 +1,459 @@
...
+			sender,
+			dependentState: {
+				[ENTITY_ACCOUNT]: dependentAccounts,
+			},
+		};
+	}
+
+	public validateSchema(): TransactionResponse {
+		const { errors: baseErrors, status } = super.validateSchema();
+		const valid = validator.validate(voteAssetFormatSchema, this.asset);
```
mitsuaki-u, line 278: validate whole schema format here?

packages/lisk-transactions/test/transactions/3_vote_transaction.ts
```diff
@@ -0,0 +1,1047 @@
...
+
+		describe('when the transaction is created with one passphrase and the votes', () => {
+			beforeEach(async () => {
+				result = VoteTransaction.create({ passphrase, votes: defaultUpvotes });
+			});
+
+			it('should create vote transaction ', async () => {
+				expect(result).to.have.property('id');
+				expect(result)
+					.to.have.property('type')
```
mitsuaki-u: can also do `property('type', 3)`

packages/lisk-transactions/src/transactions/3_vote_transaction.ts
shuse2: basic schema is validated at `super.validateSchema` so I think validating the asset only is good here

packages/lisk-transactions/src/transactions/3_vote_transaction.ts
shuse2: Yes that's true, but mu intention was in the future we only need to validate what is included in the asset =0

mitsuaki-u (dismissed):

packages/lisk-transactions/src/transactions/3_vote_transaction.ts
```diff
@@ -0,0 +1,483 @@
...
+	}
+	validatePublicKeys([...votes, ...unvotes]);
+};
+
+export class VoteTransaction extends BaseTransaction {
+	public readonly containsUniqueData = true;
+	public readonly asset: VoteAsset;
+
+	public constructor(tx: TransactionJSON) {
+		super(tx);
```
SargeKhan, line 113: Shouldn't `super` call be wrapped in try/catch block? So, we can also validate the asset body to create a complete list of errors in the transaction.

packages/lisk-transactions/src/transactions/3_vote_transaction.ts
```diff
@@ -0,0 +1,483 @@
...
+
+		if (!passphrase) {
+			return transaction;
+		}
+
+		const {
+			address: senderId,
+			publicKey: senderPublicKey,
+		} = getAddressAndPublicKeyFromPassphrase(passphrase);
+		const transactionWithSenderInfo = {
```
SargeKhan, line 155: Shouldn't this be done in the `createBaseTransaction` function?

packages/lisk-transactions/src/transactions/3_vote_transaction.ts
```diff
@@ -0,0 +1,483 @@
...
+		const errors = sameTypeTransactions.reduce(
+			(previous, tx) => {
+				if (
+					tx.asset.votes
+						.map(vote => vote.substring(1))
+						.some(publicKey => publicKeys.includes(publicKey))
+				) {
+					return [
+						...previous,
+						new TransactionError(
```
SargeKhan, line 228: Here, we should at least specify which vote is conflicting and which transaction is the current transaction conflicting with. In the current implementation, we will get multiple error messages with the same description without any useful information. What do you think?

packages/lisk-transactions/src/transactions/3_vote_transaction.ts
```diff
@@ -0,0 +1,483 @@
...
+
+		return {
+			id: this.id,
+			errors,
+			status: errors.length === 0 ? Status.OK : Status.FAIL,
+		};
+	}
+
+	public processRequiredState(state: EntityMap): RequiredVoteState {
+		const accounts = state[ENTITY_ACCOUNT];
```
SargeKhan, line 252: Why aren't we calling `super` here?

packages/lisk-transactions/src/transactions/3_vote_transaction.ts
```diff
@@ -0,0 +1,483 @@
...
+		const sender = accounts.find(acct => acct.address === this.senderId);
+		if (!sender) {
+			throw new Error('No sender account is found.');
+		}
+		const votes = this.asset.votes.map(vote => vote.substring(1));
+		const dependentAccounts = accounts.filter(acct =>
+			votes.includes(acct.publicKey),
+		);
+		if (votes.length !== dependentAccounts.length) {
+			throw new Error('Not enought accounts in dependent state.');
```
SargeKhan: typo here

packages/lisk-transactions/src/transactions/3_vote_transaction.ts
```diff
@@ -0,0 +1,483 @@
...
+					: Status.FAIL,
+			errors,
+		};
+	}
+
+	public verify({
+		sender,
+		dependentState,
+	}: RequiredVoteState): TransactionResponse {
+		const { errors: baseErrors } = super.apply({ sender });
```
SargeKhan, line 338: verify will be called after `processRequiredState`. So, do we need to check here if the passed arguments match are as expected?

SargeKhan (dismissed):

packages/lisk-transactions/src/transactions/3_vote_transaction.ts
```diff
@@ -0,0 +1,483 @@
...
+					this.id,
+					'.asset',
+				),
+			);
+		}
+
+		if (
+			this.recipientPublicKey === undefined ||
+			this.recipientPublicKey === ''
+		) {
```
mitsuaki-u: can do just `!this.recipientPublicKey`

packages/lisk-transactions/src/transactions/3_vote_transaction.ts
```diff
@@ -0,0 +1,483 @@
...
+		if (this.recipientId === undefined || this.recipientId === '') {
+			errors.push(
+				new TransactionError(
+					'RecipientId must be set for vote transaction',
+					this.id,
+					'.recipientId',
+				),
+			);
+		}
+
```
mitsuaki-u, line 311: might also need to check `recipientId` format here as it's not included in baseSchema.

packages/lisk-transactions/src/transactions/3_vote_transaction.ts
shuse2: we don't have super here `processRequiredState` is abstract, maybe we should add to BaseTx

packages/lisk-transactions/src/transactions/3_vote_transaction.ts
shuse2: typescript wise, it's hard to check without going through the basic check ='(

packages/lisk-transactions/src/transactions/3_vote_transaction.ts
shuse2: it seems like `super` has to be the first sentence, and try breaks the typescript rule...

mitsuaki-u (approved):

packages/lisk-transactions/src/transactions/3_vote_transaction.ts
```diff
@@ -0,0 +1,478 @@
...
+						),
+			  )
+			: [];
+		if (!typeValid) {
+			throw new TransactionMultiError('Invalid field types', tx.id, errors);
+		}
+		this.asset = tx.asset as VoteAsset;
+	}
+
+	public static create(input: CastVoteInput): object {
```
SargeKhan, line 131: shouldn't this function return TransactionJSON?

packages/lisk-transactions/src/transactions/3_vote_transaction.ts
SargeKhan: Is there a reason for why it's done here?

SargeKhan (approved):

packages/lisk-transactions/src/transactions/3_vote_transaction.ts
shuse2: one thing it needed to be done specific for vote transaction is  to add the `recipientId` same as `senderId`, but it has to be done after checking the passphrase

packages/lisk-transactions/src/transactions/3_vote_transaction.ts
shuse2: we can't return TransactionJSON because without passphrase, it's not enough to become a transaction ='(

PR #983, opened 2019-01-08
Update required attributes - Closes #975

### LINKED ISSUE #975: Update getRequiredAttributes function in BaseTransaction
### Description
In order to keep the extensibility, getRequiredAttributes should be updated to return an object
```
{
  'entity name': { // ex) account
    'property': ['values'] // ex) address: ['xxxL', 'yyyL']
  },
}
```

Also, we should add abstract function
```
interface EntityMap { // name to be considered
  'entity name': [Entity],
}
interface RequiredState {
  sender: Account,
  recipient?: Account,
  dependentStates?: EntityMap,
}

abstract processRequiredState(state: EntityMap): RequiredState
throws error on missing state
```

and update `verify`, `apply` and `undo` function to take `RequiredState` as an input.

shuse2:
### What was the problem?
Return type of getRequiredAttributes was not flexible enough, and we needed another function to call for filtering the entity map into usable format for each transaction.

### How did I fix it?
Add EntityMap and RequiredState types and changed the input of apply, undo and verify functions.

### How to test it?
`npm t`

### Review checklist

* The PR resolves #975 
* Linting passes
* Tests pass

### DIFF
--- a/packages/lisk-transactions/src/transactions/base.ts
+++ b/packages/lisk-transactions/src/transactions/base.ts
@@ -54,8 +54,8 @@ export interface TransactionResponse {
 }
 
 export interface Attributes {
-	readonly [attribute: string]: {
-		readonly [property: string]: string[];
+	readonly [entity: string]: {
+		readonly [property: string]: ReadonlyArray<string>;
 	};
 }
 
@@ -66,7 +66,7 @@ export enum MultisignatureStatus {
 }
 
 export interface EntityMap {
-	readonly [name: string]: ReadonlyArray<unknown>;
+	readonly [name: string]: ReadonlyArray<unknown> | undefined;
 }
 
 export interface RequiredState {
--- a/packages/lisk-transactions/src/transactions/base.ts
+++ b/packages/lisk-transactions/src/transactions/base.ts
@@ -71,8 +71,6 @@ export interface EntityMap {
 
 export interface RequiredState {
 	readonly sender: Account;
-	readonly recipient?: Account;
-	readonly dependentState?: EntityMap;
 }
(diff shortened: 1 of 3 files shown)


mitsuaki-u (approved):

packages/lisk-transactions/src/transactions/base.ts
```diff
@@ -57,6 +59,18 @@ export enum MultisignatureStatus {
 	UNKNOWN = 2,
 }
 
+export interface EntityMap {
+	readonly [name: string]: ReadonlyArray<unknown> | undefined;
+}
+
+export interface RequiredState {
```
SargeKhan, line 72: I think rather than exporting this interface, the interface for RequiredState in BaseTransaction class should only include properties required in the BaseTransaction class, whereas every child class should implement its own Interface dependent on the properties required for that transaction type. This will allow classes to have a strict type for parameters passed to the `verify`, `apply` and `undo` functions.
Therefore, in the BaseTransactions class, the RequiredState interface should only include `sender` object because the BaseTransaction class only makes use of `sender` object. In the child classes which require other objects like `recipient` and `dependentState`, they should define their own interface strictly matching the properties.
We will always be able to call `super` function from child classes because the RequiredState interface in the BaseTransaction class will always be less strict.

shuse2: I thought it would be hard to override the interface, but it seems like it works like a charm, so it's fixed to have only required values!

shuse2:
@SargeKhan Addressed

SargeKhan (approved):

PR #984, opened 2019-01-08
Implement handling of pending signatures in verifyMultisignatures

### LINKED ISSUE #982: Implement handling of pending signatures and incoming new signatures to multisignature verification flow
### Description

- [x] When a multisignature transaction is awaiting new signatures, it should return a pending transaction response. 

- [x] Add `addSignature(signature)` method to base transaction

- [x] Add verifyIncomingSignature(account, signature)` method to base transaction

mitsuaki-u:
### Description

Add functionality and refactor base transaction to account for new incoming signatures.

- [x] When a multisignature transaction is awaiting new signatures, it should return a pending transaction response. 

- [x] Add `addVerifiedMultisignature(signature)` 

- [x] Add `processMultisignatures({ sender })` 

- [x] Add `isReady()` 

### How to test it?

npm run test 

### Review checklist

* The PR resolves #982
* Linting passes
* Tests pass


shuse2:
@mitsujutsu pls rebase with vase branch

packages/lisk-transactions/src/transactions/base.ts
```diff
@@ -376,33 +392,118 @@ export abstract class BaseTransaction {
...
+
+		return {
+			id: this.id,
+			status: errors.length > 0 ? Status.FAIL : Status.OK,
+			state: { sender: updatedAccount },
+			errors,
+		};
+	}
+
+	public addSignature(
```
shuse2: Maybe ‘addMultisignature’? It’s kinda ambiguous if it’s normal signature or not

packages/lisk-transactions/src/transactions/base.ts
```diff
@@ -376,33 +392,118 @@ export abstract class BaseTransaction {
...
+				new TransactionError(
+					'Failed to add signature.',
+					this.id,
+					'.signatures',
+				),
+			],
+		};
+	}
+
+	protected verifyMultisignatures(sender: Account): TransactionResponse {
```
shuse2: input should be RequiredStates

packages/lisk-transactions/src/transactions/base.ts
```diff
@@ -376,33 +392,118 @@ export abstract class BaseTransaction {
...
+		const transactionBytes = this.signSignature
```
shuse2, line 464: isn't it suppose to be `this.getBytes()` here?
For signing multi-sig, I think whole transaction should be signed by the co-signers.
please check with core.

packages/lisk-transactions/src/transactions/base.ts
```diff
@@ -376,33 +392,118 @@ export abstract class BaseTransaction {
...
+			this.id,
+		);
 
 		return {
 			id: this.id,
-			status: Status.OK,
-			state: [updatedAccount],
-			errors: [],
+			status:
+				Array.isArray(errors) && errors[0] instanceof TransactionPendingError
```
shuse2: errors.length > 0 would be better because empty errors will be array but errors[0] will throw error

packages/lisk-transactions/src/utils/sign_and_verify.ts
```diff
@@ -71,7 +72,7 @@ export const verifySignature = (
 	};
 };
 
-export const verifyMultisignatures = (
+export const verifySignatures = (
```
shuse2: I think verifyMultiSignatures would be ok here, it's could be confusing which signature it's verifying

packages/lisk-transactions/src/utils/sign_and_verify.ts
```diff
@@ -84,51 +85,74 @@ export const verifyMultisignatures = (
 			errors: [new TransactionError('Sender does not have valid multimin', id)],
 		};
 	}
-	if (!signatures || (signatures && signatures.length < minimumValidations)) {
+	if (!Array.isArray(signatures)) {
```
shuse2: since type of signatures is `ReadonlyArray<string>`, I think it make sense to assume it's an array

shuse2: Also, it applies to the above `minimumValidations`

packages/lisk-transactions/src/utils/validation/validator.ts
```diff
@@ -29,7 +29,7 @@ import {
 export const validator = new Ajv({ allErrors: true });
 addKeywords(validator);
 
-validator.addFormat('signature', data => /^[a-f0-9]{128}$/i.test(data));
+validator.addFormat('signature',data =>  /^[a-f0-9]{128}$/i.test(data));
```
shuse2: wrong spacing?

packages/lisk-transactions/src/utils/sign_and_verify.ts
```diff
@@ -84,51 +85,74 @@ export const verifyMultisignatures = (
...
+					id,
+					'.signatures',
+				),
+			],
+		};
+	}
 
-	const verifiedSignatures: string[] = [];
-	const checkedPublicKeys: string[] = [];
+	const checkedPublicKeys = new Map();
```
shuse2: why do we need to use `Map` here? also, it's looping thorough the memerPublicKeys once, so im not sure if it's needed

packages/lisk-transactions/src/utils/sign_and_verify.ts
```diff
@@ -84,51 +85,74 @@ export const verifyMultisignatures = (
...
+	const verifiedSignatures = new Map();
```
shuse2: Probably `Set` would be better here

packages/lisk-transactions/src/utils/sign_and_verify.ts
```diff
@@ -84,51 +85,74 @@ export const verifyMultisignatures = (
...
 		});
 	});
 
-	const unverifiedSignatures = signatures.filter(
-		signature =>
-			!verifiedSignatures.find(
-				verifiedSignature => signature === verifiedSignature,
-			),
+	const invalidSignatures = signatures.filter(
+		signature => !verifiedSignatures.has(signature),
```
shuse2, line 124: having invalid signature should be validation error?

packages/lisk-transactions/src/utils/sign_and_verify.ts
shuse2: if it's needed, then `Set` would be better

packages/lisk-transactions/src/utils/sign_and_verify.ts
```diff
@@ -84,51 +85,74 @@ export const verifyMultisignatures = (
...
+				new TransactionPendingError(`Missing signatures`, id, '.signatures'),
+			],
+		};
+	}
+
 	return {
 		verified:
-			verifiedSignatures.length >= minimumValidations &&
-			unverifiedSignatures.length === 0,
+			verifiedSignatures.size >= minimumValidations &&
```
shuse2, line 139: Assuming there is no invalidSignature` at this point, # of `verifiedSignature` should equal `signatures here.
so probably checking `signatures.length >= minimumValidations` is sufficient?

shuse2 (dismissed):
I will review again after rebase

packages/lisk-transactions/src/transactions/base.ts
mitsuaki-u: I don't think so? In core `transaction.js` line 604 it is using the same `verifySignature` for multisignatures as it does for single signature. core `verifySignature` skips `signature` and `secondsignature`  bytes but includes `asset` bytes. this is what is happening here with getBasicBytes usage.

packages/lisk-transactions/src/utils/sign_and_verify.ts
mitsuaki-u: protected class method already uses `verifyMultisignatures`. how about changing this protected class method to `processMultisignatures` and util function will be `verifyMultisignatures`?

shuse2: sounds good to me! @SargeKhan is it ok from ur side?

packages/lisk-transactions/src/utils/sign_and_verify.ts
mitsuaki-u: We need `checkedPublicKeys` due to [cryptographic issue](https://github.com/LiskHQ/lisk/issues/2540) where multiple signatures can be created from the same key. For example, if one key verifies a signature, then it is possible for it to verify another signature as well. Having `checkedKeys` will avoid single key from verifying multiple signature. I will add a comment indicating the reason for clarity.  `Set/Map` makes sense over `array` as we are trying to avoid duplicates, and `has` is more efficient vs `array.includes`.

packages/lisk-transactions/src/utils/sign_and_verify.ts
mitsuaki-u: `invalid` here means it wasn't `verified`, not that formatting was incorrect. i can change `const` name to `unverifiedSignatures`.

packages/lisk-transactions/src/utils/sign_and_verify.ts
shuse2: Instead of using `Map` I think `Set` should be fine.
because we are looping through public keys once, 
`memberPublicKeys.forEach(publicKey => {`
there is no possibility of checking the same public key twice with 2 different public key?

packages/lisk-transactions/src/transaction_types.ts
```diff
@@ -54,7 +49,7 @@ export interface TransactionJSON {
 	readonly senderId?: string;
 	readonly senderPublicKey: string;
 	readonly signature?: string;
-	readonly signatures: ReadonlyArray<string>;
+	readonly signatures: string[];
```
shuse2: Do we need to change the JSON format type?

packages/lisk-transactions/src/transactions/base.ts
```diff
@@ -345,42 +356,34 @@ export abstract class BaseTransaction {
...
 				multisignatureErrors.length > 0
 			) {
-				multisignatureErrors.forEach(error => {
-					errors.push(error);
-				});
+				multisignatureErrors.forEach(error => errors.push(error));
 			}
 		} else {
-			this.isMultisignature = MultisignatureStatus.FALSE;
+			this._multisignatureStatus = MultisignatureStatus.NONMULTISIGNATURE;
```
shuse2: instead of `if/else` can we change it to early return like if it's not multisignature account, set to nonmultisignature and return?

packages/lisk-transactions/src/transactions/base.ts
```diff
@@ -345,42 +356,34 @@ export abstract class BaseTransaction {
...
-			} = verifyMultisignatures(
-				sender.multisignatures,
-				this.signatures,
-				sender.multimin,
-				transactionBytes,
-				this.id,
-			);
+			} = this.processMultisignatures({ sender });
+
+			if (status === Status.PENDING && errors.length === 0) {
```
shuse2: shouldn't pending have `error`?

packages/lisk-transactions/src/transactions/base.ts
```diff
@@ -428,11 +431,97 @@ export abstract class BaseTransaction {
 		};
 	}
 
+	public addMultisignature(
+		signature: string,
+		publicKey: string,
+	): TransactionResponse {
+		const { valid } = validateSignatureAndPublicKey(signature, publicKey);
+
```
shuse2: it might be fine like this, but don't we need to check if corresponding public key already has a valid signature in the `signatures`?

packages/lisk-transactions/src/transactions/base.ts
```diff
@@ -428,11 +431,97 @@ export abstract class BaseTransaction {
...
+		sender,
+	}: RequiredState): TransactionResponse {
+		const transactionBytes = this.signSignature
+			? Buffer.concat([
+					this.getBasicBytes(),
+					Buffer.from(this.signature, 'hex'),
+			  ])
+			: this.getBasicBytes();
+
+		if (!sender.multimin || typeof sender.multimin !== 'number') {
```
shuse2: probaly we don't need to check `typeof` multimin

packages/lisk-transactions/src/utils/sign_and_verify.ts
```diff
@@ -72,63 +72,59 @@ export const verifySignature = (
...
-		}
+	const checkedPublicKeys = new Set();
+	const verifiedSignatures = new Set();
 
+	publicKeys.forEach(publicKey => {
 		signatures.forEach((signature: string) => {
-			if (verifiedSignatures.includes(signature)) {
+			// Avoid single key from verifying more than one signature. See core issue #2540
+			if (
+				checkedPublicKeys.has(publicKey) &&
```
shuse2: shouldn't this be `||`?

packages/lisk-transactions/src/utils/validation/validate_signature_and_publickey.ts
```diff
@@ -0,0 +1,48 @@
...
+ * Removal or modification of this copyright notice is prohibited.
+ *
+ */
+import { validatePublicKey } from '.';
+import { TransactionError } from '../../errors';
+import { IsValidResponse } from '../../transaction_types';
+
+export const validateSignatureAndPublicKey = (
+	signature: string,
+	publicKey?: string,
```
shuse2: does this need to be optional?

packages/lisk-transactions/src/utils/validation/validate_signature_and_publickey.ts
```diff
@@ -0,0 +1,48 @@
...
+	id?: string,
+): IsValidResponse => {
+	const errors: TransactionError[] = [];
+
+	if (!/^[a-f0-9]{128}$/i.test(signature)) {
```
shuse2: don't we have this check in other file?

packages/lisk-transactions/src/utils/sign_and_verify.ts
```diff
@@ -72,63 +72,59 @@ export const verifySignature = (
...
-		if (checkedPublicKeys.includes(publicKey)) {
-			return;
-		}
+	const checkedPublicKeys = new Set();
+	const verifiedSignatures = new Set();
 
+	publicKeys.forEach(publicKey => {
 		signatures.forEach((signature: string) => {
-			if (verifiedSignatures.includes(signature)) {
+			// Avoid single key from verifying more than one signature. See core issue #2540
```
SargeKhan: Update the comment to explain that the issue #2540 is on lisk repository.

packages/lisk-transactions/src/utils/sign_and_verify.ts
SargeKhan: is `duplicatedSignature` correct name for the variable?

SargeKhan (dismissed):

shuse2 (dismissed):

packages/lisk-transactions/src/transaction_types.ts
mitsuaki-u: base constructor rawTransaction type is TransactionJSON, however I can cast it as `		this.signatures = rawTransaction.signatures as string[] || []` if you prefer.

packages/lisk-transactions/src/transactions/base.ts
mitsuaki-u: that's the non-multisignature `errors` which I think should be empty as if there are any previous errors it should not go to pending queue. however below in the return I will include the pending error.

packages/lisk-transactions/src/utils/sign_and_verify.ts
mitsujutsu: this needs to be `&&` otherwise it will reject valid transactions.

packages/lisk-transactions/src/transactions/base.ts
```diff
@@ -358,32 +357,39 @@ export abstract class BaseTransaction {
 		}
 		// Verify multisignatures
 		if (
-			Array.isArray(sender.multisignatures) &&
-			sender.multisignatures.length > 0 &&
-			sender.multimin
+			!(
+				Array.isArray(sender.multisignatures) &&
+				sender.multisignatures.length > 0 &&
```
shuse2, line 362: `sender.multisignatures.length > 0` is true, we can assume it's an array?

packages/lisk-transactions/src/transactions/base.ts
```diff
@@ -358,32 +357,39 @@ export abstract class BaseTransaction {
...
-			if (
-				Array.isArray(multisignatureErrors) &&
-				multisignatureErrors.length > 0
-			) {
-				multisignatureErrors.forEach(error => errors.push(error));
-			}
-		} else {
-			this._multisignatureStatus = MultisignatureStatus.NONMULTISIGNATURE;
+		if (
+			Array.isArray(multisignatureErrors) &&
```
shuse2: same as above

packages/lisk-transactions/src/utils/sign_and_verify.ts
```diff
@@ -83,35 +83,35 @@ export const verifyMultisignatures = (
 
 	publicKeys.forEach(publicKey => {
 		signatures.forEach((signature: string) => {
-			// Avoid single key from verifying more than one signature. See core issue #2540
+			// Avoid single key from verifying more than one signature.
+			// See issue #2540 at https://github.com/LiskHQ/lisk
```
shuse2: it will be even better if we can directly put link for the issue?

packages/lisk-transactions/src/utils/validation/validate_signature_and_publickey.ts
```diff
@@ -12,34 +12,34 @@
...
-                    `Public key ${publicKey} length differs from the expected 32 bytes for a public key.`,
-                    id,
-                    '.publicKey',
-                ),
-            );
-        }
-    }
+	}
+
+	if (publicKey) {
```
shuse2: we don't need this if statement?

shuse2 (changes requested):
mostly looks good, some minor syntax

packages/lisk-transactions/src/transactions/base.ts
mitsuaki-u: gives `possibly undefined` without any check. I can check if `sender.multisignatures` exists instead.

shuse2 (approved):

SargeKhan (approved):

Issue #953, opened 2018-12-20
Add integration tests for discovery for lisk-p2p


jondubois:
### Expected behavior

Need integration tests to check that nodes can connect to each other and that the discovery process works.

### Actual behavior

There are no integration tests.

Issue #897, opened 2018-11-22
Add send/request data mechanism


shuse2:
### Description
lisk-p2p needs to implement send function where it sends data to peers without waiting for the result.

Also, peer object will have the same interface, where internally it can be used to send/request data to particular peer.

jondubois:
This was implemented but only at the peer level in https://github.com/LiskHQ/lisk-elements/pull/936 

- Request: https://github.com/LiskHQ/lisk-elements/pull/936/files#diff-6ba9a8a9b5823f7d6ac59dfaccd16accR91
- Send: https://github.com/LiskHQ/lisk-elements/pull/936/files#diff-6ba9a8a9b5823f7d6ac59dfaccd16accR130

Still needs to be wired up to the main `p2p.request` and `p2p.send` methods.

Issue #896, opened 2018-11-22
Add receive data and handle response mechanism


shuse2:
### Description
lisk-p2p needs to implement response handling mechanism, similar to api handling.

internal RPC call, status and peerList should be able to handled internally, but also library user should be able to handle and respond to the request that they received.

```
p2p.on('requestReceived', async (request) => {
  // ...
  request.end(result);
  // ...
  request.error(customError);
});

p2p.on('messageReceived', async (messagePacket) => {
  // ...
});
```

PR #986, opened 2019-01-08
Implement P2P.getNetworkStatus() functionality - Closes #985 and #897, addresses #896 and #953

### LINKED ISSUE #985: Implement P2P.getNetworkStatus() functionality
### Expected behavior

The P2P library should implement a `getNetworkStatus` method (maybe as a getter?) to allow the user of the library to get information about the network's state. At the beginning, it should return an object which has at least an `activePeers` property and a `newPeers` property. A `triedPeers` property can be added later.

### Actual behavior

The `getNetworkStatus` has not bee implemented.

jondubois:
### What was the problem?

In order to implement the integration tests, there needs to be a way to get information about the network's state from the P2P library. This feature depends on the `P2P.getNetworkStatus()` method of the P2P library which itself depends on the peer request/send functionality (to allow peers to send us information about themselves).

### How did I fix it?

- Implemented outbound request/send functionality (at the `Peer` level only)
- Implemented inbound request/message handling (at the `Peer` level only)
- Implemented sending of `P2PNodeInfo` to remote peers.
- Implemented receiving `PeerInfo` from remote peers.
- Implemented `getNetworkStatus()` method on `P2P` class.

### How to test it?

Will add integration tests for all these features as part of #953 then, once passing, the logic can be locked down with unit tests.

### Review checklist

* The PR resolves #985 and #897 
* The PR addresses #896 and #953
* Linting passes
* Tests pass

### DIFF
--- a/packages/lisk-p2p/src/p2p.ts
+++ b/packages/lisk-p2p/src/p2p.ts
@@ -22,10 +22,10 @@ import { attach, SCServer, SCServerSocket } from 'socketcluster-server';
 import { Peer, PeerInfo } from './peer';
 
 import {
-	// TODO ASAP: NetworkStatus,
 	P2PConfig,
 	P2PMessagePacket,
-	P2PNodeStatus,
+	P2PNetworkStatus,
+	P2PNodeInfo,
 	P2PPenalty,
 	P2PRequestPacket,
 	P2PResponsePacket,
@@ -49,22 +49,25 @@ const BASE_10_RADIX = 10;
 
 export class P2P extends EventEmitter {
 	private readonly _config: P2PConfig;
-	private readonly _peerPool: PeerPool;
 	private readonly _httpServer: Server;
-	private readonly _scServer: SCServerUpdated; // Until we get comeplete definition for SCServer
+	private _isActive: boolean;
 	private readonly _newPeers: Set<PeerInfo>;
-	// TODO ASAP: private readonly _triedPeers: Set<PeerInfo>;
-	private _nodeStatus: P2PNodeStatus;
+
+	// TODO ASAP: private readonly _triedPeers: Set<PeerConfig>;
+	private _nodeInfo: P2PNodeInfo;
+	private readonly _peerPool: PeerPool;
+	private readonly _scServer: SCServerUpdated;
 
 	public constructor(config: P2PConfig) {
 		super();
 		this._config = config;
-		this._nodeStatus = {
+		this._nodeInfo = {
 			wsPort: config.wsPort,
 			os: platform(),
 			version: config.version,
 			height: 0,
 		};
+		this._isActive = false;
 		this._newPeers = new Set();
 		// TODO ASAP: this._triedPeers = new Set();
 
@@ -73,16 +76,43 @@ export class P2P extends EventEmitter {
 		this._scServer = attach(this._httpServer) as SCServerUpdated;
 	}
 
+	public get isActive(): boolean {
+		return this._isActive;
+	}
+
+	public get nodeInfo(): P2PNodeInfo {
+		return this._nodeInfo;
+	}
+
+	/**
+	 * This is not a declared as a setter because this method will need
+	 * invoke an async RPC on Peers to give them our new node status.
+	 */
+	public applyNodeInfo(nodeInfo: P2PNodeInfo): void {
+		this._nodeInfo = nodeInfo;
+		const peerList = this._peerPool.getAllPeers();
+		peerList.forEach(peer => {
+			peer.applyNodeInfo(nodeInfo);
+		});
+	}
+
 	/* tslint:disable:next-line: prefer-function-over-method */
 	public applyPenalty(penalty: P2PPenalty): void {
 		penalty;
 	}
 
-	// TODO ASAP: public getNetworkStatus(): NetworkStatus {};
+	public getNetworkStatus(): P2PNetworkStatus {
+		// TODO ASAP: Add additional properties.
+		return {
+			peers: this._peerPool.getAllPeerInfos(),
+		};
+	}
+
 	/* tslint:disable:next-line: prefer-function-over-method */
 	public async request<T>(
 		packet: P2PRequestPacket<T>,
 	): Promise<P2PResponsePacket> {
+		// TODO ASAP
 		return Promise.resolve({ data: packet });
 	}
 
@@ -90,19 +120,6 @@ export class P2P extends EventEmitter {
 	public send<T>(message: P2PMessagePacket<T>): void {
 		message;
 		// TODO ASAP
-	}
-
-	public get nodeStatus(): P2PNodeStatus {
-		return this._nodeStatus;
-	}
-
-	/**
-	 * This is not a declared as a setter because this method will need
-	 * invoke an async RPC on Peers to give them our new node status.
-	 */
-	public applyNodeStatus(value: P2PNodeStatus): void {
-		this._nodeStatus = value;
-		// TODO ASAP: Pass to PeerPool -> connected Peer objects
 	}
 
 	private async _startPeerServer(): Promise<void> {
@@ -126,17 +143,14 @@ export class P2P extends EventEmitter {
 					const peerId = Peer.constructPeerId(socket.remoteAddress, wsPort);
 					const existingPeer = this._peerPool.getPeer(peerId);
 					if (existingPeer === undefined) {
-						const peer = new Peer(
-							{
-								ipAddress: socket.remoteAddress,
-								os: queryObject.os,
-								version: queryObject.version,
-								wsPort,
-								nodeStatus: this._nodeStatus,
-								height: queryObject.height ? +queryObject.height : 0,
-							},
-							socket,
-						);
+						const peer = new Peer({
+							inboundSocket: socket,
+							ipAddress: socket.remoteAddress,
+							os: queryObject.os,
+							version: queryObject.version,
+							wsPort,
+						});
+						peer.applyNodeInfo(this._nodeInfo);
 						this._peerPool.addPeer(peer);
 						super.emit(EVENT_NEW_INBOUND_PEER, peer);
 						super.emit(EVENT_NEW_PEER, peer);
@@ -152,6 +166,7 @@ export class P2P extends EventEmitter {
 
 		return new Promise<void>(resolve => {
 			this._scServer.once('ready', () => {
+				this._isActive = true;
 				resolve();
 			});
 		});
@@ -161,48 +176,46 @@ export class P2P extends EventEmitter {
 		// TODO ASAP: Test this and check for potential failure scenarios.
 		return new Promise<void>(resolve => {
 			this._scServer.close(() => {
+				this._isActive = false;
 				resolve();
 			});
 		});
 	}
 
+	// TODO ASAP: This should be refactored to use the discovery function.
 	private async _loadListOfPeerListsFromSeeds(
 		seedList: ReadonlyArray<PeerInfo>,
 	): Promise<ReadonlyArray<ReadonlyArray<Peer>>> {
 		return Promise.all(
-			seedList.map(async (seedPeer: PeerInfo) => {
-				const peer = new Peer({
-					ipAddress: seedPeer.ipAddress,
-					wsPort: seedPeer.wsPort,
-					nodeStatus: this._nodeStatus,
-					height: seedPeer.height,
-					os: seedPeer.os,
-					version: seedPeer.version,
-				});
-				this._newPeers.add(peer.peerInfo);
+			seedList.map(async (seedPeerInfo: PeerInfo) => {
+				const seedPeer = new Peer(seedPeerInfo);
+				seedPeer.applyNodeInfo(this._nodeInfo);
+				this._newPeers.add(seedPeer.peerInfo);
 				/**
-				 * TODO LATER: For the LIP phase, we shouldn't add the seed peers to our
+				 * TODO later: For the LIP phase, we shouldn't add the seed peers to our
 				 * _peerPool and we should disconnect from them as soon as we've loaded their peer lists.
 				 */
-				this._peerPool.addPeer(peer);
-				// TODO LATER: For the LIP phase, the 'list' request will need to be renamed.
-				const seedNodePeerListResponse = await peer.request<void>({
+				this._peerPool.addPeer(seedPeer);
+				// TODO later: For the LIP phase, the 'list' request will need to be renamed.
+				const seedNodePeerListResponse = await seedPeer.request<void>({
 					procedure: 'list',
 				});
 				const peerListResponse = seedNodePeerListResponse.data as ProtocolPeerList;
+				
 				// TODO ASAP: Validate the response before returning. Check that seedNodePeerListResponse.data.peers exists.
 
-				return peerListResponse.peers.map(
-					(peerObject: ProtocolPeerInfo) =>
-						new Peer({
-							ipAddress: peerObject.ip,
-							wsPort: peerObject.wsPort, // TODO ASAP: Add more properties
-							nodeStatus: this._nodeStatus,
-							height: seedPeer.height,
-							os: seedPeer.os,
-							version: seedPeer.version,
-						}),
-				);
+				return peerListResponse.peers.map((peerInfo: ProtocolPeerInfo) => {
+					const peer = new Peer({
+						ipAddress: peerInfo.ip,
+						wsPort: peerInfo.wsPort, // TODO ASAP: Add more properties
+						height: peerInfo.height,
+						os: peerInfo.os,
+						version: peerInfo.version,
+					});
+					peer.applyNodeInfo(this._nodeInfo);
+
+					return peer;
+				});
 			}),
 		);
 	}
--- a/packages/lisk-p2p/src/p2p.ts
+++ b/packages/lisk-p2p/src/p2p.ts
@@ -52,8 +52,8 @@ export class P2P extends EventEmitter {
 	private readonly _httpServer: Server;
 	private _isActive: boolean;
 	private readonly _newPeers: Set<PeerInfo>;
+	private readonly _triedPeers: Set<PeerInfo>;
 
-	// TODO ASAP: private readonly _triedPeers: Set<PeerInfo>;
 	private _nodeInfo: P2PNodeInfo;
 	private readonly _peerPool: PeerPool;
 	private readonly _scServer: SCServerUpdated;
@@ -69,7 +69,7 @@ export class P2P extends EventEmitter {
 		};
 		this._isActive = false;
 		this._newPeers = new Set();
-		// TODO ASAP: this._triedPeers = new Set();
+		this._triedPeers = new Set();
 
 		this._peerPool = new PeerPool();
 		this._httpServer = http.createServer();
@@ -102,10 +102,14 @@ export class P2P extends EventEmitter {
 	}
 
 	public getNetworkStatus(): P2PNetworkStatus {
-		// TODO ASAP: Add additional properties such as _triedPeers.
+		const newPeers = [...this._newPeers.values()];
+		const triedPeers = [...this._triedPeers.values()];
+		const availablePeers = newPeers.concat(triedPeers);
 		return {
-			newPeers: [...this._newPeers.values()],
-			activePeers: this._peerPool.getAllPeerInfos(),
+			newPeers,
+			triedPeers,
+			availablePeers,
+			connectedPeers: this._peerPool.getAllPeerInfos(),
 		};
 	}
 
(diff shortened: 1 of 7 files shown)


packages/lisk-p2p/src/peer.ts
```diff
@@ -44,34 +58,144 @@ export interface PeerConnectionState {
...
+			this._startListeningForRPCs(this._inboundSocket);
+			this._startListeningForMessages(this._inboundSocket);
+		}
+		this._height = peerInfo.height ? peerInfo.height : 0;
+
+		// This needs to be an arrow function so that it can be used as a listener.
+		this._handleRPC = (packet: unknown) => {
+			// TODO later: Switch to LIP protocol format.
+			// TODO ASAP: Move validation/sanitization to a separate file with other validation logic.
+			const request = packet as ProtocolInboundRPCRequest;
```
shuse2: if we use `as` here, there is no point to use `unknown` as input.
maybe add validation like
```
const isInboundRPCRequest(packet: unknown): packet is ProtocolInboundRPCRequest {}
```

packages/lisk-p2p/src/peer.ts
```diff
@@ -44,34 +58,144 @@ export interface PeerConnectionState {
...
+			if (!request || typeof request.procedure !== 'string') {
+				this.emit(EVENT_INVALID_REQUEST_RECEIVED, request);
+
+				return;
+			}
+			if (
+				request.procedure === 'updateMyself' &&
```
shuse2: prefer to make it as a constant for known `procedure` name.

packages/lisk-p2p/src/peer.ts
```diff
@@ -44,34 +58,144 @@ export interface PeerConnectionState {
...
+			}
+			// Emit request for external use.
+			this.emit(EVENT_REQUEST_RECEIVED, request);
+		};
+
+		// This needs to be an arrow function so that it can be used as a listener.
+		this._handleMessage = (packet: unknown) => {
+			// TODO later: Switch to LIP protocol format.
+			// TODO ASAP: Move validation/sanitization to a separate file with other validation logic.
+			const message = packet as ProtocolInboundMessage;
```
shuse2: same as above, if we use `as` here there is no point to use `unknown`

packages/lisk-p2p/src/peer.ts
```diff
@@ -44,34 +58,144 @@ export interface PeerConnectionState {
...
+		if (this._inboundSocket) {
+			this._stopListeningForRPCs(this._inboundSocket);
+			this._stopListeningForMessages(this._inboundSocket);
 		}
+		this._inboundSocket = value;
+		this._startListeningForRPCs(this._inboundSocket);
+		this._startListeningForMessages(this._inboundSocket);
+	}
+
+	public get inboundSocket(): any {
```
shuse2: do we need to add getter for socket? prefer to be enclosed

packages/lisk-p2p/src/peer.ts
```diff
@@ -44,34 +58,144 @@ export interface PeerConnectionState {
...
+	public get nodeInfo(): P2PNodeInfo | undefined {
+		return this._nodeInfo;
+	}
+
+	public set outboundSocket(value: any) {
+		this._outboundSocket = value;
+	}
+
+	public get outboundSocket(): any {
+		return this._outboundSocket;
```
shuse2: same as `inboundSocket`

packages/lisk-p2p/src/peer.ts
```diff
@@ -44,34 +58,144 @@ export interface PeerConnectionState {
...
+	public static constructPeerId(ipAddress: string, port: number): string {
+		return `${ipAddress}:${port}`;
+	}
+
+	/**
+	 * This is not a declared as a setter because this method will need
+	 * invoke an async RPC on the socket to pass it the new node status.
+	 */
+	public applyNodeInfo(nodeInfo: P2PNodeInfo): void {
+		this._nodeInfo = nodeInfo;
```
shuse2, line 185: do we need to maintain `nodeInfo` per Peer?

packages/lisk-p2p/src/peer.ts
```diff
@@ -134,73 +259,40 @@ export class Peer {
...
-	}
-
-	public get inboundSocket(): any {
-		return this._inboundSocket;
-	}
-
-	public set outboundSocket(value: any) {
-		this._outboundSocket = value;
+	private _createOutboundSocket(): void {
+		if (!this._outboundSocket) {
```
shuse2: prefer early return.
```
if (this._outboundSocket) {
  return;
}
```
Also, should this be fail safe to be called twice, or just throw an error instead of making it safe

shuse2 (changes requested):

packages/lisk-p2p/src/peer.ts
jondubois: Yes because we need to pass it as a query parameter as part of the WebSocket connection (required by current protocol).

packages/lisk-p2p/src/peer.ts
jondubois: I turned this into a pure function so that should solve this issue.

packages/lisk-p2p/src/peer.ts
jondubois: There will be an additional function call (probably imported from `response_handler_sanitization.ts`) before the `const request = packet as ProtocolInboundRPCRequest;` line which will do sanitization on the unknown object. 

Because the packet comes directly from the network, we can't cast it directly to `ProtocolInboundRPCRequest` before doing some sanitization on it.

In any case, the `as ProtocolInboundRPCRequest` statement will probably be moved to a sanitization function in `response_handler_sanitization.ts`.

packages/lisk-p2p/src/p2p_types.ts
```diff
@@ -44,28 +44,45 @@ export interface P2PConfig {
 	readonly connectTimeout: number;
 	readonly ipAddress?: string;
```
ishantiw: Why ipAddress is optional here?

packages/lisk-p2p/src/p2p_types.ts
```diff
@@ -44,28 +44,45 @@ export interface P2PConfig {
...
 	readonly seedPeers: ReadonlyArray<PeerInfo>;
+	readonly version: string;
 	readonly wsEngine?: string;
 	readonly wsPort: number;
-	readonly version: string;
 }
 
-export interface NetworkStatus {}
+// Network info exposed by the P2P library.
+export interface P2PNetworkStatus {
```
ishantiw, line 53: since we have a concept of `newPeers` and `triedPeers`, maybe for clarity we should rename `newPeers` as `availablePeers` or `unConnectedPeers` at the network level. 

Also, `triedPeers` can contain peers that are connected or disconnected. So it should also be checked as well for peers that are not connected
```typescript
export interface P2PNetworkStatus {
    connectedPeers: ReadonlyArray<PeerInfo>; //  peers from peer pool
    unConnectedPeers: ReadonlyArray<PeerInfo>; // or availablePeers, peers from newPeers + triedPeers (check with peerPool)
}
```

packages/lisk-p2p/src/peer.ts
```diff
@@ -13,26 +13,42 @@
...
 export interface PeerInfo {
 	readonly ipAddress: string;
 	readonly wsPort: number;
-	readonly nodeStatus?: P2PNodeStatus; // TODO DELEEETETETE
 	readonly clock?: Date;
 	readonly height: number;
-	readonly os: string;
-	readonly version: string;
+	readonly os?: string;
+	readonly version?: string;
```
ishantiw, line 50: I thought we were making `version` compulsory. If we don't pass`version` value then at sanitization it will always throw an error

packages/lisk-p2p/src/peer.ts
```diff
@@ -153,79 +278,38 @@ export class Peer {
...
 
-	public get peerInfo(): PeerInfo {
-		return this._peerInfo;
-	}
-
-	public get id(): string {
-		return this._id;
+	private _handlePeerInfo(request: ProtocolInboundRPCRequest): void {
+		// Update peerInfo with the latest values from the remote peer.
+		Object.assign(this._peerInfo, request.data as PeerInfo);
```
ishantiw: I think we are missing sanitization of the peer object from the response. we can wrap `request.data` with `instantiatePeerFromResponse()`

packages/lisk-p2p/src/peer.ts
ishantiw: Maybe it will be better to take packet as `unknown` and pass it to `response_handler_sanitization.ts`. We can add a new handler for this in `response_handler_sanitization.ts`

packages/lisk-p2p/src/peer.ts
```diff
@@ -46,55 +62,164 @@ export interface PeerConnectionState {
...
+			this._inboundSocket.disconnect(code, reason);
+			this._stopListeningForRPCs(this._inboundSocket);
 		}
 		if (this.outboundSocket) {
 			this.outboundSocket.disconnect(code, reason);
 		}
 	}
 
+	public send<T>(packet: P2PMessagePacket<T>): void {
+		if (!this.outboundSocket) {
```
ishantiw: I think this will fail as we don't have a getter for `outboundSocket` anymore. Instead, we can use `this._outboundSocket`

packages/lisk-p2p/src/peer.ts
```diff
@@ -46,55 +62,164 @@ export interface PeerConnectionState {
...
+			this.outboundSocket = this._createOutboundSocket();
+		}
+		this.outboundSocket.emit(packet.event, {
```
ishantiw: same here, `this._outboundSocket`

packages/lisk-p2p/src/peer.ts
```diff
@@ -46,55 +62,164 @@ export interface PeerConnectionState {
...
+		this._inboundSocket = value;
+		this._startListeningForRPCs(this._inboundSocket);
+		this._startListeningForMessages(this._inboundSocket);
+	}
+
+	public get ipAddress(): string {
+		return this._ipAddress;
+	}
+
+	public set outboundSocket(value: any) {
```
ishantiw: I think this was not rebased properly, instead of `any` it should be `SCServerSocket`

packages/lisk-p2p/src/peer.ts
```diff
@@ -46,55 +62,164 @@ export interface PeerConnectionState {
...
-		return create({
-			hostname: this._ipAddress,
-			port: this._wsPort,
-			query,
-			autoConnect: false,
+	public get id(): string {
+		return this._id;
+	}
+
+	public set inboundSocket(value: any) {
```
ishantiw: it should be `SCServerSocket` instead of `any`

packages/lisk-p2p/src/peer.ts
```diff
@@ -46,55 +62,164 @@ export interface PeerConnectionState {
...
+			data: nodeInfo,
 		});
 	}
 
+	public get nodeInfo(): P2PNodeInfo | undefined {
+		return this._nodeInfo;
+	}
+
 	public connect(): void {
 		if (!this.outboundSocket) {
```
ishantiw: It should be `this._outboundSocket`

packages/lisk-p2p/src/peer.ts
```diff
@@ -46,55 +62,164 @@ export interface PeerConnectionState {
...
 			this.outboundSocket = this._createOutboundSocket();
 		}
-
 		this.outboundSocket.connect();
```
ishantiw: `this._outboundSocket`

packages/lisk-p2p/src/peer.ts
```diff
@@ -46,55 +62,164 @@ export interface PeerConnectionState {
...
 	}
 
 	public disconnect(code: number = 1000, reason?: string): void {
-		if (this.inboundSocket) {
-			this.inboundSocket.disconnect(code, reason);
+		if (this._inboundSocket) {
+			this._inboundSocket.disconnect(code, reason);
+			this._stopListeningForRPCs(this._inboundSocket);
 		}
 		if (this.outboundSocket) {
```
ishantiw: `this._outboundSocket`

ishantiw (changes requested):

packages/lisk-p2p/src/p2p_types.ts
jondubois: This is the ipAddress of the node itself.
It's this property from the current config.json: https://github.com/LiskHQ/lisk/blob/development/config/default/config.json#L4 - Default is `0.0.0.0`. We can rename this if it's confusing.

packages/lisk-p2p/src/p2p_types.ts
jondubois: unConnectedPeers is slightly different because some peers in the _newPeers map could be connected inbound only; maybe it's OK but slightly more ambiguous. 

I think newPeers is OK. Once we start moving peers into the _triedPeers table, it will make more sense. So there will be newPeers, triedPeers and ~~connectedPeers~~ activePeers. I think we should use the same name to refer to the same thing.

packages/lisk-p2p/src/peer.ts
jondubois: That's a tricky one.

The reason why version is optional is because the list of seedPeers which is fetched from the config file has this format:

`readonly seedPeers: ReadonlyArray<PeerInfo>;`

We cannot know what the version number of the seed peer is before we connect to it.
I think it's OK to expect the `version` in the sanitization logic when checking inbound PeerInfo from a remote peer, but it shouldn't be compulsory (based on our current types) at the type level because the `PeerInfo` object may be created directly in our own code or from a config file before we know anything about the peer (also would be tricky to know their OS before we connect to them).

Not sure if it's better to add yet another `Peer...` type? It's convenient that `PeerInfo` can be used for both getting information about peers and also for creating peers.

packages/lisk-p2p/src/peer.ts
jondubois: I messed up the rebase there.

packages/lisk-p2p/src/p2p_types.ts
ishantiw: maybe we can rename it to `nodeIp` to have a clear differentiation from the peer IP address. What do you think?

packages/lisk-p2p/src/p2p_types.ts
ishantiw: Actually, according to the definition in LIPS, the moment you make a connection with any peer from `newPeer` we remove it from there and add it to `triedPeer`. So, `newPeer` should never have any peer for which we made connection already.  Whereas `triedPeers` will have peers with whom we have a connection or had a connection in the past.

so, active peers should have, triedPeers peers which have a connection already and available peers will be the combination of (newPeers + triedPeers(with no socket))

jondubois: We could have `availablePeers` be `newPeers + triedPeers` and `connectedPeers` be peers from the `PeerPool`?

shuse2: > availablePeers be newPeers + triedPeers and connectedPeers be peers from the PeerPool

I think this is good, im not sure if we need to expose `newPeers` or `triedPeers`

packages/lisk-p2p/src/p2p_types.ts
```diff
@@ -17,55 +17,74 @@
...
 // This is a representation of the peer list according to the current protocol.
 // TODO later: Switch to LIP protocol format.
 export interface ProtocolPeerList {
-	readonly success: boolean;
 	readonly peers: ReadonlyArray<ProtocolPeerInfo>;
+	readonly success: boolean;
+}
+
+// TODO later: Switch to LIP protocol format.
+export interface ProtocolInboundRPCRequest {
```
shuse2: would the request and response format differ in `inbound` or `outbound`?

packages/lisk-p2p/src/peer.ts
shuse2: one idea to have `BasePeerInfo` which is extended to `SeedPeerIndo` and `PeerInfo` where the difference is version?

packages/lisk-p2p/src/peer.ts
```diff
@@ -153,79 +277,44 @@ export class Peer {
...
+			query: JSON.stringify(this._nodeInfo),
+			autoConnect: false,
 		});
 	}
 
-	public get peerInfo(): PeerInfo {
-		return this._peerInfo;
+	private _handlePeerInfo(request: ProtocolInboundRPCRequest): void {
+		// Update peerInfo with the latest values from the remote peer.
+		Object.assign(this._peerInfo, request.data as PeerInfo);
```
shuse2: Also, prefer to use it like
```
this._peerInfo = {
 ...this._peerInfo,
  ...request.data,
}
```

packages/lisk-p2p/src/peer.ts
shuse2: validation needs to be added later here too

packages/lisk-p2p/test/integration/p2p.ts
```diff
@@ -17,24 +18,62 @@ describe.skip('Integration tests for P2P library', () => {
 				});
 			});
 
-			let peerStartPromises: ReadonlyArray<
-				Promise<void>
-			> = blockchainP2PList.map(blockchainP2P => {
-				return blockchainP2P.start();
+			let peerStartPromises: ReadonlyArray<Promise<void>> = p2pNodeList.map(
```
shuse2: this should be `const`?

shuse2 (changes requested):
minor thing and some question/comment

packages/lisk-p2p/src/peer.ts
jondubois: @shuse2 Sounds good but I think that `SeedPeerInfo` might end up having the same properties as `BasePeerInfo` so maybe just `BasePeerInfo` and `PeerInfo`?

Or it could be `PeerInfo` (for constructing) and `DetailedPeerInfo` (when getting information out of the node with `getNetworkStatus()`)?

shuse2: > BasePeerInfo and PeerInfo

sounds good =)

shuse2 (approved):

packages/lisk-p2p/src/p2p_types.ts
ishantiw: Isn't `availablePeers` and `connectedPeers` is enough? I thought we don't want to expose newPeers and triedPeers

packages/lisk-p2p/src/p2p.ts
```diff
@@ -49,40 +49,75 @@ const BASE_10_RADIX = 10;
...
 	/* tslint:disable:next-line: prefer-function-over-method */
 	public applyPenalty(penalty: P2PPenalty): void {
 		penalty;
 	}
 
-	// TODO ASAP: public getNetworkStatus(): NetworkStatus {};
+	public getNetworkStatus(): P2PNetworkStatus {
+		const newPeers = [...this._newPeers.values()];
+		const triedPeers = [...this._triedPeers.values()];
+		const availablePeers = newPeers.concat(triedPeers);
```
ishantiw: is `availablePeers` are all the peers that are available but don't have any connection yet? or, does it include all the peers available or known including the connected peers? If its the former case then I think we need to filter `newPeers` and `triedPeers` to remove peers that we are already connected to and then concat.

ishantiw (commented):
Some comments related to `getNetworkStatus()`

packages/lisk-p2p/src/p2p.ts
jondubois: Since the `getNetworkStatus` call will be used by integration tests (not just API module), it's important that it shows the internal state of the node as closely as possible.

I think it's more important that this interface be detailed rather than user-friendly. Users of the method can do their own sanitisation of the raw data.

Integration test cases will want to check the state of the _newPeers and _triedPeers maps explicitly (since it's a core concept in the LIPs), so we need the name of the exposed properties to reflect the name of internal properties for which we want to allow introspection. We could call it `newPeersList` instead of `newPeers` maybe?

The less sanitisation/formatting we do on our end, the more performant this call will be.

jondubois: @shuse2 What is your opinion about this?

shuse2: I think without `availablePeers` is fine.
`availablePeers` seems ambiguous in this case because it will include all the peers regardless of connection since `triedPeer` will include the peer with outbound connection.
Also, not really useful if it's just combination of `newPeers` and `triedPeers`, user can just combine by themselves?

ishantiw: In case we want to have a concatenated list of `newPeers` and `triedPeers` and if we are exposing them then we don't need `availablePeers` list as its redundant or the user can make it out by themselves. I was thinking that network status is just about knowing (connected and unconnected Peers) but since `newPeers` and `triedPeers` will be used in the network test then its a better to add it. 

I guess in future we will need a separate class for handling peer bucket or peer directory (like I discussed once) that includes newPeer/triedPeer/bannedPeer and then it will be easier to test these lists.

ishantiw (approved):

PR #990, opened 2019-01-09
Remove asset field from base transaction - Closes #989

### LINKED ISSUE #989: Remove asset field from base transaction
### Description

In order to make the type more safe, asset should be defined in each custom transactions.

shuse2:
### What was the problem?
In order to make the base transaction more flexible yet type safe, asset should not be defined in the base transaction.

### How did I fix it?
Remove the asset field from the base transaction

### Review checklist

* The PR resolves #989 
* Linting passes
* Tests pass

### DIFF
--- a/packages/lisk-transactions/src/transactions/base.ts
+++ b/packages/lisk-transactions/src/transactions/base.ts
@@ -152,7 +152,7 @@ export abstract class BaseTransaction {
 		};
 
 		return transaction;
-	}	
+	}
 
 	public getBytes(): Buffer {
 		const transactionBytes = Buffer.concat([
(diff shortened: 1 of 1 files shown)


mitsuaki-u (approved):

packages/lisk-transactions/src/transactions/base.ts
```diff
@@ -101,7 +98,7 @@ export abstract class BaseTransaction {
 		this.receivedAt = rawTransaction.receivedAt;
 	}
 
-	public abstract assetToJSON(): TransactionAsset;
+	public abstract assetToJSON(): object;
```
SargeKhan: Why is it no longer of type TransactionAsset?

shuse2: Since asset type is defined in each child transaction, we can't expect it to be specific type.
Maybe we can use generics here, but it's only used in this class, so it wont make much sense?

SargeKhan (approved):

PR #993, opened 2019-01-09
Use peer selection to choose peers for connection in p2p - Closes#978

### LINKED ISSUE #978: Use selectForConnection in P2P
### Description

We have `selectForConnection` function to select peers for connection at P2P level but as of now, we are not using this function. So we need to call this function when the p2p module tries to select peers to connect.

ishantiw:
### Description

Use peer selection for connection method to choose peers to connect in p2p class.

### Review checklist

* The PR resolves #978
* Linting passes
* Tests pass

### DIFF
--- a/packages/lisk-p2p/src/p2p.ts
+++ b/packages/lisk-p2p/src/p2p.ts
@@ -206,17 +206,13 @@ export class P2P extends EventEmitter {
 		});
 	}
 
-	private _getPeerToConnect(): ReadonlyArray<Peer> {
+	private _connectToPeers(): void {
 		const availablePeers = Array.from(this._newPeers).map(
 			(peerInfo: PeerInfo) => new Peer(peerInfo),
 		);
+
 		const peersToConnect = selectForConnection(availablePeers);
-
-		return peersToConnect;
-	}
-
-	private _connectToPeers(): void {
-		this._getPeerToConnect().map((peer: Peer) => {
+		peersToConnect.map((peer: Peer) => {
 			peer.connect();
 			this._newPeers.delete(peer.peerInfo);
 			this._triedPeers.add(peer.peerInfo);
--- a/packages/lisk-p2p/src/p2p.ts
+++ b/packages/lisk-p2p/src/p2p.ts
@@ -212,7 +212,7 @@ export class P2P extends EventEmitter {
 		);
 
 		const peersToConnect = selectForConnection(availablePeers);
-		peersToConnect.map((peer: Peer) => {
+		peersToConnect.forEach((peer: Peer) => {
 			peer.connect();
 			this._newPeers.delete(peer.peerInfo);
 			this._triedPeers.add(peer.peerInfo);
--- a/packages/lisk-p2p/src/p2p.ts
+++ b/packages/lisk-p2p/src/p2p.ts
@@ -20,7 +20,7 @@ import querystring from 'querystring';
 import { attach, SCServer, SCServerSocket } from 'socketcluster-server';
 
 import { RequestFailError } from './errors';
-import { Peer, PeerInfo } from './peer';
+import { ConnectionState, Peer, PeerInfo } from './peer';
 import { discoverPeers } from './peer_discovery';
 import { PeerOptions } from './peer_selection';
 import { selectForConnection } from './peer_selection';
@@ -180,8 +180,10 @@ export class P2P extends EventEmitter {
 						super.emit(EVENT_NEW_PEER, peer);
 						this._newPeers.add(peer.peerInfo);
 					} else {
-						existingPeer.inboundSocket = socket;
-						this._newPeers.add(existingPeer.peerInfo);
+						if (existingPeer.state.inbound === ConnectionState.DISCONNECTED) {
+							existingPeer.inboundSocket = socket;
+							this._newPeers.add(existingPeer.peerInfo);
+						}
 					}
 				}
 			},
@@ -212,8 +214,8 @@ export class P2P extends EventEmitter {
 		);
 
 		const peersToConnect = selectForConnection(availablePeers);
-		peersToConnect.forEach((peer: Peer) => {
-			peer.connect();
+		peersToConnect.forEach(async (peer: Peer) => {
+			await peer.connect();
 			this._newPeers.delete(peer.peerInfo);
 			this._triedPeers.add(peer.peerInfo);
 		});
--- a/packages/lisk-p2p/src/p2p.ts
+++ b/packages/lisk-p2p/src/p2p.ts
@@ -213,8 +213,8 @@ export class P2P extends EventEmitter {
 		);
 
 		const peersToConnect = selectForConnection(availablePeers);
-		peersToConnect.forEach(async (peer: Peer) => {
-			await peer.connect();
+		peersToConnect.forEach((peer: Peer) => {
+			peer.connect();
 			this._newPeers.delete(peer.peerInfo);
 			this._triedPeers.add(peer.peerInfo);
 		});
--- a/packages/lisk-p2p/src/p2p.ts
+++ b/packages/lisk-p2p/src/p2p.ts
@@ -22,8 +22,7 @@ import { attach, SCServer, SCServerSocket } from 'socketcluster-server';
 import { RequestFailError } from './errors';
 import { ConnectionState, Peer, PeerInfo } from './peer';
 import { discoverPeers } from './peer_discovery';
-import { PeerOptions } from './peer_selection';
-import { selectForConnection } from './peer_selection';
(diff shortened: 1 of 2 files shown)


packages/lisk-p2p/src/p2p.ts
```diff
@@ -166,50 +166,51 @@ export class P2P extends EventEmitter {
...
-							ipAddress: peerObject.ip,
-							wsPort: peerObject.wsPort, // TODO ASAP: Add more properties
-							nodeStatus: this._nodeStatus,
-							height: seedPeer.height,
-							os: seedPeer.os,
-							version: seedPeer.version,
-						}),
-				);
-			}),
+	private _getPeerToConnect(): ReadonlyArray<Peer> {
```
shuse2: can it be pure function?

packages/lisk-p2p/src/p2p.ts
```diff
@@ -166,50 +166,51 @@ export class P2P extends EventEmitter {
...
+			(peerInfo: PeerInfo) => new Peer(peerInfo),
 		);
+		const peersToConnect = selectForConnection(availablePeers);
+
+		return peersToConnect;
+	}
+
+	private _connectToPeers(): void {
+		this._getPeerToConnect().map((peer: Peer) => {
+			peer.connect();
```
shuse2, line 217: shouldn't `peer.connect` be async?

shuse2 (changes requested):

packages/lisk-p2p/src/p2p.ts
ishantiw: I think we are not waiting for the peer to connect, that's why we have 3 status in Peer.
```javascript
export enum ConnectionState {
	CONNECTING = 0,
	CONNECTED = 1,
	DISCONNECTED = 2,
}
```

@jondubois what do you think?

packages/lisk-p2p/src/p2p.ts
ishantiw: yeah it can be or maybe we can also merge it inside `_connectToPeers()` ?

ishantiw: I merged this function inside `_connectToPeers()` because it was a bit unnecessary to make a separate function just to call selection algo

packages/lisk-p2p/src/p2p.ts
```diff
@@ -166,50 +166,47 @@ export class P2P extends EventEmitter {
...
-						}),
-				);
-			}),
+	private _connectToPeers(): void {
+		const availablePeers = Array.from(this._newPeers).map(
+			(peerInfo: PeerInfo) => new Peer(peerInfo),
 		);
+
+		const peersToConnect = selectForConnection(availablePeers);
+		peersToConnect.map((peer: Peer) => {
```
shuse2: this should be `forEach` if it's not mapping to a value

packages/lisk-p2p/src/p2p.ts
```diff
@@ -166,50 +166,47 @@ export class P2P extends EventEmitter {
...
 	}
 
 	public async start(): Promise<void> {
 		await this._startPeerServer();
-		await this._loadListOfPeerListsFromSeeds(this._config.seedPeers);
+		const discoveredPeers = await this._runPeerDiscovery(
+			this._config.seedPeers,
+		);
+		// Add all the discovered peers in newPeer list
+		discoveredPeers.map((peerInfo: PeerInfo) => {
```
shuse2: this also should be `forEach` if it's not mapping

shuse2 (changes requested):

shuse2 (approved):
If @jondubois is ok with not doing the async connect, it's ok from my side

packages/lisk-p2p/src/p2p.ts
jondubois: Not necessary to wait for connect when adding to the `_newPeers` map.

For the `_triedPeers` map, I guess it depends on the LIP. Is the act of trying to connect to a peer enough to consider it a `_triedPeer` or is is necessary to wait for the connection to be successful at least once?

It's possible to send RPCs and messages to a disconnected or 'connecting' peer (in that case it will buffer while it tries to establish the connection); so the peer is usable even if in a 'connecting' state (assuming that the peer exists and is accessible publicly). Once we've called 'connect()', the socket will always try to maintain the connection open and will automatically try to reconnect in case of network failure. Does that meet the requirements of _triedPeers from the LIP perspective?

jondubois: The peer is usable in the 'connecting' state but because we're not physically connected yet, we won't receive any messages from a peer while it's in that state.

jondubois (changes requested):
Just one thing to consider.

packages/lisk-p2p/src/p2p.ts
ishantiw: In LIP `triedPeers`,

>  Containing peers to which there is or has been a successful outgoing connection. 

So that means we need to be sure that it was successfully connected and then we move it to `triedPeers`. At first if we try to connect to any peer in `newPeers`, just after first attempt if it fails to connect we remove it from `_newPeers`. In LIP, 

>  Peers in the collection newPeers are immediately deleted after a failed connection attempt as malicious peers can advertise invalid IP addresses.

So I think we need `async`, as we want to make sure we connect successfully with a peer

ishantiw: But `SCClientSocket.connect(): void` doesn't return anything (`connect()` function in Peer simply attaches a outbound socket), we will have to explicitly get the status of a peer `this._outboundSocket.state`, after certain time so that we make sure it connects. Or, is there any other way to do it? @jondubois

jondubois: yes I think that we an make connect async and make it resolve once the `socket.on('connect', ...)` event triggers.

packages/lisk-p2p/src/p2p.ts
```diff
@@ -179,8 +179,10 @@ export class P2P extends EventEmitter {
 						super.emit(EVENT_NEW_PEER, peer);
 						this._newPeers.add(peer.peerInfo);
 					} else {
-						existingPeer.inboundSocket = socket;
-						this._newPeers.add(existingPeer.peerInfo);
+						if (existingPeer.state.inbound === ConnectionState.DISCONNECTED) {
```
shuse2: `else if`?

packages/lisk-p2p/src/peer.ts
```diff
@@ -193,11 +193,17 @@ export class Peer extends EventEmitter {
...
-		this._outboundSocket.connect();
+	public async connect(): Promise<void> {
+
+		return new Promise<void>(resolve => {
+			if (!this._outboundSocket) {
+				this._outboundSocket = this._createOutboundSocket();
+			}
+			this._outboundSocket.connect();
+			this._outboundSocket.on('connect', _ => {
+				resolve();
```
shuse2: reject case?

shuse2 (changes requested):

jondubois (changes requested):
We should be more strict with our event handling strategy; within each class, we should try to bind all event listeners for a specific child object in single place (ideally upon the object's instantiation/creation) and the events should also be unbound from a single place (ideally upon the object's destruction/disposal). If we allow multiple different methods to independently manage their own listeners, it's going to get complex.

We should treat events more like hooks and keep the lifecycle of each event handler as simple as possible.

ishantiw:
> We should be more strict with our event handling strategy; within each class, we should try to bind all event listeners for a specific child object in single place (ideally upon the object's instantiation/creation) and the events should also be unbound from a single place (ideally upon the object's destruction/disposal). If we allow multiple different methods to independently manage their own listeners, it's going to get complex.
> 
> We should treat events more like hooks and keep the lifecycle of each event handler as simple as possible.

Okay as we discussed, we can have a discussion about event handling and socket management. Maybe it's out of the scope of this PR as it only needs to use `selectForConnection` method and call connect method on each `Peer` but not implement the `connect` method. Should we just only call the connect function of Peer in this PR and we can have a separate PR for implementing peer connect method?

jondubois:
> Should we just only call the connect function of Peer in this PR and we can have a separate PR for implementing peer connect method?

Sounds good.

packages/lisk-p2p/src/peer.ts
ishantiw: it will be done in a separate PR as `Peer.connect` needs a proper implementation.

ishantiw:
Ready to review. Please ignore`Peer.connect` implmenetation as it will be covered in a separate PR as it needs socket management and event handling.

shuse2 (approved):

jondubois (approved):

PR #1002, opened 2019-01-11
Add transfer transaction class - Closes #882

### LINKED ISSUE #882: Add TransferTransaction class
### Description

Extend `BaseTransaction` implement all the specific features of transfer transactions.

mitsuaki-u:
### What was the problem?

New Transfer transaction class needed.

### How did I fix it?

Added new class.

### How to test it?

npm run test

### Review checklist

* The PR resolves #882 
* Linting passes
* Tests pass


packages/lisk-transactions/src/constants.ts
```diff
@@ -40,6 +40,21 @@ export const BYTESIZES = {
 	DATA: 64,
 };
 
+interface FeeMap {
+	readonly [type: number]: number
+}
+
+export const FEES: FeeMap = {
```
shuse2: lets remove feemap and assign it each custom transaction since it's static fee?
https://github.com/LiskHQ/lisk-elements/blob/transaction_improvement_experiment/packages/lisk-transactions/src/transactions/3_vote_transaction.ts#L110

packages/lisk-transactions/src/transactions/base.ts
```diff
@@ -12,9 +12,6 @@
  * Removal or modification of this copyright notice is prohibited.
  *
  */
-// tslint:disable-next-line no-reference
-/// <reference path="../../types/browserify-bignum/index.d.ts" />
```
shuse2: this line is still needed

packages/lisk-transactions/src/transactions/0_transfer.ts
```diff
@@ -0,0 +1,480 @@
...
+			);
+		}
+		if (data.length > BYTESIZES.DATA) {
+			throw new Error('Transaction data field cannot exceed 64 bytes.');
+		}
+	}
+};
+
+export class TransferTransaction extends BaseTransaction {
+	public readonly containsUniqueData: boolean;
```
shuse2: since `containUniqueData` is default false, no need to specify here

packages/lisk-transactions/src/transactions/0_transfer.ts
```diff
@@ -0,0 +1,480 @@
...
+							tx.id,
+							error.dataPath,
+						),
+			  )
+			: [];
+		if (!typeValid || errors.length > 0) {
+			throw new TransactionMultiError('Invalid asset types', tx.id, errors);
+		}
+		this.containsUniqueData = false;
+		this.fee = calculateFee(this.type);
```
shuse2: as mentioned above, no need to calculate fee, just simply assign is fine

packages/lisk-transactions/src/transactions/0_transfer.ts
```diff
@@ -0,0 +1,480 @@
...
+							`'${error.dataPath}' ${error.message}`,
+							tx.id,
+							error.dataPath,
+						),
+			  )
+			: [];
+		if (!typeValid || errors.length > 0) {
+			throw new TransactionMultiError('Invalid asset types', tx.id, errors);
+		}
+		this.containsUniqueData = false;
```
shuse2: as mentioned above, by default it's false

packages/lisk-transactions/src/transactions/0_transfer.ts
```diff
@@ -0,0 +1,480 @@
...
+
+	protected getAssetBytes(): Buffer {
+		const { data } = this.asset;
+
+		return data ? Buffer.from(data, 'utf8') : Buffer.alloc(0);
+	}
+
+	public assetToJSON(): TransferAsset {
+		return {
+			data: this.asset.data,
```
shuse2: no need of serializeation here so maybe 
```
return {
   ...this.asset,
}
```

packages/lisk-transactions/src/transactions/0_transfer.ts
```diff
@@ -0,0 +1,480 @@
...
+	}
+
+	public verify({
+		sender,
+		recipient,
+	}: RequiredTransferState): TransactionResponse {
+		const { errors: baseErrors } = super.apply({ sender });
+		const errors = [...baseErrors];
+		// Balance verification
+		const updatedSenderBalance = new BigNum(sender.balance).sub(this.amount);
```
shuse2: This `sender` is not mutated value, so it needs to subtract both amount and fee

packages/lisk-transactions/src/transactions/0_transfer.ts
```diff
@@ -0,0 +1,480 @@
...
+					this.id,
+				),
+			);
+		}
+
+		const updatedRecipientBalance = new BigNum(recipient.balance).add(
+			this.amount,
+		);
+
+		if (!updatedRecipientBalance.lte(MAX_TRANSACTION_AMOUNT)) {
```
shuse2, line 362: im not sure this would be possible, but probably better to have?

packages/lisk-transactions/src/transactions/0_transfer.ts
```diff
@@ -0,0 +1,480 @@
...
+							`'${error.dataPath}' ${error.message}`,
+							this.id,
+							error.dataPath,
+						),
+			  )
+			: [];
+
+		errors.push(...assetErrors);
+
+		if (!validateTransferAmount(this.amount.toString())) {
```
shuse2, line 277: probably better to add fee check

packages/lisk-transactions/src/utils/calculate_fee.ts
```diff
@@ -0,0 +1,25 @@
...
+ * Removal or modification of this copyright notice is prohibited.
+ *
+ */
+import BigNum from 'browserify-bignum';
+import { FEES } from '../constants';
+
+export const calculateFee = (type: number, multiplier?: number): BigNum => {
+    // tslint:disable no-magic-numbers
+	if (type === 4) {
+		return new BigNum(FEES[type]).mul(multiplier as number);
```
shuse2: as mentioned above, since it's static fee, each transaction can handle them

shuse2 (dismissed):

packages/lisk-transactions/src/transactions/0_transfer.ts
shuse2: Also, `TransferAsset` is true in this case, but if we need serialization, it wouldn't be true, so maybe set as `object` here?
Also, maybe we should add default assetToJSON, and override them...

packages/lisk-transactions/src/constants.ts
mitsuaki-u: For multisignature registration there will be a calculation involved. Also, would it be better to make it a function now in preparation for new fee system?

packages/lisk-transactions/src/transactions/0_transfer.ts
mitsuaki-u: yes in core I believe it's being checked. would be good to check in case of errors.

packages/lisk-transactions/src/transactions/0_transfer.ts
mitsuaki-u: not sure if can set a default `assetToJSON` in base transaction since asset field does not exist there.

packages/lisk-transactions/src/constants.ts
shuse2: maybe for fee, change it to `_fee: BigNum` and add `get fee(): BigNum` and `set fee(): BigNum` for now?

packages/lisk-transactions/src/transactions/base.ts
```diff
@@ -153,6 +153,18 @@ export abstract class BaseTransaction {
...
 
+	public get fee(): BigNum {
+		if(!this._fee) {
+			throw new Error('fee is required to be set before use')
+		}
+
+		return this._fee;
+	}
+
+	public set fee(fee: BigNum) {
```
shuse2: probably we dont need setter? only updated from the constructor?

shuse2 (dismissed):
One small thing, but otherwise looks good

packages/lisk-transactions/src/transactions/3_vote_transaction.ts
```diff
@@ -107,7 +106,6 @@ const validateInputs = ({
 export class VoteTransaction extends BaseTransaction {
 	public readonly containsUniqueData = true;
 	public readonly asset: VoteAsset;
-	public readonly fee: BigNum = new BigNum(VOTE_FEE);
 
 	public constructor(tx: TransactionJSON) {
```
shuse2, line 112: we need to add assignment of the `_fee` in constructor

mitsuaki-u: Addressed.

shuse2 (approved):

SargeKhan (approved):

PR #1012, opened 2019-01-17
Add receive data and handle response mechanism - Addresses #896

jondubois:
### What was the problem?

We need to be more structured with event handling.

### How did I fix it?

- Bind event handlers in a single place and unbind events handlers from a single place.
- Improved handling of inbound PeerInfo (peer headers).
- Renamed the sanitization file and some of the methods and variables inside it to be more generic (since its also used to check the payload of `updateMyself`; not just responses).

### How to test it?

Integration tests.

### Review checklist

* The PR addresses #896

### DIFF
--- a/packages/lisk-p2p/package.json
+++ b/packages/lisk-p2p/package.json
@@ -29,7 +29,7 @@
 		"lint": "tslint --format verbose --project .",
 		"lint:fix": "npm run lint -- --fix",
 		"test": "TS_NODE_PROJECT=./test/tsconfig.json nyc mocha test/{,/**/}/*.ts",
-		"test:integration": "TS_NODE_PROJECT=./test/tsconfig.json nyc mocha test/integration/*.ts",
+		"test:integration": "TS_NODE_PROJECT=./test/tsconfig.json nyc mocha test/integration/*.ts --reporter spec --timeout 3000 --slow 3000",
 		"test:watch": "npm test -- --watch",
 		"test:watch:min": "npm run test:watch -- --reporter=min",
 		"test:node": "npm run build:check",
--- a/packages/lisk-p2p/package.json
+++ b/packages/lisk-p2p/package.json
@@ -29,7 +29,7 @@
 		"lint": "tslint --format verbose --project .",
 		"lint:fix": "npm run lint -- --fix",
 		"test": "TS_NODE_PROJECT=./test/tsconfig.json nyc mocha test/{,/**/}/*.ts",
-		"test:integration": "TS_NODE_PROJECT=./test/tsconfig.json nyc mocha test/integration/*.ts --reporter spec --timeout 3000 --slow 3000",
+		"test:integration": "TS_NODE_PROJECT=./test/tsconfig.json nyc mocha test/integration/*.ts",
 		"test:watch": "npm test -- --watch",
 		"test:watch:min": "npm run test:watch -- --reporter=min",
 		"test:node": "npm run build:check",
--- a/packages/lisk-p2p/src/p2p.ts
+++ b/packages/lisk-p2p/src/p2p.ts
@@ -200,7 +200,6 @@ export class P2P extends EventEmitter {
 				this._isActive = false;
 				resolve();
 			});
-			this._httpServer.close();
 		});
 	}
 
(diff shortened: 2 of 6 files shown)


packages/lisk-p2p/package.json
```diff
@@ -29,7 +29,7 @@
 		"lint": "tslint --format verbose --project .",
 		"lint:fix": "npm run lint -- --fix",
 		"test": "TS_NODE_PROJECT=./test/tsconfig.json nyc mocha test/{,/**/}/*.ts",
-		"test:integration": "TS_NODE_PROJECT=./test/tsconfig.json nyc mocha test/integration/*.ts",
+		"test:integration": "TS_NODE_PROJECT=./test/tsconfig.json nyc mocha test/integration/*.ts --reporter spec --timeout 3000 --slow 3000",
```
shuse2: do we need this setup? i think it's covered in mochaopt

packages/lisk-p2p/src/p2p.ts
```diff
@@ -195,16 +192,31 @@ export class P2P extends EventEmitter {
...
 
-	private async _stopPeerServer(): Promise<void> {
-		// TODO ASAP: Test this and check for potential failure scenarios.
+	private async _stopHTTPServer(): Promise<void> {
+		return new Promise<void>(resolve => {
+			this._httpServer.close(() => {
+				this._isActive = false;
+				resolve();
+			});
+			this._httpServer.close();
```
shuse2: closing twice...?

shuse2 (changes requested):

packages/lisk-p2p/src/p2p.ts
jondubois: Copy-paste gone wrong.

packages/lisk-p2p/src/peer.ts
```diff
@@ -131,12 +133,10 @@ export class Peer extends EventEmitter {
 
 	public set inboundSocket(value: SCServerSocket) {
```
ishantiw: instead of `value`, maybe a camelcase name `scServerSocket`

packages/lisk-p2p/src/peer.ts
```diff
@@ -278,48 +280,63 @@ export class Peer extends EventEmitter {
...
-		socket: SCServerSocket | SCClientSocket,
+	// All event handlers for the outbound socket should be unbound in this method.
+	/* tslint:disable-next-line:prefer-function-over-method*/
+	private _unbindHandlersFromOutboundSocket(
+		outboundSocket: SCClientSocket,
 	): void {
-		// If an existing listener is bound, remove it.
-		socket.off(REMOTE_EVENT_RPC_REQUEST, this._handleRPC);
-		socket.on(REMOTE_EVENT_RPC_REQUEST, this._handleRPC);
+		outboundSocket.off();
```
ishantiw, line 310: should it be `outboundSocket.destroy()` ? Or, we can call `dropOutboundConnection()` from #1014 since we don't need this socket so its better to destroy?

packages/lisk-p2p/src/peer.ts
```diff
@@ -278,48 +280,63 @@ export class Peer extends EventEmitter {
...
-		socket: SCServerSocket | SCClientSocket,
-	): void {
-		socket.off(REMOTE_EVENT_RPC_REQUEST, this._handleMessage);
+	private _handlePeerInfo(request: ProtocolRPCRequest): void {
+		// Update peerInfo with the latest values from the remote peer.
+		// TODO ASAP: Validate and/or sanitize the request.data as a PeerInfo object.
+		try {
+			const newPeerInfo = sanitizePeerInfo(request.data);
+			// Merge new info with existing info.
+			this._peerInfo = {
```
ishantiw, line 333: I think it should only allow changing few values, in this case however even the `ipAddress`  and `port` could be updated. Maybe only `height`, `version` should be allowed to be updated?

packages/lisk-p2p/src/peer.ts
jondubois: Destroying the socket doesn't unbind the listeners. But because the socket becomes inactive, those listeners should never trigger again anyway. But it's still good practice to cleanup the listeners as a separate step anyway; in case we decide to switch back to disconnect instead of destroy later (in that case cleaning up the listeners will become crucial again).

ishantiw (approved):

shuse2 (approved):

PR #1014, opened 2019-01-17
Add separate functions to drop inbound and outbound connections -Closes #1005

### LINKED ISSUE #1005: Add separate functions to drop inbound and outbound connections
### Description

As of now, we have a `disconnect()` function that drops both inbound and outbound connection. Moving forward we want more granular functions to drop either inbound or outbound connection.

ishantiw:
### Description

Add function to drop inbound and outbound connections of a peer to explicitly drop either connection

### Review checklist

* The PR resolves #1005 
* Linting passes
* Tests pass

### DIFF
--- a/packages/lisk-p2p/src/peer.ts
+++ b/packages/lisk-p2p/src/peer.ts
@@ -214,6 +214,19 @@ export class Peer extends EventEmitter {
 		}
 	}
 
+	public dropInboundConnection(code: number = 1000, reason?: string): void {
+		if (this._inboundSocket) {
+			this._inboundSocket.destroy(code, reason);
+			this._stopListeningForRPCs(this._inboundSocket);
+		}
+	}
+
+	public dropOutboundConnection(code: number = 1000, reason?: string): void {
+		if (this._outboundSocket) {
+			this._outboundSocket.destroy(code, reason);
+		}
+	}
+
 	public send<T>(packet: P2PMessagePacket<T>): void {
 		if (!this._outboundSocket) {
 			this._outboundSocket = this._createOutboundSocket();
--- a/packages/lisk-p2p/src/peer.ts
+++ b/packages/lisk-p2p/src/peer.ts
@@ -205,13 +205,8 @@ export class Peer extends EventEmitter {
 	}
 
 	public disconnect(code: number = 1000, reason?: string): void {
-		if (this._inboundSocket) {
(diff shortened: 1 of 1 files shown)


packages/lisk-p2p/src/peer.ts
```diff
@@ -214,6 +214,19 @@ export class Peer extends EventEmitter {
...
+		if (this._inboundSocket) {
+			this._inboundSocket.destroy(code, reason);
+			this._stopListeningForRPCs(this._inboundSocket);
+		}
+	}
+
+	public dropOutboundConnection(code: number = 1000, reason?: string): void {
+		if (this._outboundSocket) {
+			this._outboundSocket.destroy(code, reason);
+		}
```
shuse2, line 222: we don't need to stop listening for RPC for outbound?

ishantiw: I think we are only listening for RPC's on inbound connection [peer.ts#L142](https://github.com/LiskHQ/lisk-elements/blob/1a786675868225175bd77db05f2a91ebfb099cd3/packages/lisk-p2p/src/peer.ts#L142) and not listening for RPC on outbound.

packages/lisk-p2p/src/peer.ts
```diff
@@ -214,6 +214,19 @@ export class Peer extends EventEmitter {
 		}
 	}
 
+	public dropInboundConnection(code: number = 1000, reason?: string): void {
+		if (this._inboundSocket) {
+			this._inboundSocket.destroy(code, reason);
+			this._stopListeningForRPCs(this._inboundSocket);
```
shuse2, line 215: `_stopListeningForMessages`? also, probably we should stop listening before destroying

ishantiw: That's true, I will also modify [disconnect()](https://github.com/LiskHQ/lisk-elements/blob/1a786675868225175bd77db05f2a91ebfb099cd3/packages/lisk-p2p/src/peer.ts#L207) function as well, as these functions are derived from it

ishantiw: @shuse2 As discussed with @jondubois, it’s generally better to destroy and then stop listening; that way it will capture the `disconnect` event

shuse2 (approved):

jondubois (approved):

PR #1020, opened 2019-01-18
Add functions in sanitization to handler ProtocolRPCRequest and ProtocolMessage - Closes#1017

### LINKED ISSUE #1017: Add functions in sanitization to handler ProtocolRPCRequest and ProtocolMessage
### Description

Add functions in sanitization to handle ProtocolRPCRequest and ProtocolMessage coming from the network instead of handling it inside P2P.

ishantiw:
### Description

Add sanitization functions to handle protocol RPC requests and messages

### Review checklist

* The PR resolves #1017 
* Linting passes
* Tests pass


packages/lisk-p2p/src/sanitization.ts
```diff
@@ -92,3 +97,32 @@ export const sanitizePeerInfoList = (
 		throw new InvalidRPCResponse('Invalid response type');
 	}
 };
+
+export const sanitizeRPCRequest = (request: unknown): ProtocolRPCRequest => {
```
shuse2: instead of `sanitize`, can we use `validate`?

packages/lisk-p2p/src/errors.ts
```diff
@@ -69,3 +69,17 @@ export class RequestFailError extends VError {
 		this.name = 'RequestFailError';
 	}
 }
+
+export class InvalidRPCRequest extends VError {
```
shuse2: prefer `InvalidRPCRequestError`

packages/lisk-p2p/src/errors.ts
```diff
@@ -69,3 +69,17 @@ export class RequestFailError extends VError {
...
+	public constructor(message: string) {
+		super(message);
+		this.name = 'InvalidRPCRequest';
+	}
+}
+
+export class InvalidProtocolMessage extends VError {
```
shuse2: prefer `InvalidProtocolMessageError`

shuse2 (changes requested):

ishantiw:
Addressed all the comments and rebased. Also, I have included the changes related to `socketcluster-server` library in #1021, where I forgot to upgrade the version of the library

packages/lisk-p2p/src/peer.ts
```diff
@@ -106,49 +104,34 @@ export class Peer extends EventEmitter {
...
-				// This request doesn't need to propagate to its parent class.
-				this._handlePeerInfo(request);
+				const request = new P2PRequest(
+					rawRequest.procedure,
+					rawRequest.data,
+					respond,
+				);
 
-				return;
+				this.emit(EVENT_REQUEST_RECEIVED, request);
```
jondubois: It's dangerous to emit within a `try` block because event listeners which are declared outside this class will be called synchronously; so if a listener is attached to the Peer from outside; for example using `peer.on(EVENT_REQUEST_RECEIVED, myListenerFunction)`; it means that if `myListenerFunction` throws an error or has a bug then it will cause the catch block below to execute and `this.emit(EVENT_INVALID_REQUEST_RECEIVED, packet);` will be called even though the request is fine; makes it very hard to debug.

An alternative approach might be to `return` from inside the `catch` block just after calling `this.emit(EVENT_INVALID_REQUEST_RECEIVED, packet);` and on the next line outside the catch block (which will only run if request is valid), we can emit the request with `this.emit(EVENT_REQUEST_RECEIVED, request);`.

It's easy to do with `let` but may be kind of tricky to do with only `const`. Maybe it's worth separating the validation logic from the sanitization logic and only have the try-catch block around the validation part?

Another option would be to emit asynchronously (e.g. inside a setTimeout/setImmediate) but that is more ugly and computationally expensive.

packages/lisk-p2p/src/peer.ts
```diff
@@ -106,49 +104,34 @@ export class Peer extends EventEmitter {
...
 			// TODO later: Switch to LIP protocol format.
-			// TODO ASAP: Move validation/sanitization to sanitization.ts with other validation logic.
-			const rawMessage = packet as ProtocolMessagePacket;
-			if (!rawMessage || typeof rawMessage.event !== 'string') {
-				this.emit(EVENT_INVALID_MESSAGE_RECEIVED, rawMessage);
-
-				return;
+			try {
+				const protocolMessage = sanitizeProtocolMessage(packet);
+				this.emit(EVENT_MESSAGE_RECEIVED, protocolMessage);
```
jondubois: Same problem as above.

packages/lisk-p2p/src/sanitization.ts
```diff
@@ -92,3 +98,32 @@ export const sanitizePeerInfoList = (
 		throw new InvalidRPCResponse('Invalid response type');
 	}
 };
+
+export const validateRPCRequest = (request: unknown): P2PRequest => {
+	if (!request) {
+		throw new InvalidRPCRequestError('Invalid request');
+	}
+
+	const rpcRequest = request as P2PRequest;
```
jondubois: We can't cast the `request` object into a `P2PRequest ` because `P2PRequest` is a usable request object which can be responded to using its `request.end(...)` or `request.error(...)` method. Maybe this validation/sanitization function should just return a validated/sanitized `ProtocolRPCRequestPacket`; that's also more consistent with the `sanitizeProtocolMessage` function below. We will have to instantiate the `P2PRequest` as a separate step (not part of validation/sanitization); it needs the SocketCluster `respond` callback as third argument to its constructor.

jondubois (changes requested):

packages/lisk-p2p/src/sanitization.ts
```diff
@@ -92,3 +98,32 @@ export const sanitizePeerInfoList = (
...
+		throw new InvalidRPCRequestError('Request procedure name is not a string');
+	}
+	if (typeof rpcRequest.data !== 'object') {
+		throw new InvalidRPCRequestError('Invalid request data');
+	}
+
+	return rpcRequest;
+};
+
+export const sanitizeProtocolMessage = (message: unknown): ProtocolMessagePacket => {
```
shuse2: name of this should also be `validate`?

packages/lisk-p2p/src/sanitization.ts
```diff
@@ -14,8 +14,14 @@
  */
```
shuse2, line 14: also, filename should be consistent, and all the function in this file?

shuse2 (changes requested):

packages/lisk-p2p/src/sanitization.ts
ishantiw: actually before, all the functions were prefixed as process or validate. But later @jondubois renamed all of them to sanitize including the file name because these functions are not only validating but type casting or sanitizing it to some protocol type or preferred data structure.

packages/lisk-p2p/src/sanitization.ts
ishantiw: same here, should we rename it back to `validate` prefix? what do you think @jondubois ?

packages/lisk-p2p/src/sanitization.ts
shuse2: Mmm we have similar functionality in transaction with validating and the checking so maybe it’s better to standardize?

ishantiw: Okay then I think it makes sense to just use `validate` here too  👍

ishantiw:
Addressed @jondubois @shuse2

jondubois (approved):

shuse2 (approved):

PR #1023, opened 2019-01-18
Move peer discovery and peer connection related stuff to peer pool - Closes#1019

### LINKED ISSUE #1019: Move peer discovery and peer connection related stuff to peer pool
### Description

As per discussion with @jondubois , it will be better if we could move `Peer` object initialization and connect/disconnect in `PeerPool` so that we make sure we only make connections and disconnect from a `Peer` in `PeerPool`. In this way, we don't accidentally connect to a peer somewhere where we should have no connection or we dropped connection already for that peer. Also, we will bind and unbind only in `PeerPool` to avoid confusion and make it simpler.

- Move peer selection for connection to `PeerPool`
- Move peer discovery to `PeerPool` and have an abstract function at `P2P` level to call it
- Pass `PeerInfo` to `PeerPool` instead of creating a `Peer` object for incoming connection in `P2P` and let `PeerPool` handle the connection

ishantiw:
### Description

- Move peer selection for connection to `PeerPool`
- Move peer discovery to `PeerPool` and have an abstract function at `P2P` level to call it
- Pass `PeerInfo` to `PeerPool` instead of creating a `Peer` object for incoming connection in `P2P` and let `PeerPool` handle the connection

### Review checklist

* The PR resolves #1019 
* Linting passes
* Tests pass

### DIFF
--- a/packages/lisk-p2p/src/p2p.ts
+++ b/packages/lisk-p2p/src/p2p.ts
@@ -218,7 +218,7 @@ export class P2P extends EventEmitter {
 						version: queryObject.version,
 					};
 
-					const isNewPeer = this._peerPool.addInBoundPeer(
+					const isNewPeer = this._peerPool.addInboundPeer(
 						peerId,
 						incomingPeerInfo,
 						socket,
--- a/packages/lisk-p2p/src/p2p.ts
+++ b/packages/lisk-p2p/src/p2p.ts
@@ -298,14 +298,7 @@ export class P2P extends EventEmitter {
 	public async start(): Promise<void> {
 		await this._startPeerServer();
 
-		const discoveredPeers = await this._runPeerDiscovery(
-			this._config.seedPeers,
-		);
-
-		// Add all the discovered peers in newPeer list
-		discoveredPeers.forEach((peerInfo: PeerInfo) => {
-			this._newPeers.add(peerInfo);
-		});
+		await this._runPeerDiscovery(this._config.seedPeers);
 
 		this._connectToPeers([...this._newPeers]);
 	}
--- a/packages/lisk-p2p/src/p2p.ts
+++ b/packages/lisk-p2p/src/p2p.ts
@@ -228,9 +228,10 @@ export class P2P extends EventEmitter {
 					if (isNewPeer) {
 						this.emit(EVENT_NEW_INBOUND_PEER, incomingPeerInfo);
 						this.emit(EVENT_NEW_PEER, incomingPeerInfo);
-						this._newPeers.add(incomingPeerInfo);
-					} else {
-						this._newPeers.add(incomingPeerInfo);
+
+						if (!this._newPeers.has(incomingPeerInfo)) {
+							this._newPeers.add(incomingPeerInfo);
+						}
 					}
 				}
 			},
--- a/packages/lisk-p2p/src/p2p.ts
+++ b/packages/lisk-p2p/src/p2p.ts
@@ -187,7 +187,7 @@ export class P2P extends EventEmitter {
 			'connection',
 			(socket: SCServerSocket): void => {
 				if (!socket.request.url) {
-					super.emit(EVENT_FAILED_TO_ADD_INBOUND_PEER);
+					this.emit(EVENT_FAILED_TO_ADD_INBOUND_PEER);
 					socket.disconnect(
 						INVALID_CONNECTION_URL_CODE,
 						INVALID_CONNECTION_URL_REASON,
@@ -205,7 +205,7 @@ export class P2P extends EventEmitter {
 						INVALID_CONNECTION_QUERY_CODE,
 						INVALID_CONNECTION_QUERY_REASON,
 					);
-					super.emit(EVENT_FAILED_TO_ADD_INBOUND_PEER);
+					this.emit(EVENT_FAILED_TO_ADD_INBOUND_PEER);
 				} else {
 					const wsPort: number = parseInt(queryObject.wsPort, BASE_10_RADIX);
 					const peerId = Peer.constructPeerId(socket.remoteAddress, wsPort);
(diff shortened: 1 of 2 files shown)


packages/lisk-p2p/src/p2p.ts
```diff
@@ -158,28 +157,28 @@ export class P2P extends EventEmitter {
...
+
+					const isNewPeer = this._peerPool.addInBoundPeer(
+						peerId,
+						incomingPeerInfo,
+						socket,
+						this._nodeInfo,
+					);
+
+					if (isNewPeer) {
+						super.emit(EVENT_NEW_INBOUND_PEER, incomingPeerInfo);
```
shuse2: we are extending the EventEmitter, so just use `this.emit`?

packages/lisk-p2p/src/peer_pool.ts
```diff
@@ -43,10 +50,72 @@ export class PeerPool extends EventEmitter {
...
+	public addInBoundPeer(
+		peerId: string,
+		peerinfo: PeerInfo,
+		socket: SCServerSocket,
+		nodeInfo: P2PNodeInfo,
+	): boolean {
+		const existingPeer = this.getPeer(peerId);
+
+		if (existingPeer) {
+			if (existingPeer.state.inbound === ConnectionState.DISCONNECTED) {
```
shuse2, line 149: should we check if socket exists or not?

packages/lisk-p2p/src/p2p.ts
```diff
@@ -158,28 +157,28 @@ export class P2P extends EventEmitter {
...
+						super.emit(EVENT_NEW_PEER, incomingPeerInfo);
+						this._newPeers.add(incomingPeerInfo);
+					} else {
+						this._newPeers.add(incomingPeerInfo);
```
shuse2: if it's not new peer, still add to the list? if anyway adding to the `_newPeers` prefer to do it out side of `if/else`

shuse2 (changes requested):

packages/lisk-p2p/src/peer_pool.ts
ishantiw: Whenever we are adding any peer `this._peerMap` we assign socket to it so in this way, we make sure we always have some socket assigned to a peer if its present in `this._peerMap`. We only assign a new socket when the current socket is in `DISCONNECTED` state.

packages/lisk-p2p/src/p2p.ts
ishantiw: Forgot to remove `else` condition but anyways, we need a separate condition to check if it exists in `_newPeer` or not

packages/lisk-p2p/src/p2p.ts
```diff
@@ -187,32 +186,33 @@ export class P2P extends EventEmitter {
...
+
+					const incomingPeerInfo: PeerInfo = {
+						ipAddress: socket.remoteAddress,
+						wsPort,
+						height: queryObject.height ? +queryObject.height : 0,
+						os: queryObject.os,
+						version: queryObject.version,
+					};
+
+					const isNewPeer = this._peerPool.addInBoundPeer(
```
jondubois: Rename `addInBoundPeer` to `addInboundPeer`.

packages/lisk-p2p/src/p2p.ts
```diff
@@ -250,35 +250,30 @@ export class P2P extends EventEmitter {
 		await this._stopHTTPServer();
 	}
 
-	private _connectToPeers(): void {
-		const availablePeers = Array.from(this._newPeers).map(
-			(peerInfo: PeerInfo) => new Peer(peerInfo),
-		);
+	private _connectToPeers(peers: ReadonlyArray<PeerInfo>): void {
```
jondubois: The `this._peerPool.connectToPeers(...)` seems like an implementation detail which could be part of the `this._peerPool.runDiscovery(...)` method. Also, the name `connectToPeers` is confusing because the PeerPool will only connect to some of the peers we provide (not all of them).

If we make the `connectToPeers` method private and integrate it as part of the `runDiscovery` method, maybe we can find a simpler way to populate the `_newPeers` and `_triedPeers` maps:

- We can emit new kinds of events on the PeerPool and then listen to those events from the P2P class to populate the _newPeers and _triedPeers?
- Or we can move the _newPeers and _triedPeers maps into the PeerPool (and expose them to the P2P class via getter functions)?

jondubois (changes requested):

packages/lisk-p2p/src/p2p.ts
ishantiw: We don't want to make `this._peerPool.connectToPeers(...)` part of `this._peerPool.runDiscovery(...)` because we still want to keep them independent as they could be called separately as well. For current protocol, of course, it runs discovery and `connectToPeers` together on startup of a node but later when we will implement lip, it doesn't necessary means that we will always run connectToPeers after discovery or we might want to call  `connectToPeers` separately without running discovery, for example, in periodic shuffling(it doesn't involve discovery and it connects with peers randomly).

shuse2 (approved):

packages/lisk-p2p/src/p2p.ts
jondubois: @ishantiw When we introduce shuffling, we could add a new `peerPool.shufflePeers()` method? Shuffling will also involve disconnecting from old peers (e.g. `disconnectFromPeers`); both `connectToPeers` and `disconnectFromPeers` could be private methods of the PeerPool used by the `peerPool.shufflePeers()`.

My concern is also partly related to the name of the method `peerPool.connectToPeers(peers)`. It sounds like it will connect to all provided peers which is not the case. Maybe it should be called `peerPool.runPeerSelectionAndConnection(peers)` (or something similar); that way it sounds like the same level of abstraction as `peerPool.runDiscovery()`.

jondubois (changes requested):

packages/lisk-p2p/src/p2p.ts
ishantiw: actually after merging #1030  we don't really need `p2p._connectToPeers()` because all we are doing in this method is run the method `connectToPeers()` of `PeerPool` and getting a list of `PeerInfo` in return which was chosen to connect but we are not actually sure that we connected. So I will remove this function in P2P and add the statement to run `peerPool.connectToPeers` rename to `peerPool.selectPeersToConnect` in p2p.Start(), like you said that we need a proper name. 

Yes, for shuffling I think we will do it `PeerPool` in a separate handler. But still, we need to notify p2p to manage peer bucket. I guess it could become easier if we could have buckets in `PeerPool` 🤔

jondubois: 👍 `peerPool.selectPeersToConnect` sounds good.

jondubois (approved):

PR #1039, opened 2019-01-25
Fix the bug in verifying validated transactions - Closes #963

### LINKED ISSUE #963: Fix the bug in verifying validated transactions
### Expected behavior
During the verification of validated transactions, only the transactions which are verified against the current blockchain state should be moved to the verified queue.
### Actual behavior
There is an edge case in which transactions which are verified against the outdated blockchain state are moved to the verified queue. The description of the edge case is as follow:
Transaction pool registers a job which verifies the transactions from the validated queue. This job is called "verifyValidatedTransactions". This job runs asynchronously. The steps included are as follow: 
1- It peeks transactions from the validated queue, 
2- Verifies transactions
3- Moves the verified transactions which still exist in the validated queue to the verified queue.
4- Removes unverifiable transactions from the validated queue.
Now imagine during the processing of "verifyValidatedTransactions" job, blockchain state changes which may affect some transactions which are currently being processed in the job. 
The job will try to move these transactions to the verified queue even though they were verified against outdated blockchain state. This will lead to inconsistent data in the blockchain.

SargeKhan:
### What was the problem?
There was an edge case where if transactions in the validated queue are being verified, and during this operation, the blockchain state changes, the transactions will be moved to the verified queue although they were verified against the old blockchain state. 
### How did I fix it?
When transactions which exist in the validated queue are affected by the blockchain state change, they are moved to the received queue. This ensures that the race condition between blockchain state change event and verifying transactions is prevented. 
### How to test it?
Run tests.
### Review checklist

* The PR resolves #963 
* Linting passes
* Tests pass


shuse2 (approved):

packages/lisk-transaction-pool/src/transaction_pool.ts
```diff
@@ -239,13 +239,26 @@ export class TransactionPool extends EventEmitter {
 	): void {
 		const { received, validated, ...otherQueues } = this._queues;
 
+		// Move transactions from the validated queue to the received queue where account was a receipient in the verified removed transactions
+		// Rationale is explained in issue #963
+		const remoevdTransactionsByRecipientIdFromValidatedQueue = this._queues.validated.removeFor(
```
mitsuaki-u: `remoevd` -> `removed`

packages/lisk-transaction-pool/src/transaction_pool.ts
```diff
@@ -310,17 +323,38 @@ export class TransactionPool extends EventEmitter {
 		);
 
 		const { received, validated, ...otherQueues } = this._queues;
+
+		const confirmedTransactionsWithUniqueData = transactions.filter(
```
mitsujutsu: where is this being used?

mitsuaki-u (approved):

PR #1041, opened 2019-01-25
Add getAddressFromPrivateKey - Closes #952

### LINKED ISSUE #952: new crypto method "getAddressFromPrivateKey"
I saw cryptography module always assume the use of a passphrase to get the keys.

Is there a way through the `cryptography` module to craft a `getAddressFromPrivateKey` method?
and so the get a public or address from a privateKey instead of a passphrase?

I used this at the moment `nacl.sign.keyPair.fromSecretKey(Buffer.from(privKey, 'hex'))`

shuse2:
### What was the problem?
There was no way to generate address from private key

### How did I fix it?
Add a function to generate publicKey from private key, and then add exposed function to convert private key to address

### How to test it?
`npm t`

### Review checklist

* The PR resolves #952 
* Linting passes
* Tests pass

### DIFF
--- a/packages/lisk-cryptography/test/keys.ts
+++ b/packages/lisk-cryptography/test/keys.ts
@@ -21,6 +21,7 @@ import {
 	getKeys,
 	getAddressAndPublicKeyFromPassphrase,
 	getAddressFromPassphrase,
+	getAddressFromPrivateKey,
 } from '../src/keys';
 // Require is used for stubbing
 const buffer = require('../src/buffer');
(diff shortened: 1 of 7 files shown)


AndreasKendziorra (dismissed):

shuse2:
@AndreasKendziorra Addressed
> shouldn't it be tweetnacl.sign.secretKeyLength?

this was 64 bytes, so it couldn't be used. I created an internal constant instead.

packages/lisk-cryptography/test/keys.ts
```diff
@@ -133,4 +134,12 @@ describe('keys', () => {
 			);
 		});
 	});
+
+	describe('#getAddressFromPrivateKey', () => {
+		it('should create correct address', () => {
+			return expect(
+				getAddressFromPrivateKey(defaultPrivateKey.slice(0, 64)),
```
AndreasKendziorra, line 141: I think the same test, but with 32 byte input, would be good. Or do you think its a bit of overdoing since both cases or covered with tests for getPublicKey?

shuse2: yea i think it's covered by the getPublicKey, so not needed here

AndreasKendziorra: ok

AndreasKendziorra (approved):

mitsuaki-u (approved):

Issue #900, opened 2018-11-22
Add discovery process


shuse2:
### Description

lisk-p2p needs to implement discovery process described in https://github.com/LiskHQ/lips/blob/master/proposals/lip-0004.md

jondubois:
The initial round of discovery is working but we need to run it multiple times periodically.

PR #1048, opened 2019-01-28
Make current integration tests pass - Closes #971 and addresses #900

### LINKED ISSUE #971: Make current integration tests pass
### Description

Make the core integration tests pass.

jondubois:
### What was the problem?

Core integration tests related to launching a node, connecting to seed peers and doing the first round of discovery were failing. We hadn't yet brought the node into a verifiable working state.

### How did I fix it?

- Fixed bugs in several places; two related to removing event listeners incorrectly.
- Allowed using real timers in integration tests. Added an environment variable which can be used to optionally enable them (though they're still disabled by default).
- Renamed PeerInfo to P2PPeerInfo since it is exposed publicly by the API (e.g. as part of `getNetworkStatus`).
- Modified `P2PPeerInfo` and `P2PNodeInfo` to be more generic; added an `options` field which can be used to specify custom fields/values on `P2PPeerInfo`/`P2PNodeInfo` objects. This allows the P2P module to support more different kinds of applications.
- More consistent handling of `P2PNodeInfo` and `P2PPeerInfo` throughout `P2P -> PeerPool -> Peer`.
- Added some `TODO` comments.
- Made `selectForConnection` work with `P2PPeerInfo` objects instead of `Peer` objects; it doesn't make sense to filter `Peer` objects if we are yet to connect to them; it forced us to instantiate `Peer` objects before we needed them. We should think about other selection functions as well if we want to use `P2PPeerInfo` instead of `Peer`; the `P2PPeerInfo` is safer to expose to users but may offer less control/flexibility in certain scenarios.
- ~~`selectPeers` function is now used by the `list` endpoint because we can't send back our full peer list and the `selectPeers` function seemed like a good way to do that type of filtering. As a result of this, `selectPeers` was changed to only return tried peers for which we have complete info.~~
- ^ selectPeers relies on peer height which is not relevant for sending back our peer list. Instead, the node now sends back a subset of their randomly shuffled peer list; uses the same algorithm as current Lisk core node.
- Made integration tests work with basic discovery scenarios; we still need to add assertions to existing test cases and also add more new test cases; right now the node only does a single round of discovery and this is a problem when all nodes launch at the same time (which is the current situation in integration tests).

### How to test it?

`npm run test:integration`

### Review checklist

* The PR resolves #971 and addresses #900
* Linting passes
* Tests pass


packages/lisk-p2p/src/p2p.ts
```diff
@@ -57,36 +58,34 @@ export const EVENT_NEW_INBOUND_PEER = 'newInboundPeer';
...
 	private readonly _httpServer: Server;
 	private _isActive: boolean;
-	private readonly _newPeers: Set<PeerInfo>;
-	private readonly _triedPeers: Set<PeerInfo>;
+	private readonly _newPeers: Set<P2PPeerInfo>;
+	private readonly _triedPeers: Set<P2PPeerInfo>;
+	private readonly _platform: string;
 
-	private _nodeInfo: P2PNodeInfo;
+	private _nodeInfo: P2PNodeInfo | undefined;
```
shuse2: why is this undefinable?
shouldn't information always supplied in constructor?

packages/lisk-p2p/src/sanitization.ts
```diff
@@ -0,0 +1,93 @@
+/*
+ * Copyright © 2018 Lisk Foundation
+ *
+ * See the LICENSE file at the top-level directory of this distribution
```
shuse2: did this file came back? i think name changed to `validation`

packages/lisk-p2p/test/_setup.ts
```diff
@@ -22,6 +22,10 @@ process.env.NODE_ENV = 'test';
...
 [sinonChai, chaiAsPromised].forEach(plugin => chai.use(plugin));
 
-global.sandbox = sinon.createSandbox({
-	useFakeTimers: true,
-});
+if (process.env.USE_REAL_TIMERS === 'true') {
+	global.sandbox = sinon.createSandbox({});
+} else {
+	global.sandbox = sinon.createSandbox({
+		useFakeTimers: true,
```
shuse2: maybe below is cleaner?
```
global.sandbox = sinon.createSandbox({
	useFakeTimers: process.env.USE_REAL_TIMERS !== 'true',
});
```

shuse2 (changes requested):
Could you create issue for each `todo` you added?
Also, there is lint problem
```
ERROR: (ordered-imports) /home/lisk/workspace/lisk-elements_PR-1048/packages/lisk-p2p/src/peer_discovery.ts[16, 1]: Import sources within a group must be alphabetized.

ERROR: (ordered-imports) /home/lisk/workspace/lisk-elements_PR-1048/packages/lisk-p2p/src/peer_selection.ts[17, 1]: Import sources within a group must be alphabetized.

ERROR: (ordered-imports) /home/lisk/workspace/lisk-elements_PR-1048/packages/lisk-p2p/src/sanitization.ts[18, 10]: Named imports must be alphabetized.

```

packages/lisk-p2p/src/p2p.ts
```diff
@@ -122,24 +121,32 @@ export class P2P extends EventEmitter {
...
-		const peerList = this._peerPool.getAllPeers();
-		peerList.forEach(peer => {
-			peer.applyNodeInfo(nodeInfo);
-		});
+		this._nodeInfo = {
+			os: this._platform,
+			version: this._config.version,
+			wsPort: this._config.wsPort,
+			height: nodeInfo.height,
+			options: nodeInfo.options,
```
ishantiw: I think `nodeInfo.options` could be `undefined` as `options` is an optional field in `P2PNodeInfo` so maybe we can have check if it's `undefined`, if it is then atleast assign a blank object

packages/lisk-p2p/src/p2p.ts
```diff
@@ -122,24 +121,32 @@ export class P2P extends EventEmitter {
...
+		};
+
+		this._peerPool.applyNodeInfo(this._nodeInfo);
+	}
+
+	public get nodeInfo(): P2PNodeInfo | undefined {
```
ishantiw: I think we can set basic` nodeInfo` in the constructor using values coming with `P2PConfig`, except for field height, which we can default to `0`

packages/lisk-p2p/src/peer_pool.ts
```diff
@@ -196,7 +207,22 @@ export class PeerPool extends EventEmitter {
...
 
 	private _handleGetAllPeersRequest(request: P2PRequest): void {
-		request.end(this.getAllPeerInfos());
+		const peerSelectionParams: PeerOptions = {
+			lastBlockHeight: this._nodeInfo ? this._nodeInfo.height : 0,
+		};
+
+		request.end({
+			success: true,
+			peers: this.selectPeers(peerSelectionParams).map((peer: Peer) => {
```
ishantiw: maybe it's better to move this calculation above and assign to a const variable. Will be easier to debug I guess

packages/lisk-p2p/test/unit/peer_selection.ts
```diff
@@ -28,6 +28,8 @@ describe('peer selector', () => {
 			netHash: '73458irc3yb7rg37r7326dbt7236',
 		};
 
+		// TODO ASAP: Consider updating failing test cases. The peer selector currently only returns Peer objects for tried peers for which we have complete info.
+		// TODO ASAP: We need to decide if this filtering needs to happen at the level of the peer selector or if there are other ways.
 		describe('get list of n number of good peers', () => {
```
ishantiw: `get a list`

ishantiw (changes requested):
Maybe its not a good idea to expose `isTriedPeer` field to the user as its more low level concept of managing peer bucket. Maybe its better if we could have a separate `P2PPeerInfo` without `isTriedPeer` and we can extend it to have a internal PeerInfo definition that will hold more fields related to peer bucket management and selection algo like `isBannedPeer` from LIP

packages/lisk-p2p/src/p2p.ts
jondubois: It's possible to provide nodeInfo to the constructor but then I think we should be clear about how it's provided to the constructor; maybe by adding an explicit `nodeInfo` field. I think this makes sense actually. `nodeInfo` will be information which may be used internally by the node and which should be exposed publicly to other nodes.

jondubois:
@ishantiw About `isTriedPeer`; that sounds good. We should do this as a separate PR though because otherwise this PR will get huge and it's already addressing too many problems. At least once we've merged this PR we can change stuff and then run the integration tests to make sure that we don't break things.

packages/lisk-p2p/src/p2p.ts
jondubois: I changed it so that the user needs to pass a new `nodeInfo` fields to the P2P constructor. This guarantees that P2PNodeInfo is always defined and is more explicit.

ishantiw (approved):
Looks good

shuse2 (approved):

PR #1065, opened 2019-01-31
Remove fromJSON and create function from custom transaction - Closes #1057

### LINKED ISSUE #1057: Remove fromJSON and create function from custom transaction
### Description
- Remove `fromJSON` for each custom transactions because having 2 ways to instantiate the class is bit confusing for user.

- Remove `create` for each custom transactions to obtain the signed transaction.
Instead, each create transaction function (0_transfer, etc) should use the custom class created instead.

Comparing the create method in the transaction and having different file to create transaction, creating the transaction is simple enough using the class, and but still clear to do in the different file.

pablitovicente:
### What was the problem?

- `fromJSON()` method provided an extra way of creating transactions that was not needed
- `create()` method had to be removed

### How did I fix it?

Remove fromJSON for each custom transactions because having 2 ways to instantiate the class is bit confusing for user.

Remove create for each custom transactions to obtain the signed transaction.
Instead, each create transaction function (0_transfer, etc) should use the custom class created instead.

### How to test it?

- Build should be green

### Review checklist

* The PR resolves #1057
* Linting passes
* Tests pass


packages/lisk-transactions/src/0_transfer.ts
```diff
@@ -44,81 +30,48 @@ export interface TransferInputs {
 	readonly timeOffset?: number;
 }
 
-const validateInputs = ({
```
shuse2: we should keep this `validateInputs` to this file as it's only used here

packages/lisk-transactions/src/0_transfer.ts
```diff
@@ -44,81 +30,48 @@ export interface TransferInputs {
...
 		: undefined;
 	const recipientId = inputs.recipientId
 		? inputs.recipientId
 		: recipientIdFromPublicKey;
 
-	const transaction: PartialTransaction = {
-		type: 0,
-		amount: amount.toString(),
+	const transaction = {
+		...createBaseTransaction(inputs),
```
shuse2, line 98: let's move this function in utils

packages/lisk-transactions/src/0_transfer.ts
```diff
@@ -44,81 +30,48 @@ export interface TransferInputs {
...
-			);
-		}
-		if (data.length > BYTESIZES.DATA) {
-			throw new Error('Transaction data field cannot exceed 64 bytes.');
-		}
-	}
-};
-
-export const transfer = (inputs: TransferInputs): TransferTransaction => {
+export const transfer = (inputs: TransferInputs): object => {
```
shuse2: definition of `TransferInputs` is only used here, so let's move it here.
Also, return type would be probably bit better if it's Partial<TransactionJSON>.

packages/lisk-transactions/src/transactions/0_transfer.ts
```diff
@@ -78,7 +77,7 @@ export const transferAssetFormatSchema = {
 	},
 };
 
-const validateInputs = ({
+export const validateInputs = ({
```
shuse2: mentioned above, let's move this to `transfer.ts` above

shuse2 (changes requested):

packages/lisk-transactions/src/0_transfer.ts
pablitovicente: Done. Resolving.

packages/lisk-transactions/src/transactions/0_transfer.ts
pablitovicente: Done. Resolving.

packages/lisk-transactions/src/0_transfer.ts
pablitovicente: Done. Resolving.

packages/lisk-transactions/src/0_transfer.ts
pablitovicente: Both Done. Type changed to `Partial<TransactionJSON>`. Resolving.

packages/lisk-transactions/src/0_transfer.ts
```diff
@@ -54,24 +46,19 @@ const validateInputs = ({
 		throw new Error('Amount must be a valid number in string format.');
 	}
 
-	if (!recipientId && !recipientPublicKey) {
-		throw new Error(
-			'Either recipientId or recipientPublicKey must be provided.',
-		);
+	if (!recipientId) {
```
shuse2: I think original check either recipientId or recipientPublicKey must be present was correct.

packages/lisk-transactions/src/0_transfer.ts
```diff
@@ -88,37 +75,50 @@ const validateInputs = ({
 	}
 };
 
-export const transfer = (inputs: TransferInputs): TransferTransaction => {
+export const transfer = (
+	inputs: TransferInputs,
+): Partial<iTransferTransaction | TransactionJSON> => {
```
shuse2: do we really need `|` here?
maybe just `Partial<TransactionJSON>` is fine?

packages/lisk-transactions/test/1_register_second_passphrase.ts
```diff
@@ -132,7 +135,7 @@ describe('#registerSecondPassphrase transaction', () => {
 			).and.not.be.empty;
 		});
 
-		it('should not have a signSignature property', () => {
+		it.skip('should not have a signSignature property', () => {
```
shuse2: why was it skipped?
maybe checked to be `undefined`?

shuse2 (changes requested):

packages/lisk-transactions/test/1_register_second_passphrase.ts
pablitovicente: Was skipped to get this kind of feedback. 💯
Will check for undefined

packages/lisk-transactions/src/0_transfer.ts
pablitovicente: We can just go for TransactionJSON I thought this way was more descriptive. Will go ahead and leave TransactionJSON only

pablitovicente: Done. Resolving.

packages/lisk-transactions/src/0_transfer.ts
pablitovicente: Done. Resolving.

packages/lisk-transactions/src/0_transfer.ts
```diff
@@ -69,9 +58,8 @@ const validateInputs = ({
 	}
 
 	if (
-		recipientId &&
 		recipientPublicKey &&
```
shuse2, line 62: here recipientId is not inserted yet, so probably need to be checked?

pablitovicente: Fixed. Resolving.

packages/lisk-transactions/test/1_register_second_passphrase.ts
```diff
@@ -132,10 +132,10 @@ describe('#registerSecondPassphrase transaction', () => {
...
 		});
 
-		it.skip('should not have a signSignature property', () => {
-			return expect(registerSecondPassphraseTransaction).not.to.have.property(
-				'signSignature',
-			);
+		it('should have a signSignature property when passphrase is provided', () => {
+			return expect(
+				registerSecondPassphraseTransaction.signSignature,
+			).not.to.be.eql(null);
```
shuse2: probably it's better to expect something, and if we can't then 
`not.to.be.undefined`

packages/lisk-transactions/test/0_transfer.ts
```diff
@@ -138,10 +138,8 @@ describe('#transfer transaction', () => {
 					.and.be.an('object').and.be.empty;
 			});
 
-			it.skip('should not have the second signature property', () => {
-				return expect(transferTransaction).not.to.have.property(
-					'signSignature',
-				);
+			it('second signature property should be undefined', () => {
+				return expect(transferTransaction.signSignature).to.be.eql(undefined);
```
shuse2:
```suggestion
				return expect(transferTransaction.signSignature).to.be.undefined;
```

shuse2 (changes requested):

packages/lisk-transactions/test/1_register_second_passphrase.ts
pablitovicente: Done. Resolving.

packages/lisk-transactions/test/1_register_second_passphrase.ts
```diff
@@ -135,7 +137,7 @@ describe('#registerSecondPassphrase transaction', () => {
 		it('should have a signSignature property when passphrase is provided', () => {
 			return expect(
 				registerSecondPassphraseTransaction.signSignature,
-			).not.to.be.eql(null);
+			).to.be.eql(signSignature);
```
shuse2: this test should be reverted.
and in SecondSignatureTransaction, it should override the `sign` function not to assign signSignature

pablitovicente: Done. Resolving.

shuse2 (dismissed):
LGTM Great work =)

packages/lisk-transactions/src/0_transfer.ts
```diff
@@ -12,28 +12,17 @@
...
 
-const createAsset = (data?: string): TransferAsset => {
-	if (data && data.length > 0) {
-		return { data };
-	}
-
-	return {};
-};
-
 export interface TransferInputs {
```
mitsuaki-u, line 26: maybe can remove `timeoffset` here and in other class inputs now?

packages/lisk-transactions/src/index.ts
```diff
@@ -12,12 +12,12 @@
  * Removal or modification of this copyright notice is prohibited.
  *
  */
+import { transfer } from './0_transfer';
 import { registerMultisignature } from './4_register_multisignature_account';
 import * as constants from './constants';
 import { createSignatureObject } from './create_signature_object';
 import {
```
mitsuaki-u, line 23: I just realized should also include `OutTransferClass` here too

packages/lisk-transactions/src/index.ts
```diff
@@ -47,12 +36,8 @@ export {
 	VoteTransaction,
 	InTransferTransaction,
 	transfer,
-	registerDelegate,
```
mitsuaki-u: we are no longer exporting create functions other than `transfer` and `registerMultisignature`?

packages/lisk-transactions/src/transactions/1_register_second_passphrase.ts
```diff
@@ -330,4 +275,11 @@ export class SecondSignatureTransaction extends BaseTransaction {
 			state: { sender: strippedSender },
 		};
 	}
+
+	public sign(passphrase: string): void {
```
mitsuaki-u, line 279: do we need to set `_signature` to undefined if setting it again below

mitsuaki-u (changes requested):

packages/lisk-transactions/src/transactions/1_register_second_passphrase.ts
shuse2: I think we do because `getBytes` will take into account of the signature if exists

packages/lisk-transactions/src/index.ts
pablitovicente: OutTransferClass is now being exported.

packages/lisk-transactions/src/0_transfer.ts
pablitovicente: Good catch! Removed!

pablitovicente: Actually is being used in createBaseTransaction so I will revert this change.

pablitovicente: Done. We need to keep this as it is part of the inputs and being used by `createBaseTransaction()`

packages/lisk-transactions/src/transactions/1_register_second_passphrase.ts
pablitovicente: Yeah I think we need it as in fact getBytes uses  `this._signature` in a ternary operation. Resolving.

packages/lisk-transactions/src/index.ts
pablitovicente: Done. Resolving.

shuse2 (approved):

mitsuaki-u (approved):

PR #1066, opened 2019-01-31
Need a separate peer info interface for basic and detailed peers info - Closes#1055

### LINKED ISSUE #1055: Need a separate peer info interface for basic and detailed peers info
### Description

We need separate Peer info for basic peer information that we don't have much information yet and a detailed PeerInfo for discovered peers for which we have more information.

ishantiw:
### Description

We need separate Peer info for basic peer information that we don't have much information yet and a detailed PeerInfo for discovered peers for which we have more information.

### Review checklist

* The PR resolves #1055 
* Linting passes
* Tests pass

### DIFF
--- a/packages/lisk-p2p/src/p2p.ts
+++ b/packages/lisk-p2p/src/p2p.ts
@@ -30,8 +30,8 @@ import {
 } from './disconnect_status_codes';
 
 import {
-	DiscoveredPeerInfo,
 	P2PConfig,
+	P2PDiscoveredPeerInfo,
 	P2PMessagePacket,
 	P2PNetworkStatus,
 	P2PNodeInfo,
@@ -220,7 +220,7 @@ export class P2P extends EventEmitter {
 					const wsPort: number = parseInt(queryObject.wsPort, BASE_10_RADIX);
 					const peerId = Peer.constructPeerId(socket.remoteAddress, wsPort);
 
-					const incomingPeerInfo: DiscoveredPeerInfo = {
+					const incomingPeerInfo: P2PDiscoveredPeerInfo = {
 						...queryObject,
 						ipAddress: socket.remoteAddress,
 						wsPort,
--- a/packages/lisk-p2p/src/p2p.ts
+++ b/packages/lisk-p2p/src/p2p.ts
@@ -30,6 +30,7 @@ import {
 } from './disconnect_status_codes';
 
 import {
+	DiscoveredPeerInfo,
 	P2PConfig,
 	P2PMessagePacket,
 	P2PNetworkStatus,
@@ -219,7 +220,7 @@ export class P2P extends EventEmitter {
 					const wsPort: number = parseInt(queryObject.wsPort, BASE_10_RADIX);
 					const peerId = Peer.constructPeerId(socket.remoteAddress, wsPort);
 
-					const incomingPeerInfo: P2PPeerInfo = {
+					const incomingPeerInfo: DiscoveredPeerInfo = {
 						...queryObject,
 						ipAddress: socket.remoteAddress,
 						wsPort,
--- a/packages/lisk-p2p/src/peer.ts
+++ b/packages/lisk-p2p/src/peer.ts
@@ -18,7 +18,7 @@ import * as querystring from 'querystring';
 import { RPCResponseError } from './errors';
 
 import {
-	DiscoveredPeerInfo,
+	P2PDiscoveredPeerInfo,
 	P2PMessagePacket,
 	P2PNodeInfo,
 	P2PPeerInfo,
@@ -78,7 +78,7 @@ export class Peer extends EventEmitter {
 	private readonly _wsPort: number;
 	private readonly _height: number;
 	private _peerInfo: P2PPeerInfo;
-	private _peerDetailedInfo: DiscoveredPeerInfo | undefined;
+	private _peerDetailedInfo: P2PDiscoveredPeerInfo | undefined;
 	private _nodeInfo: P2PNodeInfo | undefined;
 	private _inboundSocket: SCServerSocketUpdated | undefined;
 	private _outboundSocket: SCClientSocket | undefined;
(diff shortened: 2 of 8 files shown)


shuse2 (dismissed):

packages/lisk-p2p/src/p2p.ts
```diff
@@ -30,6 +30,7 @@ import {
 } from './disconnect_status_codes';
 
 import {
+	DiscoveredPeerInfo,
```
jondubois: This interface should be exposed publicly. It should be called `P2PDiscoveredPeerInfo`.

packages/lisk-p2p/src/peer.ts
```diff
@@ -168,19 +170,25 @@ export class Peer extends EventEmitter {
 		this._outboundSocket = scClientSocket;
 	}
 
-	public updatePeerInfo(newPeerInfo: P2PPeerInfo): void {
+	public updatePeerInfo(newPeerInfo: DiscoveredPeerInfo): void {
 		this._peerInfo = {
-			...newPeerInfo,
+			height: newPeerInfo.height,
```
jondubois, line 181: It's correct to remove`...newPeerInfo` (it's risky and not needed anymore), but we should include the `options` field at least.

packages/lisk-p2p/src/peer_pool.ts
```diff
@@ -176,7 +177,14 @@ export class PeerPool extends EventEmitter {
 			return false;
 		}
 
-		this.addPeer({ ...peerInfo }, socket);
+		this.addPeer(
+			{
+				ipAddress: peerInfo.ipAddress,
+				wsPort: peerInfo.wsPort,
+				height: peerInfo.height,
+			},
```
jondubois, line 202: We should pass as many fields as possible. In the case of an inbound connection, we have enough data from the socket to supply a DiscoveredPeerInfo. Inbound peers should be considered as discovered since all their peer info is passed along with the inbound connection.

Not sure what the solution is; maybe we need a separate method like `peerPool.addDiscoveredPeer`?

packages/lisk-p2p/src/peer_pool.ts
```diff
@@ -228,17 +236,32 @@ export class PeerPool extends EventEmitter {
 			// TODO ASAP: We need a new type to account for complete P2PPeerInfo which has all possible fields (e.g. P2PDiscoveredPeerInfo) that way we don't need to have all these checks below.
 			peers: this._pickRandomPeers(MAX_PEER_LIST_BATCH_SIZE).map(
 				(peer: Peer) => {
+					const peerDetailedInfo = peer.detailedPeerInfo;
 					const peerInfo = peer.peerInfo;
 
+					if (peerDetailedInfo) {
```
jondubois: Every peer passed to this map callback should always have `peer.detailedPeerInfo` because `_pickRandomPeers` already checked `peer.peerInfo.isDiscoveredPeer` before. Ideally the type system should ensure this. The `_handleGetAllPeersRequest` method should never send back a undiscovered peers.

- Maybe you can change the `_pickRandomPeers` function to filter based on the existence of `peer.detailedPeerInfo` instead of `peer.peerInfo.isDiscoveredPeer`; the two should always we set together.

packages/lisk-p2p/src/peer_pool.ts
```diff
@@ -228,17 +236,32 @@ export class PeerPool extends EventEmitter {
...
+							version: peerDetailedInfo.version ? peerDetailedInfo.version : '',
+							wsPort: String(peerInfo.wsPort),
+						};
+					}
+
 					return {
-						broadhash: peerInfo.options
-							? (peerInfo.options.broadhash as string)
-							: '',
+						broadhash: '',
```
jondubois: As mentioned in previous comment, no need to include this in the list unless we have complete discovered peer info; ideally, this case should never happen.

jondubois (changes requested):

packages/lisk-p2p/src/peer.ts
ishantiw: `options` field is being captured and maintained by `_detailedPeerInfo`, otherwise it will get redundant.

packages/lisk-p2p/src/peer_pool.ts
ishantiw: Actually, `_pickRandomPeers` is iterating through Peer object and gives out a list of peer objects filtering `isDiscovered`, but Typescript doesn't understand if `isDiscoveredPeer` is true then `detailedpeerInfo` will be available. So in the end, I have to check for this condition before using it.

packages/lisk-p2p/src/peer_pool.ts
ishantiw: I had to return it because of the above comment, I had to keep a default return statement

packages/lisk-p2p/src/peer_pool.ts
ishantiw: yeah this is a bit tricky, I guess we should have separate method to add peer in peer_pool because it will get complicated to handle two types in `addPeer()`.
So if we will have a separate `addDiscoveredPeer` then we can explicitly call `updatePeerInfo`to update `detailedPeerInfo` of Peer as its constructor only accepts `P2PPeerInfo`

packages/lisk-p2p/src/peer_pool.ts
jondubois: In that case, maybe it can return null and then we can add a `filter` function to remove all the null references from the mapped list? Or we can use filter and map (to typecast) before so that we are sure that this function only handles `DiscoveredPeerInfo` objects.

We should never send back a peer which has not been discovered yet so the code should not make any effort to support this.

jondubois (changes requested):

shuse2 (approved):

jondubois (approved):

PR #1067, opened 2019-01-31
Change selectPeers function to only handle detailed PeerInfo objects; not live Peer objects Closes#1059

### LINKED ISSUE #1059: Change selectPeers function to only handle detailed PeerInfo objects; not live Peer objects
### Expected behavior

We should make the `selectPeers` function in `peer_selection.ts` work with **detailed** peer info objects instead of live `Peer` objects.

### Actual behavior

Currently, the `selectPeers` function accepts and returns a list of `Peer` objects (`ReadonlyArray<Peer>`).

note that this issue depends on https://github.com/LiskHQ/lisk-elements/issues/1055 - We need to separate the basic peer info from detailed peer info before this can be implemented.

ishantiw:
### Description

Currently, the `selectPeers` function accepts and returns a list of `Peer` objects (`ReadonlyArray<Peer>`). We should make the `selectPeers` function in `peer_selection.ts` work with detailed peer info objects instead of live Peer objects.

### Review checklist

* The PR resolves #1059
* Linting passes
* Tests pass

### DIFF
--- a/packages/lisk-p2p/src/p2p.ts
+++ b/packages/lisk-p2p/src/p2p.ts
@@ -170,13 +170,13 @@ export class P2P extends EventEmitter {
 
 	// TODO ASAP: Change selectPeers to return PeerInfo list and then make request on peerPool itself; pass peerInfo as argument.
 	public async request(packet: P2PRequestPacket): Promise<P2PResponsePacket> {
-		const response = await this._peerPool.requestPeer(packet, this._nodeInfo);
+		const response = await this._peerPool.requestPeer(packet);
 
 		return response;
 	}
 
 	public send(message: P2PMessagePacket): void {
-		this._peerPool.sendToPeers(message, this._nodeInfo);
+		this._peerPool.sendToPeers(message);
 	}
 
 	private async _startPeerServer(): Promise<void> {
--- a/packages/lisk-p2p/src/p2p.ts
+++ b/packages/lisk-p2p/src/p2p.ts
@@ -22,9 +22,7 @@ interface SCServerUpdated extends SCServer {
 	readonly isReady: boolean;
 }
 
-import { RequestFailError } from './errors';
-import { constructPeerId, constructPeerIdFromPeerInfo, Peer } from './peer';
-import { PeerOptions } from './peer_selection';
+import { constructPeerId, constructPeerIdFromPeerInfo } from './peer';
 
 import {
 	INVALID_CONNECTION_QUERY_CODE,
--- a/packages/lisk-p2p/src/p2p.ts
+++ b/packages/lisk-p2p/src/p2p.ts
@@ -170,30 +170,13 @@ export class P2P extends EventEmitter {
 
 	// TODO ASAP: Change selectPeers to return PeerInfo list and then make request on peerPool itself; pass peerInfo as argument.
 	public async request(packet: P2PRequestPacket): Promise<P2PResponsePacket> {
-		const peerSelectionParams: PeerOptions = {
-			lastBlockHeight: this._nodeInfo.height,
-		};
-		const selectedPeer = this._peerPool.selectPeers(peerSelectionParams, 1);
-
-		if (selectedPeer.length <= 0) {
-			throw new RequestFailError(
-				'Request failed due to no peers found in peer selection',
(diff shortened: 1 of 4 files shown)


packages/lisk-p2p/src/peer.ts
```diff
@@ -365,6 +365,10 @@ export class Peer extends EventEmitter {
 		inboundSocket.off(REMOTE_EVENT_MESSAGE, this._handleRawMessage);
 	}
 
+	public static constructPeerIdFromPeerInfo(peerInfo: P2PPeerInfo): string {
```
shuse2, line 381: maybe just export as a function?

shuse2 (changes requested):

packages/lisk-p2p/src/p2p.ts
```diff
@@ -164,30 +162,13 @@ export class P2P extends EventEmitter {
...
-		};
-		const selectedPeer = this._peerPool.selectPeers(peerSelectionParams, 1);
-
-		if (selectedPeer.length <= 0) {
-			throw new RequestFailError(
-				'Request failed due to no peers found in peer selection',
-			);
-		}
-		const response: P2PResponsePacket = await selectedPeer[0].request(packet);
+		const response = await this._peerPool.requestPeer(packet, this._nodeInfo);
```
jondubois: We shouldn't pass `this._nodeInfo` as second argument. We should only support a single way of updating the nodeInfo; using `applyNodeInfo`.

Because the `PeerPool` now holds a reference to the latest `_nodeInfo` state internally, you can just use `this. _nodeInfo` from inside the `PeerPool` instead.

packages/lisk-p2p/src/p2p.ts
```diff
@@ -164,30 +162,13 @@ export class P2P extends EventEmitter {
...
 	public send(message: P2PMessagePacket): void {
-		const peerSelectionParams: PeerOptions = {
-			lastBlockHeight: this._nodeInfo.height,
-		};
-		const selectedPeers = this._peerPool.selectPeers(peerSelectionParams);
-
-		selectedPeers.forEach((peer: Peer) => {
-			peer.send(message);
-		});
+		this._peerPool.sendToPeers(message, this._nodeInfo);
```
jondubois: Similar as previous comment.

packages/lisk-p2p/src/peer.ts
jondubois: If we export as a function then we should do the same for the `Peer.constructPeerId(address, port)` which is now a static method.

Or we can do this in a separate PR because I also rely on the same static method right now for my next PR.

packages/lisk-p2p/src/peer_pool.ts
```diff
@@ -108,16 +111,64 @@ export class PeerPool extends EventEmitter {
...
 
 		return selectedPeers;
 	}
 
+	public async requestPeer(
+		packet: P2PRequestPacket,
+		nodeInfo: P2PNodeInfo,
+	): Promise<P2PResponsePacket> {
+		const peerSelectionParams: PeerOptions = {
+			lastBlockHeight: nodeInfo.height,
```
jondubois: As mentioned in previous comment, you can just use `this._nodeInfo` instead of getting the nodeInfo as a second argument to this function.

packages/lisk-p2p/src/peer_pool.ts
```diff
@@ -108,16 +111,64 @@ export class PeerPool extends EventEmitter {
...
+				`No such Peer exist in PeerPool with the selected peer with Id: ${selectedPeerId}`,
+			);
+		}
+
+		const response: P2PResponsePacket = await peer.request(packet);
+
+		return response;
+	}
+
+	public sendToPeers(message: P2PMessagePacket, nodeInfo: P2PNodeInfo): void {
```
jondubois: Same as previous comment about nodeInfo.

jondubois (changes requested):

ishantiw:
comments are addressed

shuse2 (dismissed):

shuse2 (approved):

jondubois (approved):

PR #1075, opened 2019-02-05
Update P2P header should include options - Closes #1071

### LINKED ISSUE #1071: Update P2P header should include options
### Expected behavior
When NodeInfo.options is provided to constructor, it should send the `nodeInfo.Options` as query.

### Actual behavior
query on the socket connection is empty

jondubois:
### What was the problem?

The `querystring.stringify()` function which was used to stringify `P2PNodeInfo` objects could not handle the nested `options` object so it was removing it.

### How did I fix it?

Used `JSON.stringify` to stringify the nested `options` object; that way it can be passed as a query parameter.

### How to test it?

`npm run test:integration`

Note that this affects the protocol.

### Review checklist

* The PR resolves #1071
* Linting passes
* Tests pass

### DIFF
--- a/packages/lisk-p2p/src/p2p.ts
+++ b/packages/lisk-p2p/src/p2p.ts
@@ -215,12 +215,14 @@ export class P2P extends EventEmitter {
 					const peerId = constructPeerId(socket.remoteAddress, wsPort);
 
 					const incomingPeerInfo: P2PDiscoveredPeerInfo = {
-						...queryObject,
 						ipAddress: socket.remoteAddress,
 						wsPort,
 						height: queryObject.height ? +queryObject.height : 0,
+						isDiscoveredPeer: true,
 						os: queryObject.os,
 						version: queryObject.version,
+						options: typeof queryObject.options === 'string' ?
+							JSON.parse(queryObject.options) : undefined,
 					};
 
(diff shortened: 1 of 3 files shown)


packages/lisk-p2p/src/p2p.ts
```diff
@@ -215,12 +215,14 @@ export class P2P extends EventEmitter {
 					const peerId = constructPeerId(socket.remoteAddress, wsPort);
 
 					const incomingPeerInfo: P2PDiscoveredPeerInfo = {
-						...queryObject,
 						ipAddress: socket.remoteAddress,
 						wsPort,
 						height: queryObject.height ? +queryObject.height : 0,
+						isDiscoveredPeer: true,
```
shuse2, line 221: is this `true`? it sounds like if it's incoming connection, it should be `false`?

jondubois: This is correct because the `isDiscoveredPeer` property of a `P2PDiscoveredPeerInfo` should always be true. It allows us to distinguish complete from incomplete peer info at runtime.

shuse2: maybe change the name in other issue to make it clear?

shuse2 (approved):

PR #1078, opened 2019-02-06
Fix compatibility issues with Core v1.x version - Closes #1077

### LINKED ISSUE #1077: Fix compatibility issues with Core v1.x version
### Expected behavior

- The P2P library should be able to connect to the existing Lisk core testnet and communicate with the nodes as a v1 lisk-core node would.
- A socket error in the P2P library should be handled and emitted as an event so that it does not crash the P2P node.

### Actual behavior

- The P2P library has trouble connecting and interacting with the current Lisk core version. There are issues with missing fields which are causing handshake failure.
- If one of the peers causes socket failure, it crashes the P2P library.

### Steps to reproduce

Check the branch `add_p2p_sample_script`

jondubois:
### What was the problem?

There were some remaining issues related to incompatible fields which prevented the P2P library from connecting and interacting with lisk-core v1.x.x nodes.

### How did I fix it?

- Fixed fields to match those in the current v1 protocol exactly.
- Handled socket errors and emit them as events so that they do not crash the P2P node.
- Disable ping timeouts since they are not needed for our use case.
- Do not force the `os` field to be an alpha-only string. Often they have dots, dashes and version numbers in them which is OK.

### How to test it?

- `npm run test`
- Use the `sample/index` script in `add_p2p_sample_script` branch to test against v1 testnet.

### Review checklist

* The PR resolves #1077
* Linting passes
* Tests pass


packages/lisk-p2p/src/validation.ts
```diff
@@ -67,7 +67,7 @@ export const validatePeerInfo = (
 	const version = protocolPeer.version;
 	const wsPort = +protocolPeer.wsPort;
 	const os =
-		protocolPeer.os && isAlpha(protocolPeer.os.toString())
+		protocolPeer.os
```
shuse2, line 70: should't we check `isAlpha` here?

packages/lisk-p2p/src/peer.ts
```diff
@@ -83,9 +94,9 @@ export const constructPeerIdFromPeerInfo = (peerInfo: P2PPeerInfo): string =>
 const convertNodeInfoToLegacyFormat = (nodeInfo: P2PNodeInfo): ProtocolNodeInfo => (
 	{
 		...nodeInfo,
-		wsPort: nodeInfo.wsPort,
+		httpPort: nodeInfo.options ? nodeInfo.options.httpPort as number : 0,
 		broadhash: nodeInfo.options ? nodeInfo.options.broadhash as string : '',
-		nonce: nodeInfo.options ? nodeInfo.options.nonce as string : '',
+		nonce: nodeInfo.options ? nodeInfo.options.nonce as string : ''
```
shuse2: let's' keep ending `,`

shuse2 (changes requested):

packages/lisk-p2p/src/validation.ts
jondubois: Right now in core v1, the os string contains dots and dashes and version numbers so alpha is not suitable.

shuse2 (approved):

PR #1083, opened 2019-02-07
Receiving blocks, transactions and signatures from v1 lisk core - Closes #1081

### LINKED ISSUE #1081: Receiving blocks, transactions and signatures from v1 lisk core
### Expected behavior

P2P library should be able to receive blocks, transactions and signatures from v1 core nodes.

### Actual behavior

The node is not receiving blocks, transactions or signatures from v1 core testnet.

### Steps to reproduce

Run a sample P2P node on a remote network (since we need to handle inbound sockets).

jondubois:
### What was the problem?

Inbound messages such as blocks, transactions and signatures were not received.

### How did I fix it?

Added legacy event handlers to handle the three legacy cases.

### How to test it?

Run node on remote testnet.

### Review checklist

* The PR resolves #1081
* Linting passes
* Tests pass

### DIFF
--- a/packages/lisk-p2p/src/peer.ts
+++ b/packages/lisk-p2p/src/peer.ts
@@ -239,10 +239,10 @@ export class Peer extends EventEmitter {
 		};
 
 		this._peerDetailedInfo = {
+			...this._peerInfo,
 			os: newPeerInfo.os,
 			version: newPeerInfo.version,
 			options: newPeerInfo.options,
-			...this._peerInfo,
 		};
 	}
 
--- a/packages/lisk-p2p/src/peer.ts
+++ b/packages/lisk-p2p/src/peer.ts
@@ -115,6 +115,9 @@ export class Peer extends EventEmitter {
 		respond: (responseError?: Error, responseData?: unknown) => void,
 	) => void;
 	private readonly _handleRawMessage: (packet: unknown) => void;
+	private readonly _handleRawLegacyMessagePostBlock: (packet: unknown) => void;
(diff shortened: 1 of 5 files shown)


packages/lisk-p2p/src/peer.ts
```diff
@@ -211,7 +237,11 @@ export class Peer extends EventEmitter {
 			isDiscoveredPeer: true,
 		};
 
-		this._peerDetailedInfo = newPeerInfo;
+		// Some fields from the main _peerInfo cannot be overridden.
+		this._peerDetailedInfo = {
```
shuse2, line 241: maybe explicitly assigning what needs to be assign is more clear what cannot be overridden?

shuse2 (changes requested):

packages/lisk-p2p/src/peer.ts
```diff
@@ -293,10 +323,20 @@ export class Peer extends EventEmitter {
...
 		}
-		this._outboundSocket.emit(REMOTE_EVENT_MESSAGE, {
-			event: packet.event,
-			data: packet.data,
-		});
+		// TODO later: Legacy events will no longer be required after migrating to the LIP protocol version.
+		if (
+			packet.event === 'postBlock' || 
+			packet.event === 'postTransactions' || 
+			packet.event === 'postSignatures'
```
shuse2: maybe below, because lint is failing here
```
const legacyEvents = ['postBlock', 'postTransactions', 'postSignatures'];
if (legacyEvents.includes(packet.event))
```

shuse2 (changes requested):

shuse2 (approved):

PR #1093, opened 2019-02-11
Add get or default to explicitly get the default account

### LINKED ISSUE #1091: Add getOrDefault for StateStore.account
### Expected behavior
User should know when to initialize the account.
Also, Account.publicKey should be optional

### Actual behavior
StateStore.account.get expects to create account if doesn't exist

shuse2:
### What was the problem?
It was ambiguous when account get initialized by default value.

### How did I fix it?
Add getOrDefault function from store

### How to test it?
`npm t`

### Review checklist

* The PR resolves #1091 
* Linting passes
* Tests pass


mitsuaki-u (approved):
should it be named `getOrCreate`?

SargeKhan (approved):

shuse2:
@mitsujutsu I think user cannot `create`, so default in this case would be better

PR #1095, opened 2019-02-11
Add discovery process - Closes #900

### LINKED ISSUE #900: Add discovery process

jondubois:
### What was the problem?

Peer discovery was only running once. It needs to run on an interval to guarantee full discovery of the network.

### How did I fix it?

- Made discovery run on a configurable interval.
- Minor code cleanup.
- Added new integration test case to check that all peers eventually get discovered.

### How to test it?

```
npm run test:integration
```

### Review checklist

* The PR resolves #900 
* Linting passes
* Tests pass


packages/lisk-p2p/src/peer_pool.ts
```diff
@@ -142,7 +146,7 @@ export class PeerPool extends EventEmitter {
 		return selectedPeers;
 	}
 
-	public async requestPeer(
+	public async requestFromPeer(
```
shuse2, line 149: why change the name to `requestFromPeer`? `requestPeer` sounds more straight forward?

shuse2 (dismissed):

packages/lisk-p2p/src/peer_pool.ts
jondubois: - The other method is called 'sendToPeer', so 'requestFromPeer' sounds more consistent.

- requestPeer is a bit confusing because someone who doesn't know the code could expect the method to return a Peer object.

I guess the second point is not a big deal with TypeScript though but I think we should be consistent with the naming. If we call it 'requestPeer' then maybe we should call the other related method 'sendPeer'?

shuse2 (dismissed):

shuse2 (approved):

PR #1101, opened 2019-02-13
Add publicKey in base transaction - Closes #1099

### LINKED ISSUE #1099: Add publicKey from sender if it doesn't exist in account
### Expected behavior
if sender does not have publickey, senderPublicKey should be inserted in the base_transaction.
Also, when we get the sender in the base_transaction, it should use getOrDefault

### Actual behavior
it doesn't insert the public key when receive sender without public key

mitsuaki-u:
### What was the problem?

In case of defaulting to new account via `getOrDefault`, base transactions needs to set `publicKey` for the sender. 

### How did I fix it?

Add `publicKey` to sender state in apply and undo. 

### How to test it?

npm test

### Review checklist

* The PR resolves #1099 
* Linting passes
* Tests pass

### DIFF
--- a/packages/lisk-transactions/src/0_transfer_transaction.ts
+++ b/packages/lisk-transactions/src/0_transfer_transaction.ts
@@ -46,7 +46,6 @@ export const transferAssetTypeSchema = {
 
 export const transferAssetFormatSchema = {
 	type: 'object',
-	additionalProperties: false,
 	properties: {
 		data: {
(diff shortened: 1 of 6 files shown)


shuse2 (commented):
Changes looks good but test are failing ='(

shuse2 (approved):

SargeKhan (approved):

PR #1102, opened 2019-02-13
Add check for additional asset fields - Closes #1090

### LINKED ISSUE #1090: Add no additional asset properties check for transaction asset field
### Expected behavior

Should not allow additional fields in asset field during schema format check.

### Actual behavior

No check in place.

version 2.1.0

mitsuaki-u:
### What was the problem?

There was no schema check to prevent additional asset fields.

### How did I fix it?

Added ajv schema check. 

### How to test it?

npm test

### Review checklist

* The PR resolves #1090 
* Linting passes
* Tests pass

### DIFF
--- a/packages/lisk-transactions/src/7_out_transfer_transaction.ts
+++ b/packages/lisk-transactions/src/7_out_transfer_transaction.ts
@@ -35,6 +35,7 @@ export interface OutTransferAsset {
 
 export const outTransferAssetTypeSchema = {
 	type: 'object',
+	additionalProperties: false,
 	required: ['outTransfer'],
 	properties: {
 		outTransfer: {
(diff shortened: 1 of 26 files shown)


packages/lisk-transactions/src/7_out_transfer_transaction.ts
```diff
@@ -54,6 +55,7 @@ export const outTransferAssetTypeSchema = {
 
 export const outTransferAssetFormatSchema = {
 	type: 'object',
+	additionalProperties: false,
```
shuse2: if we set this to `all` I think we don't need tis `additionalProperties: false`

shuse2 (dismissed):

shuse2 (approved):

PR #1104, opened 2019-02-14
Disconnect old peers, cleanup node and improve timeouts - Closes #1086

### LINKED ISSUE #1086: Disconnect old peers, cleanup node and improve timeouts
### Expected behavior

Disconnect old peers which are no longer active.

### Actual behavior

Currently peers are never disconnected.

jondubois:
### What was the problem?

- There was no mechanism to fully disconnect from peers which could not be reached.
- There was an issue in the integration tests which meant that P2P node would not aways stop correctly.
- There was no mechanism for configuring timeout values on the P2P node.

### How did I fix it?

- If a peer cannot be reached from our outbound connection, we remove them from our PeerPool. It's possible that the node may try again at a later time when the next round of discovery will run; this is configurable by the user.
- Fixed a bug related to a missing option passed to the client socket which means that sockets were being reused and event handlers were accumulating on them.
- Added an new `ackTimeout` to allow the user to specify an upper bound on how long it should take to perform requests (including discovery).
- Cleanup issue related to the P2P stop() method not resolving in some scenarios.
- Added an integration test case to verify that stopping one of the P2P nodes in the network causes other peers to remove that node from their PeerPool.

**EDIT** (after PR feedback)
- The `isDiscoveredPeer` property is now compulsory on the `P2PDiscoveredPeerInfo` interface because we need to always differentiate between these two objects internally.
- The P2P node now makes a `peer.fetchStatus()` 'status' call after connecting on the outbound connection; we need to make this call in order to get the full `P2PDiscoveredPeerInfo` from the outbound connection.
- Changed `_triedPeers` to be of type `Map<string, P2PDiscoveredPeerInfo>` (only the full discovered peers info is allowed); we should not have undiscovered peers in there as it will cause issues in other parts of the logic.
- `_triedPeers` is now populated using a new event on the `PeerPool` called `EVENT_DISCOVERED_PEER`; this event carries the full `P2PDiscoveredPeerInfo`
- Improved the solution for the P2P stop() method not resolving.

### How to test it?

`npm run test:integration`

### Review checklist

* The PR resolves #1086 
* Linting passes
* Tests pass

### DIFF
--- a/packages/lisk-p2p/src/p2p.ts
+++ b/packages/lisk-p2p/src/p2p.ts
@@ -36,6 +36,7 @@ import {
 import { PeerInboundHandshakeError } from './errors';
 
 import {
+	P2PClosePacket,
 	P2PConfig,
 	P2PDiscoveredPeerInfo,
 	P2PMessagePacket,
@@ -51,6 +52,7 @@ import { P2PRequest } from './p2p_request';
 export { P2PRequest };
 
 import {
+	EVENT_CLOSE_OUTBOUND,
 	EVENT_CONNECT_ABORT_OUTBOUND,
 	EVENT_CONNECT_OUTBOUND,
 	EVENT_FAILED_TO_PUSH_NODE_INFO,
@@ -62,6 +64,9 @@ import {
 } from './peer_pool';
 
 export {
+	EVENT_CLOSE_OUTBOUND,
+	EVENT_CONNECT_ABORT_OUTBOUND,
+	EVENT_CONNECT_OUTBOUND,
 	EVENT_REQUEST_RECEIVED,
 	EVENT_MESSAGE_RECEIVED,
 	EVENT_OUTBOUND_SOCKET_ERROR,
@@ -95,6 +100,7 @@ export class P2P extends EventEmitter {
 	private readonly _handleFailedToPushNodeInfo: (error: Error) => void;
 	private readonly _handlePeerConnect: (peerInfo: P2PPeerInfo) => void;
 	private readonly _handlePeerConnectAbort: (peerInfo: P2PPeerInfo) => void;
+	private readonly _handlePeerClose: (closePacket: P2PClosePacket) => void;
 	private readonly _handleOutboundSocketError: (error: Error) => void;
 	private readonly _handleInboundSocketError: (error: Error) => void;
 
@@ -126,15 +132,21 @@ export class P2P extends EventEmitter {
 			if (!this._triedPeers.has(peerId)) {
 				this._triedPeers.set(peerId, peerInfo);
 			}
-			this.emit(EVENT_CONNECT_OUTBOUND);
+			this.emit(EVENT_CONNECT_OUTBOUND, peerInfo);
 		};
+
 		this._handlePeerConnectAbort = (peerInfo: P2PPeerInfo) => {
 			// Re-emit the message to allow it to bubble up the class hierarchy.
 			const peerId = constructPeerIdFromPeerInfo(peerInfo);
 			if (this._triedPeers.has(peerId)) {
 				this._triedPeers.delete(peerId);
 			}
-			this.emit(EVENT_CONNECT_ABORT_OUTBOUND);
+			this.emit(EVENT_CONNECT_ABORT_OUTBOUND, peerInfo);
+		};
+
+		this._handlePeerClose = (closePacket: P2PClosePacket) => {
+			// Re-emit the message to allow it to bubble up the class hierarchy.
+			this.emit(EVENT_CLOSE_OUTBOUND, closePacket);
 		};
 
 		this._handleFailedToPushNodeInfo = (error: Error) => {
@@ -324,7 +336,6 @@ export class P2P extends EventEmitter {
 	private async _stopHTTPServer(): Promise<void> {
 		return new Promise<void>(resolve => {
 			this._httpServer.close(() => {
-				this._isActive = false;
 				resolve();
 			});
 		});
@@ -333,7 +344,6 @@ export class P2P extends EventEmitter {
 	private async _stopWSServer(): Promise<void> {
 		return new Promise<void>(resolve => {
 			this._scServer.close(() => {
-				this._isActive = false;
 				resolve();
 			});
 		});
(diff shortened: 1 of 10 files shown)


packages/lisk-p2p/src/p2p.ts
```diff
@@ -321,27 +336,10 @@ export class P2P extends EventEmitter {
...
-				resolve();
-			});
-		});
-	}
-
-	private async _stopPeerServer(): Promise<void> {
-		await this._stopWSServer();
-		await this._stopHTTPServer();
+	private _stopPeerServer(): void {
+		this._scServer.close();
```
shuse2: we don't need to wait for the callback to make sure?

packages/lisk-p2p/src/p2p.ts
```diff
@@ -126,15 +132,21 @@ export class P2P extends EventEmitter {
 			if (!this._triedPeers.has(peerId)) {
 				this._triedPeers.set(peerId, peerInfo);
 			}
-			this.emit(EVENT_CONNECT_OUTBOUND);
+			this.emit(EVENT_CONNECT_OUTBOUND, peerInfo);
```
ishantiw, line 138: If this will be emitted, then the user will know which peer we are making an outbound connection with. Do you think its okay or we should not emit this info?

packages/lisk-p2p/src/peer.ts
```diff
@@ -441,15 +464,16 @@ export class Peer extends EventEmitter {
 		});
 
 		outboundSocket.on('connect', () => {
-			this.emit(EVENT_CONNECT_OUTBOUND, this._peerInfo);
+			this.emit(EVENT_CONNECT_OUTBOUND, this);
```
ishantiw: why we are emitting the whole peer object instead of peerInfo?

packages/lisk-p2p/src/peer.ts
```diff
@@ -441,15 +464,16 @@ export class Peer extends EventEmitter {
...
 		});
 
 		outboundSocket.on('connectAbort', () => {
-			this.emit(EVENT_CONNECT_ABORT_OUTBOUND, this._peerInfo);
+			this.emit(EVENT_CONNECT_ABORT_OUTBOUND, this);
```
ishantiw: same here

packages/lisk-p2p/src/peer_pool.ts
```diff
@@ -64,24 +67,33 @@ export {
...
 };
 
+interface PeerPoolConfig {
+	readonly connectTimeout?: number;
+	readonly ackTimeout?: number;
+}
+
 export const MAX_PEER_LIST_BATCH_SIZE = 100;
 export const MAX_PEER_DISCOVERY_PROBE_SAMPLE_SIZE = 100;
+export const MAX_PEER_CONNECT_ATTEMPTS = 100;
```
ishantiw: do you think that this number should be smaller? like 5 connect attempts or less?

According to LIP,

> If a connection attempt fails, a different peer is selected by the procedure above. Peers in the collection triedPeers are moved to the collection newPeers after 3 consecutive failed attempts (i.e., the peer is selected randomly 3 times and every time the connection failed).

jondubois: Ah I meant to remove `MAX_PEER_CONNECT_ATTEMPTS` - It's not used in the code. Right now it removed the peer from the peer pool if it fails a single attempt. It will always be able to discover later so it's fine. We should do this for the LIP phase definitely.

packages/lisk-p2p/src/p2p.ts
jondubois: Ah yes we did. I confused this issue for something else. There was a legitimate reason why it wasn't shutting down properly.

packages/lisk-p2p/src/p2p.ts
jondubois: We can expose it for doing analysis and integration tests but I agree that it's not a very important event. We could leave it out of the main documentation.

packages/lisk-p2p/src/p2p.ts
```diff
@@ -342,15 +372,29 @@ export class P2P extends EventEmitter {
...
+			return;
+		}
 
 		const discoveredPeers = await this._peerPool.runDiscovery(
 			allKnownPeers,
 			this._config.blacklistedPeers,
 		);
+
+		// Make sure that we do not try to connect to peers if the P2P node is no longer active.
+		if (!this._isActive) {
```
shuse2, line 394: it's not enough to check above?

shuse2:
@jondubois Also, it's failing on Jenkins with
```
@liskhq/lisk-p2p: test/unit/peer_discovery.ts(54,52): error TS2345: Argument of type 'P2PPeerInfo[]' is not assignable to parameter of type 'ReadonlyArray<P2PDiscoveredPeerInfo>'.

@liskhq/lisk-p2p:   Property 'version' is missing in type 'P2PPeerInfo' but required in type 'P2PDiscoveredPeerInfo'.
```

packages/lisk-p2p/src/p2p.ts
jondubois: This is intentional because there are two lines with `await`. Each time we await the state could change asynchronously behind the scenes. It's a rare edge case that should be covered.

jondubois: I changed the second comment that way it doesn't look like an accidental copy-paste mistake.

shuse2 (approved):

packages/lisk-p2p/src/p2p_types.ts
```diff
@@ -66,9 +67,16 @@ export interface P2PNodeInfo {
 	readonly options?: P2PInfoOptions;
 }
 
+export interface P2PClosePacket {
+	readonly peerInfo: P2PPeerInfo,
```
ishantiw, line 71: I think at this point, peer which is getting closed will have `P2PDiscoveredPeerInfo` so maybe we can use that?

ishantiw (commented):
Looks good, just one minor comment

packages/lisk-p2p/src/p2p_types.ts
jondubois: I also thought this before but it is not guaranteed because we don't get the peer's full `P2PDiscoveredPeerInfo` until after we successfully made a `peer.fetchStatus` call to them. If the socket disconnects before we get the response from the `peer.fetchStatus` call, then we won't have the full `P2PDiscoveredPeerInfo`.

ishantiw (approved):

PR #1107, opened 2019-02-14
Remove duplicate balance error - Closes #1105

### LINKED ISSUE #1105: Transaction returns incorrect error when sender does not have enough balance
### Expected behavior
Not enough balance error should reflect the original balance

### Actual behavior
Balance is shown for subtracted value.
This applies to both BaseTransaction and TransferTransaction
```
Account does not have enough LSK: 123L, balance: -0.1
```

### Steps to reproduce
1. Create a new TransferTransaction
2. input a sender who doesn't have enough balance for fee, or who does have balance for fee but not for amount
3. observe the error

2.1-alpha.0

mitsuaki-u:
### What was the problem?

`verifyBalance` called in Base (for fee check) and in Transfer (and other `amount` transactions) causing duplicate insufficient balance errors.

### How did I fix it?

In Transfer, InTransfer, and OutTransfer: added extra check to make sure balance had at least enough funds for the fee, before returning balance error for `amount` to avoid duplicate errors. 

### How to test it?

npm test

### Review checklist

* The PR resolves #1105 
* Linting passes
* Tests pass

### DIFF
--- a/packages/lisk-transactions/src/0_transfer_transaction.ts
+++ b/packages/lisk-transactions/src/0_transfer_transaction.ts
@@ -179,9 +179,9 @@ export class TransferTransaction extends BaseTransaction {
 	protected applyAsset(store: StateStore): ReadonlyArray<TransactionError> {
 		const errors: TransactionError[] = [];
 		const sender = store.account.get(this.senderId);
-
 		const balanceError = verifyBalance(this.id, sender, this.amount);
-		if (balanceError) {
+		// Only return error if account has enough funds for fee to avoid duplicate errors
+		if (balanceError && new BigNum(sender.balance).gte(0)) {
 			errors.push(balanceError);
 		}
 		const updatedSenderBalance = new BigNum(sender.balance).sub(this.amount);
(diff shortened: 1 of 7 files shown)


packages/lisk-transactions/src/0_transfer_transaction.ts
```diff
@@ -179,9 +179,9 @@ export class TransferTransaction extends BaseTransaction {
 	protected applyAsset(store: StateStore): ReadonlyArray<TransactionError> {
 		const errors: TransactionError[] = [];
 		const sender = store.account.get(this.senderId);
-
 		const balanceError = verifyBalance(this.id, sender, this.amount);
-		if (balanceError) {
+		// Only return error if account has enough funds for fee to avoid duplicate errors
+		if (balanceError && new BigNum(sender.balance).gte(0)) {
```
shuse2: maybe check if the sender balance first before executing `verifyBalance`, and we could add simple function in utils/verify.ts to capture repeating `new BigNum(sender.balance).gte(0)`?
Also, this wouldn't solve the problem with the error message when sender has fee, but not amount case ='(.

shuse2 (changes requested):

shuse2 (approved):

PR #1109, opened 2019-02-14
Validate null character - Closes #1108

### LINKED ISSUE #1108: Disallow null character on transfer and dapp transaction
### Expected behavior
Expect Dapp registration transaction to be rejected if using unicode null character (`\u0000`) in `tag` or `description` attributes.

### Actual behavior
There is no check regards to the null character in Dapp registration

### Steps to reproduce
Post a Dapp registration transaction with `\u0000` in the `description` or `tag` attribute

Related to https://github.com/LiskHQ/lisk/issues/2886

2.1-alpha.0

shuse2:
### What was the problem?
By protocol, Transfer Transaction and Dapp transaction should not allow null character in string field.

### How did I fix it?
Add the same check done as https://github.com/LiskHQ/lisk/commit/5c7e2944383bf369f1343fba62c71e84de6f9664.
Also, added test to check schema validation works.

### How to test it?
`npm t`

### Review checklist

* The PR resolves #1108 
* Linting passes
* Tests pass

### DIFF
--- a/packages/lisk-transactions/test/utils/validation/validation.ts
+++ b/packages/lisk-transactions/test/utils/validation/validation.ts
@@ -29,6 +29,7 @@ import {
 	isGreaterThanMaxTransactionId,
 	isNumberString,
 	isValidInteger,
+	isNullByteIncluded,
 } from '../../../src/utils/validation/validation';
 
 describe('validation', () => {
@@ -367,6 +368,20 @@ describe('validation', () => {
 
 		it('should return true when negative integer was provided', () => {
 			return expect(isValidInteger(-6)).to.be.true;
+		});
+	});
+
+	describe('#isNullByteIncluded', () => {
+		const nullCharacterList = ['\0', '\x00', '\u0000', '\\U00000000'];
+
+		it('should return false when string was provided', () => {
+			return expect(isNullByteIncluded('1234')).to.be.false;
+		});
+
+		it('should return true using unicode null characters', () => {
+			nullCharacterList.forEach(nullChar => {
+				expect(isNullByteIncluded(`${nullChar} hey :)`)).to.be.true;
+			});
 		});
 	});
 });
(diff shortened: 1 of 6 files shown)


lsilvs (commented):
Should `noNullByte` check be also included for delegate username?
In Core we are doing so. It's not really required because it doesn't allow `\` character but we included for consistency.
Other than that, looks good :)

packages/lisk-transactions/test/utils/validation/validator.ts
```diff
@@ -592,4 +592,47 @@ describe('validator', () => {
 			).to.be.false;
 		});
 	});
+
+	describe('noNullBytes', () => {
+		let validate: ValidateFunction;
+		beforeEach(() => {
```
mitsuaki-u, line 598: prefer `async` and no promise for consistency. same for below

mitsuaki-u (approved):

lsilvs (approved):

webmaster128:
The term _noNullByte_ is highly misleading as you are checking JavaScript strings, which is a list of unicode characters. There is no _byte_ in a string. A string can be converted to a list of bytes by encoding it but a string itself is independent of any binary representation.

So this should be called _noNullCharacter_.

packages/lisk-transactions/test/utils/validation/validation.ts
```diff
@@ -369,4 +370,47 @@ describe('validation', () => {
 			return expect(isValidInteger(-6)).to.be.true;
 		});
 	});
+
+	describe('#isNullByteIncluded', () => {
+		const validStrings = [
+			'lorem ipsum',
+			'lorem\u0001 ipsum',
+			'loremU00000001 ipsum',
```
webmaster128, line 378: 'U00000000' should also be valid but it is not (see https://github.com/LiskHQ/lisk/commit/5c7e2944383bf369f1343fba62c71e84de6f9664)

PR #1117, opened 2019-02-18
Make sure that all P2PNodeInfo/peerHeaders has an options object to support custom properties - Closes #1115

### LINKED ISSUE #1115: Make sure that all P2PNodeInfo/peerHeaders has an options object to support custom properties
### Expected behavior

- When a peer asks our node for our peer list, we need to make sure that the peer info objects have an an `options` property if available for those peers.
- When validating incoming peer info (from legacy protocol), the options object cannot contain the whole legacy peer info; we need to construct the options object by taking the current legacy top-level fields (`nonce`, `broadhash` and `httpPort`) and combining them with any additional custom fields from the options object (if the options object exists). This is necessary to allow the P2P library to stay generic.

### Actual behavior

- Inside validation.ts, the options field is not being constructed correctly and this could lead to issues.
- When the PeerPool provides its list of peers, it does not add the options object to the payload.

jondubois:
### What was the problem?

- The options object was not constructed properly in some parts of the code when receiving peer info.
- One of the validation functions turned out to be unnecessary.
- When converting the node info into the legacy protocol format, we were not adding the options property; this is necessary to allow the node to pass custom fields to peers.

### How did I fix it?

- Added missing options fields where required and improved how the options object is constructed.
- Reused validation functionality where possible when handling inbound peer info; we did not need a separate function.
- Exposed new events from the peer which were not being used before.

### How to test it?

```
npm run test
```

### Review checklist

* The PR resolves #1115
* Linting passes
* Tests pass

### DIFF
--- a/packages/lisk-p2p/src/peer.ts
+++ b/packages/lisk-p2p/src/peer.ts
@@ -25,7 +25,6 @@ import {
 	P2PRequestPacket,
 	P2PResponsePacket,
 	ProtocolNodeInfo,
-	ProtocolPeerInfo,
 } from './p2p_types';
 
 import { P2PRequest } from './p2p_request';
@@ -37,7 +36,6 @@ import {
 	validatePeerInfoList,
 	validateProtocolMessage,
 	validateRPCRequest,
-	validateStatusProtocolPeerInfo,
 } from './validation';
 
 // This interface is needed because pingTimeoutDisabled is missing from ClientOptions in socketcluster-client.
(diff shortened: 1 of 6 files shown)


packages/lisk-p2p/src/peer.ts
```diff
@@ -112,21 +110,6 @@ const convertNodeInfoToLegacyFormat = (
 	nonce: nodeInfo.options ? (nodeInfo.options.nonce as string) : '',
 });
 
-const convertLegacyPeerInfoToNewFormat = (
```
ishantiw, line 115: maybe we can add this function in `validation.ts`?

ishantiw (commented):
Looks good, just a minor comment

packages/lisk-p2p/src/peer.ts
jondubois: Sounds good. Maybe this could it be done as a separate issue? I didn't want to change the existing validation function and risk breaking it (or the unit tests).

shuse2 (approved):

packages/lisk-p2p/src/peer.ts
ishantiw: Yes, it's a good idea to do it in a separate PR.

ishantiw (approved):

packages/lisk-p2p/src/peer.ts
shuse2: @jondubois Please make another issue regards to this

PR #1119, opened 2019-02-18
Allow custom peer selection function to be passed and used in P2P - Closes#1113

### LINKED ISSUE #1113: Pass peer selection algorithm as an argument to p2p constructor
### Expected behavior

We want a user to use its own peer selection algorithm based on his preferences and parameters. To achieve that we will allow the peer selection algorithm to be passed as an argument to the p2p constructor.

### Actual behavior

Right now it's hardcoded to use only default peer selection algorithm that can't be modified or extended by a user.

ishantiw:
### Description

With the introduction of custom Peer Selection function, a user could pass and use P2P library with its own Peer Selection algorithm based on its preferred custom parameters. Below are the changes,

* Allow P2P to accept custom peer selection algorithm that will be used for send/request and also peer selection for connection in PeerPool, hence initialize PeerPool constructor by passing this custom Peer Selection function. 

* Based on P2P class change, introduce new type definition for peer selection function and arguments that will be available to user as well.

* Modify PeerPool constructor to accept peer selection functions as argument, if undefined then use default ones

* Make default peer selection algorithm to use type definition from `p2p_types.ts` to have consistency.

* Add an argument to `selectForConnection` function to accept `selectionParams` so that in future it could be modified as well to accept custom params for selection.

* Update peer selection tests

### Review checklist

* The PR resolves #1113  
* Linting passes
* Tests pass

### DIFF
--- a/packages/lisk-p2p/src/p2p.ts
+++ b/packages/lisk-p2p/src/p2p.ts
@@ -43,7 +43,6 @@ import {
 	P2PNetworkStatus,
 	P2PNodeInfo,
 	P2PPeerInfo,
-	P2PPeerSelectionFunctions,
 	P2PPenalty,
 	P2PRequestPacket,
 	P2PResponsePacket,
@@ -107,7 +106,9 @@ export class P2P extends EventEmitter {
 
 	private readonly _handlePeerPoolRPC: (request: P2PRequest) => void;
 	private readonly _handlePeerPoolMessage: (message: P2PMessagePacket) => void;
-	private readonly _handleDiscoveredPeer: (discoveredPeerInfo: P2PDiscoveredPeerInfo) => void;
+	private readonly _handleDiscoveredPeer: (
+		discoveredPeerInfo: P2PDiscoveredPeerInfo,
+	) => void;
 	private readonly _handleFailedToPushNodeInfo: (error: Error) => void;
 	private readonly _handleFailedToFetchPeerInfo: (error: Error) => void;
 	private readonly _handlePeerConnect: (peerInfo: P2PPeerInfo) => void;
@@ -118,7 +119,7 @@ export class P2P extends EventEmitter {
 	private readonly _handleOutboundSocketError: (error: Error) => void;
 	private readonly _handleInboundSocketError: (error: Error) => void;
 
-	public constructor(config: P2PConfig, peerSelectionFunction?: P2PPeerSelectionFunctions) {
+	public constructor(config: P2PConfig) {
 		super();
 		this._config = config;
 		this._isActive = false;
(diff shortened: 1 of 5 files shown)


packages/lisk-p2p/src/p2p.ts
```diff
@@ -111,7 +112,7 @@ export class P2P extends EventEmitter {
 	private readonly _handleOutboundSocketError: (error: Error) => void;
 	private readonly _handleInboundSocketError: (error: Error) => void;
 
-	public constructor(config: P2PConfig) {
+	public constructor(config: P2PConfig, peerSelectionFunction?: P2PPeerSelectionFunctions) {
```
shuse2: name should be `peerSelectionFunctions`?
Also, should we put it in the config then we don't need `peerSelectionFunctions` and we can put one by one

jondubois: I would also prefer them to be in the `P2PConfig` individually as top-level optional fields.

packages/lisk-p2p/src/p2p_types.ts
```diff
@@ -105,6 +105,25 @@ export interface ProtocolNodeInfo {
 	readonly options?: P2PInfoOptions;
 }
 
+export interface P2PPeerSelectionOptions {
+	readonly [key: string]: string | number;
```
jondubois: Should the value be restricted to only string or number? Maybe it should be `unknown`? 

Or if unknown doesn't work, maybe we can make an exception to TSLint here and allow `any`? Then it would be the user's responsibility to enforce type constraints.

Ideally, the user should be able to override our types with more specific (but compatible) types in their own function declaration.

packages/lisk-p2p/src/peer_pool.ts
```diff
@@ -90,12 +93,15 @@ export class PeerPool extends EventEmitter {
 	private readonly _handlePeerOutboundSocketError: (error: Error) => void;
 	private readonly _handlePeerInboundSocketError: (error: Error) => void;
 	private _nodeInfo: P2PNodeInfo | undefined;
+	private readonly _peerSelectForSendRequest: P2PPeerSelectionForSendRequest;
+	private readonly _peerSelectForConnection: P2PPeerSelectionForConnection;
 
-	public constructor(peerPoolConfig: PeerPoolConfig) {
+	public constructor(peerPoolConfig: PeerPoolConfig, peerSelectionFunction?: P2PPeerSelectionFunctions) {
```
jondubois: If the functions are moved to P2PConfig, then it should be done here too.

jondubois (changes requested):

packages/lisk-p2p/src/p2p_types.ts
ishantiw: Instead of using `unknown` or `any`, we can also add `object` along with `string` and `number`?

jondubois: As discussed, consider making the selection function accept `P2PNodeInfo` as a parameter instead of `P2PPeerSelectionOptions/PeerOptions`. Then the selection function should use `nodeInfo.height` instead of `selectionParams.lastBlockHeight`. That would allow the selection function to operate on any custom field which might be in the nodeInfo's `options` object.

jondubois (dismissed):

packages/lisk-p2p/src/p2p.ts
```diff
@@ -200,13 +202,18 @@ export class P2P extends EventEmitter {
 		this._peerPool = new PeerPool({
 			connectTimeout: this._config.connectTimeout,
 			ackTimeout: this._config.ackTimeout,
+			peerSelectionForSendRequest: config.peerSelectionForSendRequest,
```
shuse2: should we assign default here? i feel like it's more clear than assign the default internally

ishantiw: Yes, it's better to assign it in P2P constructor. I have made the relevant changes.

shuse2 (approved):

jondubois (approved):

PR #1124, opened 2019-02-20
Add scenarios in integration tests relevant to lisk-p2p

ishantiw:
### Description

Add missing integration tests on lisk-p2p. Cover different scenarios like, for example, when a couple of nodes become unresponsive due to some reason, other nodes connected should be able to detect and update their network status gracefully. 

### Review checklist

* The PR resolves #INSERT_ISSUE_NUMBER
* Linting passes
* Tests pass

### DIFF
--- a/packages/lisk-p2p/test/integration/p2p.ts
+++ b/packages/lisk-p2p/test/integration/p2p.ts
@@ -282,8 +282,8 @@ describe('Integration tests for P2P library', () => {
 			});
 
 			describe('P2P.send when peers are at different heights', () => {
-				let collectedMessages: Array<any> = [];
 				const randomPeerIndex = Math.floor(Math.random() * NETWORK_PEER_COUNT);
+				let collectedMessages: Array<any> = [];
 				let randomP2PNode: any;
 
 				beforeEach(async () => {
@@ -301,8 +301,8 @@ describe('Integration tests for P2P library', () => {
 
 				it('should send messages to subset of peers within the network with updated heights; should reach multiple peers with even distribution', async () => {
 					const TOTAL_SENDS = 100;
-
 					const nodePortToMessagesMap: any = {};
+
 					p2pNodeList.forEach(p2p => {
 						p2p.applyNodeInfo({
 							os: platform(),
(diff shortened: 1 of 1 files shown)


packages/lisk-p2p/test/integration/p2p.ts
```diff
@@ -46,9 +47,9 @@ describe('Integration tests for P2P library', () => {
 		afterEach(async () => {
 			await Promise.all(
 				p2pNodeList.map(async p2p => {
-					try {
+					if (p2p.isActive) {
```
shuse2: maybe filter first?
`p2pNodeList.filter(p2p => p2p.isActive).map(xxx`

packages/lisk-p2p/test/integration/p2p.ts
```diff
@@ -198,15 +211,177 @@ describe('Integration tests for P2P library', () => {
 				await wait(DISCOVERY_INTERVAL * 5);
 
 				p2pNodeList.forEach(p2p => {
-					let {connectedPeers} = p2p.getNetworkStatus();
+					let { connectedPeers } = p2p.getNetworkStatus();
```
shuse2: `const` here?

packages/lisk-p2p/test/integration/p2p.ts
```diff
@@ -58,6 +59,18 @@ describe('Integration tests for P2P library', () => {
 				expect(p2p).to.have.property('isActive', true);
 			});
 		});
+
+		describe('P2P.request', () => {
+			it('should throw an error when not able to get any peer in peer selection', async () => {
```
shuse2, line 62: does user have control over if peer can be obtained in selection or not?
or was if it at least has seed, this one happen?

shuse2 (changes requested):

packages/lisk-p2p/test/integration/p2p.ts
ishantiw: No, user doesn't have control over the default peer selection but I think that needs to be covered as well where the user is passing its own peer selection function. But here, we are just testing in a completely disconnected network with no peers and seed nodes, we expect it to throw this error.

ishantiw:
I will add test cases related to the custom selection function. Issue created #1131

shuse2 (approved):

jondubois (approved):

PR #1132, opened 2019-02-22
Update P2PPeerInfo to contain discovered peer info and options field Closes#1130

### LINKED ISSUE #1130: Allow selection function to use P2PDiscoveredPeerInfo
### Expected behavior

- Default or custom selection should be able to use more details of a peer to select a peer like version, broadhash, os, etc.

- Also, it affects `triedPeer` and `newPeer` list as it only holds `P2PPeerInfo` right now. So it loses the information it discovered about other peers during discovery or `getStatus` calls

### Actual behavior

Right now, selection only gets an argument that has `ip`, `wsPort` and `height` as peer info.

We have to change, `P2PPeerInfo` to include detailed peer info to make it more robust and easy to pass around `P2P` and `PeerPool`

ishantiw:
### Description

Update `PeerInfo` to contain `DiscoveredPeerInfo` so that it could be passed to selection function and also newPeer/triedPeer can hold it. Make relevant changes in all the relevant classes and module.

- `P2PPeerInfo` contains `discoveredInfo: DiscoveredPeerInfo` and `options: P2PInfoOptions`.
- Change all the classes and modules to use `P2PPeerInfo`.
- Update Validation functions to use updated `P2PPeerInfo`.
- Fix all the tests related to this change.

### Review checklist

* The PR resolves #1130
* Linting passes
* Tests pass


packages/lisk-p2p/src/p2p.ts
```diff
@@ -340,17 +337,20 @@ export class P2P extends EventEmitter {
...
 					wsPort,
 					height: queryObject.height ? +queryObject.height : 0,
 					isDiscoveredPeer: true,
-					version: queryObject.version,
-					options:
-						typeof queryObject.options === 'string'
-							? JSON.parse(queryObject.options)
-							: undefined,
+					discoveredInfo: {
+						version: queryObject.version,
```
shuse2: prefer to put it in the root level, and make it mandatory.
maybe separate the type for `seed` and others?
tbd

packages/lisk-p2p/src/p2p.ts
```diff
@@ -463,7 +467,10 @@ export class P2P extends EventEmitter {
 			throw new Error('Cannot start the node because it is already active');
 		}
 		await this._startPeerServer();
-		await this._startDiscovery(this._config.seedPeers);
+		// TODO: Once we will a new peer discovery then we can remove this typecasting to P2PDiscoveredPeerInfo
+		await this._startDiscovery(this._config.seedPeers as ReadonlyArray<
+			P2PDiscoveredPeerInfo
+		>);
```
jondubois, line 473: It seems that `P2PDiscoveredPeerInfo` requires the `version` field, does this transpile (I mean the `as` typecasting)?

packages/lisk-p2p/src/peer.ts
```diff
@@ -97,17 +97,18 @@ export interface PeerConnectionState {
 export const constructPeerId = (ipAddress: string, wsPort: number): string =>
 	`${ipAddress}:${wsPort}`;
 
-export const constructPeerIdFromPeerInfo = (peerInfo: P2PPeerInfo): string =>
-	`${peerInfo.ipAddress}:${peerInfo.wsPort}`;
+export const constructPeerIdFromPeerInfo = (
+	peerInfo: P2PPeerInfo | P2PDiscoveredPeerInfo,
+): string => `${peerInfo.ipAddress}:${peerInfo.wsPort}`;
```
jondubois: `P2PDiscoveredPeerInfo` inherits from `P2PPeerInfo`, so you shouldn't need the `|`. The argument type could just be `P2PPeerInfo`.

Don't worry about this if it's intentional (if you prefer to be more explicit).

packages/lisk-p2p/src/peer_pool.ts
```diff
@@ -422,30 +422,29 @@ export class PeerPool extends EventEmitter {
...
-						if (!peerDetailedInfo) {
+						if (!peerDiscoveredInfo) {
 							return undefined;
 						}
 
 						// The options property is not read by the current legacy protocol but it should be added anyway for future compatibility.
 						return {
-							broadhash: peerDetailedInfo.options
-								? (peerDetailedInfo.options.broadhash as string)
+							broadhash: peerDiscoveredInfo.options
```
jondubois: `peerDiscoveredInfo.options` looks like it's not supposed to be here.

packages/lisk-p2p/src/validation.ts
```diff
@@ -76,21 +76,20 @@ export const validatePeerInfo = (
...
-		options: {
-			httpPort: protocolPeer.httpPort,
-			broadhash: protocolPeer.broadhash,
-			nonce: protocolPeer.nonce,
-			...protocolPeer.options
-		},
-		isDiscoveredPeer: true,
+		httpPort: protocolPeer.httpPort,
+		broadhash: protocolPeer.broadhash,
+		nonce: protocolPeer.nonce,
```
jondubois: Do these fields still need to be listed explicitly? They should already be added as part of the expansion `...protocolPeer`.

packages/lisk-p2p/src/peer_pool.ts
```diff
@@ -422,30 +422,29 @@ export class PeerPool extends EventEmitter {
 			peers: this._pickRandomDiscoveredPeers(MAX_PEER_LIST_BATCH_SIZE)
 				.map(
 					(peer: Peer): ProtocolPeerInfo | undefined => {
-						const peerDetailedInfo: P2PDiscoveredPeerInfo | undefined =
-							peer.detailedPeerInfo;
+						const peerDiscoveredInfo: P2PDiscoveredPeerInfo | undefined =
```
shuse2: do we need to define type here?

packages/lisk-p2p/test/integration/p2p.ts
```diff
@@ -151,11 +150,10 @@ describe('Integration tests for P2P library', () => {
 		beforeEach(async () => {
 			p2pNodeList = [...Array(NETWORK_PEER_COUNT).keys()].map(index => {
 				// Each node will have the next node in the sequence as a seed peer.
-				const seedPeers = [
+				const seedPeers: ReadonlyArray<P2PPeerInfo> = [
```
shuse2: do we need to define here? same as below

packages/lisk-p2p/src/validation.ts
```diff
@@ -76,21 +76,20 @@ export const validatePeerInfo = (
...
 	};
 
-	return peerInfo;
+	const { ip, ...peerInfoUpdated } = peerInfo;
```
jondubois, line 87: Nice 👍

packages/lisk-p2p/src/p2p_types.ts
```diff
@@ -64,11 +57,12 @@ export interface P2PNodeInfo {
 	readonly nethash: string;
 	readonly wsPort: number;
 	readonly height: number;
-	readonly options?: P2PInfoOptions;
+	// tslint:disable-next-line: no-mixed-interface
+	readonly [key: string]: unknown;
```
shuse2, line 61: if we are communicating through header/url is it possible to be `object`?

packages/lisk-p2p/src/p2p.ts
jondubois: After adding the new peer discovery mechanism (e.g. in the next PR), maybe this `_startDiscovery` method could be renamed to `_startShuffling` or `_startPeerShuffling` and then we could create a new/separate discovery function/interval from scratch. Is that the intention?

My thought is that the new discovery interval/algorithm should be different enough that it should be implemented from scratch as an independent interval.

In the current version of the protocol, the shuffling should only add/connect to peers (it should never intentionally remove/disconnect from peers) but once we migrate to the LIP protocol, we can adjust the shuffling function to make it remove some peers too (so each interval, it will add some peers and also remove some peers to maintain an optimum amount of peers).

jondubois (changes requested):

packages/lisk-p2p/test/integration/p2p.ts
ishantiw: I just added to use autocompletion to fill up the field names, but yeah we don't need it, I'll remove it.

packages/lisk-p2p/src/p2p_types.ts
ishantiw: Yes, it is possible that it could be an object, I mean we could still pass `options?` field, but now it can accept a wide range of types of values

packages/lisk-p2p/src/peer.ts
ishantiw: Yeah I was just trying to be specific, you are right we can only use `P2PPeerInfo`

packages/lisk-p2p/src/peer_pool.ts
ishantiw: Yeah I missed this one

packages/lisk-p2p/src/p2p.ts
ishantiw: > It seems that `P2PDiscoveredPeerInfo` requires the `version` field, does this transpile (I mean the `as` typecasting)?

It does transpile with `as`, it is also missing required height but still it works but this is only temporary as I mentioned in the comments that this will be removed with the new discovery function and then we don't have to typecast. 



> After adding the new peer discovery mechanism (e.g. in the next PR), maybe this `_startDiscovery` method could be renamed to `_startShuffling` or `_startPeerShuffling` and then we could create a new/separate discovery function/interval from scratch. Is that the intention?
> 
> My thought is that the new discovery interval/algorithm should be different enough that it should be implemented from scratch as an independent interval.
> 
> In the current version of the protocol, the shuffling should only add/connect to peers (it should never intentionally remove/disconnect from peers) but once we migrate to the LIP protocol, we can adjust the shuffling function to make it remove some peers too (so each interval, it will add some peers and also remove some peers to maintain an optimum amount of peers).

Yes actually we can change it to `peerStartShuffling` but maybe when we start with phase 2 . Like you said, in current protocol it doesn;t remove any connection so it doesn;t make sense now to have it. 
What I wanted was, when we should trigger discovery again, like we do it in the start up of a node and then we should do it either at certain interval or at some trigger like if number of active connections are lower than some threshold.
I guess having both periodic and trigger is what is required, do you think we should still keep _startDiscovery periodically and using the new discovery?

ishantiw:
@jondubois @shuse2 Addressed

shuse2 (approved):

packages/lisk-p2p/src/p2p.ts
jondubois: My thinking is that we shouldn't treat seed peers as special; instead, I propose that we should combine them with the _newPeers and _triedPeers maps as input to the discovery function.

If we implement a new discovery function to discover from seed peers (who only have ip and port) then we should be able to run that exact same function on an interval against any kind of `P2PPeerInfo` including `P2PDiscoveredPeerInfo` (e.g. from the `_newPeers` and `_triedPeers` maps) and it should work exactly the same. Since the selection function is not involved in discovery, we don't need to worry about having the full `P2PDiscoveredPeerInfo`, we can just make the input type to the discovery function be `P2PPeerInfo` because all it will ever need is `wsPort` and `ipAddress`.

If that seed discovery function can be used to cover all cases and works with any kind of peer info (just by invoking it on an interval), I don't see why we need a second discovery function. We do need a function to connect to discovered peers though, but it doesn't classify as discovery anymore.

jondubois: Also, from the protocol perspective, it would be ugly to have two different kinds of discovery protocols; a short one where the connection disconnects after a few milliseconds and a long one which stays open. I propose we just use the short approach (probe-only discovery); that makes the protocol more consistent, predictable and easier to debug.

ishantiw: > My thinking is that we shouldn't treat seed peers as special; instead, I propose that we should combine them with the _newPeers and _triedPeers maps as input to the discovery function.
> 
> If we implement a new discovery function to discover from seed peers (who only have ip and port) then we should be able to run that exact same function on an interval against any kind of `P2PPeerInfo` including `P2PDiscoveredPeerInfo` (e.g. from the `_newPeers` and `_triedPeers` maps) and it should work exactly the same. Since the selection function is not involved in discovery, we don't need to worry about having the full `P2PDiscoveredPeerInfo`, we can just make the input type to the discovery function be `P2PPeerInfo` because all it will ever need is `wsPort` and `ipAddress`.
> 
> If that seed discovery function can be used to cover all cases and works with any kind of peer info (just by invoking it on an interval), I don't see why we need a second discovery function. We do need a function to connect to discovered peers though, but it doesn't classify as discovery anymore.

Yes I agree, having two different discovery function instead of that I will update the discovery to be more generic and accept a list of peers with Ip and Port that will be used to fetch peers and close the connection and will be independent of selection function and peer management in Peer Pool

ishantiw: @jondubois I will cover it another PR (issue #1152) which is only to work on the discovery process in general. Could we merge this one if there are no more comments from your end.

jondubois (approved):

PR #1134, opened 2019-02-25
Fix init script and template - Closes #1133

### LINKED ISSUE #1133: Remove babel and eslint setting and suffix and update init command
### Description
After moving to TypeScript, eslint and babel files are not needed anymore.
Therefore, it should be removed.

Also, `npm run init` shell is outdated because of that, so it should be updated too

shuse2:
### What was the problem?
- template filename was outdated
- init file did not have package.json
- test config needed to be created

### How did I fix it?
- change the path of template and fix symlink

### How to test it?
`npm t`
`npm run init new-package`

### Review checklist

* The PR resolves #1133 
* Linting passes
* Tests pass


templates/package.json.tmpl
```diff
@@ -0,0 +1,89 @@
...
+		"clean": "./scripts/clean.sh",
+		"format": "prettier --write \"*.{ts,js,json}\" \"{src,test}/**/*.{ts,js,json}\"",
+		"lint": "tslint --format verbose --project .",
+		"lint:fix": "npm run lint -- --fix",
+		"test": "TS_NODE_PROJECT=./test/tsconfig.json nyc mocha test/{,/**/,/**/**/}/*.ts",
+		"test:watch": "npm test -- --watch",
+		"test:watch:min": "npm run test:watch -- --reporter=min",
+		"test:node": "npm run build:check",
+		"serve:start": "http-server -p 11545 ./browsertest &",
+		"serve:stop": "kill $(lsof -t -i:11545) || true",
```
yatki: Here we had to provide a different port numbers for each package. Having them hard coded in the template may cause a problem. 

Maybe we should get the port number the user or setting an invalid port number like `-p PORT_HERE` so after initialisation force user to fix it?

yatki (changes requested):
just one concern with the package.json

yatki (approved):

PR #1136, opened 2019-02-25
Fix multisignature double check - Closes #1122

### LINKED ISSUE #1122: Multisignature registration transaction always fails with error: 'Account is not a multisignature'
### Expected behavior
For the multisignature registration transaction, the transaction asset contains signatures property and they should be verified against the keysgroup present in the transaction asset. 
### Actual behavior
The transaction fails because BaseTransaction tries to verify signatures against the sender. Since sender is not a multisignature account, it fails with 'Account is not a multisignature account'.
2.1.0-alpha.1

shuse2:
### What was the problem?
`MultisignatureTransaction` was checking the sender account for the keysgroup.

### How did I fix it?
- override processMultisignatures in `MultisignatureTransaction`
- updated test to check processMultisignatures in `MultisignatureTransaction`

### How to test it?
`npm t`

### Review checklist

* The PR resolves #1122
* Linting passes
* Tests pass

### DIFF
--- a/packages/lisk-transactions/test/4_multisignature_transaction.ts
+++ b/packages/lisk-transactions/test/4_multisignature_transaction.ts
@@ -48,6 +48,9 @@ describe('Multisignature transaction class', () => {
 		nonMultisignatureSender = nonMultisignatureAccount;
 		multisignatureSender = validMultisignatureAccount;
 		storeAccountGetStub = sandbox
+			.stub(store.account, 'getOrDefault')
+			.returns(nonMultisignatureSender);
+		storeAccountGetStub = sandbox
 			.stub(store.account, 'get')
 			.returns(nonMultisignatureSender);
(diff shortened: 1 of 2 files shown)


pablitovicente (approved):

packages/lisk-transactions/test/4_multisignature_transaction.ts
```diff
@@ -321,18 +383,6 @@ describe('Multisignature transaction class', () => {
 			expect(errors).not.to.be.empty;
 			expect(errors[0].dataPath).to.be.equal('.signatures');
 		});
-
-		it('should return error when transaction signatures missing', async () => {
```
SargeKhan, line 325: Why did we remove this test?

SargeKhan: Never mind. I found the reason.

SargeKhan (approved):

PR #1151, opened 2019-02-27
Second signature registration transaction should update secondSignature flag for account - Closes #1147

### LINKED ISSUE #1147: Second signature registration transaction should update secondSignature flag for account
### Expected behavior
In the applyAsset function for Second signature transaction, the secondPublicKey and secondSignature property should be updated.
### Actual behavior
The applyAsset function only updates the secondPublicKey property.
2.1.0

pablitovicente:
### What was the problem?

- secondSignature flag was not being set

### How did I fix it?

- secondSignature flag is being added on apply asset and removed in undo

### How to test it?

Build should be green

### Review checklist

* The PR resolves #1147 
* Linting passes
* Tests pass

### DIFF
--- a/packages/lisk-transactions/src/1_second_signature_transaction.ts
+++ b/packages/lisk-transactions/src/1_second_signature_transaction.ts
@@ -200,7 +200,7 @@ export class SecondSignatureTransaction extends BaseTransaction {
 		const updatedSender = {
 			...sender,
 			secondPublicKey: this.asset.signature.publicKey,
-			secondSignature: true,
+			secondSignature: 1,
 		};
 		store.account.set(updatedSender.address, updatedSender);
 
@@ -212,7 +212,7 @@ export class SecondSignatureTransaction extends BaseTransaction {
 		const strippedSender = {
 			...sender,
 			secondPublicKey: undefined,
-			secondSignature: false,
+			secondSignature: 0,
 		};
 
 		store.account.set(strippedSender.address, strippedSender);
--- a/packages/lisk-transactions/src/1_second_signature_transaction.ts
+++ b/packages/lisk-transactions/src/1_second_signature_transaction.ts
@@ -209,8 +209,11 @@ export class SecondSignatureTransaction extends BaseTransaction {
 
 	protected undoAsset(store: StateStore): ReadonlyArray<TransactionError> {
 		const sender = store.account.get(this.senderId);
-		const { secondPublicKey, ...strippedSender } = sender;
-		strippedSender.secondSignature = false;
+		const strippedSender = {
+			...sender,
+			secondPublicKey: undefined,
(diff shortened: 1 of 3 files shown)


packages/lisk-transactions/src/1_second_signature_transaction.ts
```diff
@@ -209,6 +210,8 @@ export class SecondSignatureTransaction extends BaseTransaction {
 	protected undoAsset(store: StateStore): ReadonlyArray<TransactionError> {
 		const sender = store.account.get(this.senderId);
 		const { secondPublicKey, ...strippedSender } = sender;
```
shuse2:
```suggestion
		const { secondPublicKey, secondSignature, ...strippedSender } = sender;
```

pablitovicente: @shuse2 I think we need the properties to be present for stateStore to handle them correctly so instead changed to de-structuring as we discussed.

packages/lisk-transactions/src/1_second_signature_transaction.ts
```diff
@@ -200,6 +200,7 @@ export class SecondSignatureTransaction extends BaseTransaction {
 		const updatedSender = {
 			...sender,
 			secondPublicKey: this.asset.signature.publicKey,
+			secondSignature: true,
```
SargeKhan: It should be a number, with values either 0 or 1.

pablitovicente: Done!

shuse2 (dismissed):

packages/lisk-transactions/src/1_second_signature_transaction.ts
```diff
@@ -208,7 +209,12 @@ export class SecondSignatureTransaction extends BaseTransaction {
 
 	protected undoAsset(store: StateStore): ReadonlyArray<TransactionError> {
 		const sender = store.account.get(this.senderId);
-		const { secondPublicKey, ...strippedSender } = sender;
+		const strippedSender = {
+			...sender,
+			secondPublicKey: undefined,
```
SargeKhan: In Lisk-core, on undo action, the secondPublicKey property is assigned with the value `null`. Therefore, I think we should keep it consistent.

pablitovicente: @SargeKhan  We have a TS lint rule that forbids null and favours undefined should I skip it in this case?

SargeKhan: Yeah, I guess we have to do it in this particular case.

SargeKhan (approved):

shuse2 (approved):

PR #1159, opened 2019-03-04
Fix receivedAt to use iso date format as input and store as date - Closes #1156

### LINKED ISSUE #1156: Transaction receiveAt doesn't match the input/output
### Expected behavior
transaction.toJSON() should return instanciatable object.

### Actual behavior
toJSON() returns string, but constructor expects Data object

2.0-rc.2-

shuse2:
### What was the problem?
property receivedAt did not have matching property input/output

### How did I fix it?
`receivedAt` takes in valid string format of date, and outputs ISODateFormat.
internally it stores Date object

### How to test it?
`npm t`

### Review checklist

* The PR resolves #1156
* Linting passes
* Tests pass

### DIFF
--- a/packages/lisk-transactions/src/base_transaction.ts
+++ b/packages/lisk-transactions/src/base_transaction.ts
@@ -57,12 +57,6 @@ export interface TransactionResponse {
 	readonly errors: ReadonlyArray<TransactionError>;
 }
 
-export interface Attributes {
-	readonly [entity: string]: {
-		readonly [property: string]: ReadonlyArray<string>;
-	};
-}
-
 export interface StateStoreGetter<T> {
 	get(key: string): T;
 	find(func: (item: T) => boolean): T | undefined;
@@ -162,7 +156,9 @@ export abstract class BaseTransaction {
 		this._signSignature = rawTransaction.signSignature;
 		this.timestamp = rawTransaction.timestamp;
 		this.type = rawTransaction.type;
-		this.receivedAt = rawTransaction.receivedAt || new Date();
+		this.receivedAt = rawTransaction.receivedAt
+			? new Date(rawTransaction.receivedAt)
+			: new Date();
 	}
 
 	public get id(): string {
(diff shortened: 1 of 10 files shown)


pablitovicente (dismissed):

packages/lisk-transactions/src/base_transaction.ts
```diff
@@ -162,7 +156,9 @@ export abstract class BaseTransaction {
 		this._signSignature = rawTransaction.signSignature;
 		this.timestamp = rawTransaction.timestamp;
 		this.type = rawTransaction.type;
-		this.receivedAt = rawTransaction.receivedAt || new Date();
```
SargeKhan, line 165: `receivedAt` is a property which is only maintained in memory and is allowed to be undefined. Transactions saved in the blockchain don't maintain this property and every node maintains it's own value for `receivedAt` property. Therefore, I think it should only be explicitly set, if the `receivedAt` property isn't provided to the constructor, it should be undefined/null.

SargeKhan (changes requested):

packages/lisk-transactions/src/base_transaction.ts
shuse2: If we allow `receivedAt` as undefined, how should we calculate `timeElapsed` in the `isExpired` function?
it needs current time and the created time

SargeKhan: I guess receivedAt is only used in the transaction pool. Therefore, it's the responsibility of the transaction pool to assign receivedAt property for each transaction as soon as it enters transaction pool. It's done here: https://github.com/LiskHQ/lisk-elements/blob/development/packages/lisk-transaction-pool/src/transaction_pool.ts#L439

shuse2: so if `isExpired` is called without the receivedAt, it should throw an error?

pablitovicente (approved):

SargeKhan (approved):

PR #1170, opened 2019-03-07
Add connectAndFetchPeers and new Discovery function Closes#1152

### LINKED ISSUE #1152: Add a function to connectAndFetch peers to be used in discovery process
### Expected behavior

Discovery should just take `IP` address and `wPort` and should be able to connect and fetch peer list independently of `Peer` object. This will remove the need to create `Peer` objects for the seed peers just to fetch peer list from them.  Also, it will be independent of the peer pool in terms of choosing peers with Peer objects already and could be handled independently.

### Actual behavior

Discovery function accepts basic peer info `P2PPeerInfo` (with only `ip` and `port`) and uses `connectAndFetch` function to connect and fetch a peer list instead of creating Peer object for it. This will also simplify to filter out the black, white and fixed list in the future.

ishantiw:
### Description

Discovery should just take `IP` address and `wPort` and should be able to connect and fetch peer list independently of `Peer` object. This will remove the need to create `Peer` objects for the seed peers just to fetch peer list from them. Also, it will be independent of the peer pool in terms of choosing peers with Peer objects already and could be handled independently.

- Add new `connectAndFetchPeers` method that will fetch all the peers from `P2PPeerInfo ({ipAddress, wsPort})`, this will assign an outbound socket and destroy it after fetching peers.

- Update discovery function `discoverPeer` to use `connectAndFetchPeers` to fetch peers from the peers passed as arguments.

- Update P2P `start()` method and subsequent `_startDiscovery()` methods to use and accept new updated discovery method.

### Review checklist

* The PR resolves #1152
* Linting passes
* Tests pass

### DIFF
--- a/packages/lisk-p2p/src/p2p.ts
+++ b/packages/lisk-p2p/src/p2p.ts
@@ -347,8 +347,13 @@ export class P2P extends EventEmitter {
 
 				const wsPort: number = parseInt(queryObject.wsPort, BASE_10_RADIX);
 				const peerId = constructPeerId(socket.remoteAddress, wsPort);
+				const queryOptions =
+					typeof queryObject.options === 'string'
+						? JSON.parse(queryObject.options)
+						: undefined;
 
 				const incomingPeerInfo: P2PDiscoveredPeerInfo = {
+					...queryOptions,
 					...queryObject,
 					ipAddress: socket.remoteAddress,
 					wsPort,
@@ -372,6 +377,7 @@ export class P2P extends EventEmitter {
 				}
 			},
 		);
+
 		this._httpServer.listen(this._nodeInfo.wsPort, NODE_HOST_IP);
 		if (this._scServer.isReady) {
 			this._isActive = true;
--- a/packages/lisk-p2p/src/p2p.ts
+++ b/packages/lisk-p2p/src/p2p.ts
@@ -22,7 +22,11 @@ interface SCServerUpdated extends SCServer {
 	readonly isReady: boolean;
 }
 
-import { constructPeerId, constructPeerIdFromPeerInfo } from './peer';
+import {
+	connectAndFetchStatus,
+	constructPeerId,
+	constructPeerIdFromPeerInfo,
+} from './peer';
 
 import {
 	INCOMPATIBLE_NETWORK_CODE,
@@ -180,6 +184,9 @@ export class P2P extends EventEmitter {
 		this._handleDiscoveredPeer = (detailedPeerInfo: P2PDiscoveredPeerInfo) => {
 			const peerId = constructPeerIdFromPeerInfo(detailedPeerInfo);
 			if (!this._triedPeers.has(peerId)) {
+				if (this._newPeers.has(peerId)) {
+					this._newPeers.delete(peerId);
+				}
 				this._triedPeers.set(peerId, detailedPeerInfo);
 			}
 			// Re-emit the message to allow it to bubble up the class hierarchy.
@@ -344,17 +351,13 @@ export class P2P extends EventEmitter {
 
 				const wsPort: number = parseInt(queryObject.wsPort, BASE_10_RADIX);
 				const peerId = constructPeerId(socket.remoteAddress, wsPort);
-				const queryOptions =
-					typeof queryObject.options === 'string'
-						? JSON.parse(queryObject.options)
-						: undefined;
 
 				const incomingPeerInfo: P2PDiscoveredPeerInfo = {
+					...queryObject,
 					ipAddress: socket.remoteAddress,
 					wsPort,
 					height: queryObject.height ? +queryObject.height : 0,
 					version: queryObject.version,
-					...queryOptions,
 				};
 
 				const isNewPeer = this._peerPool.addInboundPeer(
@@ -366,9 +369,10 @@ export class P2P extends EventEmitter {
 				if (isNewPeer) {
 					this.emit(EVENT_NEW_INBOUND_PEER, incomingPeerInfo);
 					this.emit(EVENT_NEW_PEER, incomingPeerInfo);
-					if (!this._newPeers.has(peerId)) {
-						this._newPeers.set(peerId, incomingPeerInfo);
-					}
+				}
+
+				if (!this._newPeers.has(peerId) && !this._triedPeers.has(peerId)) {
+					this._newPeers.set(peerId, incomingPeerInfo);
 				}
 			},
 		);
@@ -412,18 +416,15 @@ export class P2P extends EventEmitter {
 	private async _discoverPeers(
 		knownPeers: ReadonlyArray<P2PDiscoveredPeerInfo> = [],
 	): Promise<void> {
-		const allKnownPeers: ReadonlyArray<
-			P2PDiscoveredPeerInfo
-		> = knownPeers.concat([...this._triedPeers.values()]);
-
 		// Make sure that we do not try to connect to peers if the P2P node is no longer active.
 		if (!this._isActive) {
 			return;
 		}
 
 		const discoveredPeers = await this._peerPool.runDiscovery(
-			allKnownPeers,
+			knownPeers,
 			this._config.blacklistedPeers,
+			this._nodeInfo,
 		);
 
 		// Stop discovery if node is no longer active. That way we don't try to connect to peers.
@@ -462,15 +463,40 @@ export class P2P extends EventEmitter {
 		clearInterval(this._discoveryIntervalId);
 	}
 
+	private async _fetchSeedPeerStatus(
+		seedPeers: ReadonlyArray<P2PPeerInfo>,
+	): Promise<ReadonlyArray<P2PDiscoveredPeerInfo>> {
+		const peerConfig = {
+			ackTimeout: this._config.ackTimeout,
+			connectTimeout: this._config.connectTimeout,
+		};
+		const seedPeerUpdatedInfos = await Promise.all(
+			seedPeers.map(async seedPeer =>
+				connectAndFetchStatus(seedPeer, this._nodeInfo, peerConfig),
+			),
+		);
+
+		return seedPeerUpdatedInfos;
+	}
+
 	public async start(): Promise<void> {
 		if (this._isActive) {
 			throw new Error('Cannot start the node because it is already active');
 		}
 		await this._startPeerServer();
+		// Fetch status of all the seed peers and then start the discovery
+		const seedPeerInfos = await this._fetchSeedPeerStatus(
+			this._config.seedPeers,
+		);
+		// Add seed's peerinfos in tried peer as we already tried them to fetch status
+		seedPeerInfos.forEach(seedInfo => {
+			const peerId = constructPeerIdFromPeerInfo(seedInfo);
+			if (!this._triedPeers.has(peerId)) {
+				this._triedPeers.set(peerId, seedInfo);
+			}
+		});
 		// TODO: Once we will a new peer discovery then we can remove this typecasting to P2PDiscoveredPeerInfo
-		await this._startDiscovery(this._config.seedPeers as ReadonlyArray<
-			P2PDiscoveredPeerInfo
-		>);
+		await this._startDiscovery(seedPeerInfos);
 	}
 
 	public async stop(): Promise<void> {
--- a/packages/lisk-p2p/src/p2p.ts
+++ b/packages/lisk-p2p/src/p2p.ts
@@ -424,7 +424,6 @@ export class P2P extends EventEmitter {
 		const discoveredPeers = await this._peerPool.runDiscovery(
 			knownPeers,
 			this._config.blacklistedPeers,
-			this._nodeInfo,
 		);
 
 		// Stop discovery if node is no longer active. That way we don't try to connect to peers.
--- a/packages/lisk-p2p/src/p2p.ts
+++ b/packages/lisk-p2p/src/p2p.ts
@@ -496,7 +496,7 @@ export class P2P extends EventEmitter {
 				this._triedPeers.set(peerId, seedInfo);
 			}
 		});
-		// TODO: Once we will a new peer discovery then we can remove this typecasting to P2PDiscoveredPeerInfo
+
 		await this._startDiscovery(seedPeerInfos);
 	}
 
--- a/packages/lisk-p2p/src/p2p.ts
+++ b/packages/lisk-p2p/src/p2p.ts
@@ -22,11 +22,7 @@ interface SCServerUpdated extends SCServer {
 	readonly isReady: boolean;
 }
 
-import {
-	connectAndFetchStatus,
-	constructPeerId,
-	constructPeerIdFromPeerInfo,
-} from './peer';
+import { constructPeerId, constructPeerIdFromPeerInfo } from './peer';
 
 import {
 	INCOMPATIBLE_NETWORK_CODE,
@@ -469,10 +465,10 @@ export class P2P extends EventEmitter {
 			ackTimeout: this._config.ackTimeout,
 			connectTimeout: this._config.connectTimeout,
 		};
-		const seedPeerUpdatedInfos = await Promise.all(
-			seedPeers.map(async seedPeer =>
-				connectAndFetchStatus(seedPeer, this._nodeInfo, peerConfig),
-			),
+		const seedPeerUpdatedInfos = await this._peerPool.fetchStatusAndCreatePeer(
+			seedPeers,
+			this._nodeInfo,
+			peerConfig,
 		);
 
 		return seedPeerUpdatedInfos;
--- a/packages/lisk-p2p/src/p2p.ts
+++ b/packages/lisk-p2p/src/p2p.ts
@@ -471,7 +471,7 @@ export class P2P extends EventEmitter {
 			ackTimeout: this._config.ackTimeout,
 			connectTimeout: this._config.connectTimeout,
 		};
-		const seedPeerUpdatedInfos = await this._peerPool.fetchStatusAndCreatePeer(
+		const seedPeerUpdatedInfos = await this._peerPool.fetchStatusAndCreatePeers(
 			seedPeers,
 			this._nodeInfo,
 			peerConfig,
--- a/packages/lisk-p2p/src/peer.ts
+++ b/packages/lisk-p2p/src/peer.ts
@@ -694,7 +694,7 @@ export const connectAndFetchPeerInfo = async (
 	basicPeerInfo: P2PPeerInfo,
 	nodeInfo?: P2PNodeInfo,
 	peerConfig?: PeerConfig,
-): Promise<PeerInfoAndOutboundConnection | Error> => {
+): Promise<PeerInfoAndOutboundConnection> => {
 	try {
 		const { responsePacket, socket } = await connectAndRequest(
 			basicPeerInfo,
@@ -713,6 +713,6 @@ export const connectAndFetchPeerInfo = async (
 
 		return { peerInfo, socket };
 	} catch (error) {
-		return error;
+		throw error;
 	}
 };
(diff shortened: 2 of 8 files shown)


ishantiw:
I have kept it in the draft state because there are a lot of changes that affect overall discovery,  connecting with incoming connections and maintaining Peers in PeerPool.

NOTE: Need to resolve a couple of tests related to stopping the node. I have skipped the partial network test (where nodes start at once without going into discovery) because it's not relevant anymore due to the change in initial discovery process.

ishantiw:
I'm reverting back the changes related to the use of independent `connectToFetchPeers` call in discovery as it was it creating a lot of instability in tests. Since its making independent calls in some interval, the connection doesn't close properly sometimes. Also, in the beginning, if there are no peers in the peer list of any node in the network then it cannot fetch peers of each other and hence throws the error. I'm still keeping the fetch `connectAndRequest` to fetch status of seed peers and adding them to peer pool in the initial discovery process.

packages/lisk-p2p/src/peer.ts
```diff
@@ -572,3 +572,149 @@ export class Peer extends EventEmitter {
...
+
+export const connectAndRequest = async (
+	basicPeerInfo: P2PPeerInfo,
+	procedure: string,
+	nodeInfo?: P2PNodeInfo,
+	peerConfig?: PeerConfig,
+): Promise<ConnectAndFetchResponse> =>
+	new Promise<ConnectAndFetchResponse>(
+		(
+			resolve: (result: ConnectAndFetchResponse) => void,
```
shuse2: probably no need to define type for `resolve` and `reject` here

packages/lisk-p2p/src/peer.ts
```diff
@@ -572,3 +572,149 @@ export class Peer extends EventEmitter {
...
+				multiplex: false,
+				autoConnect: false,
+				autoReconnect: false,
+				pingTimeoutDisabled: true,
+			};
+
+			const outboundSocket = socketClusterClient.create(clientOptions);
+			// Attaching handlers for various events that could be used future for logging or any other application
+			outboundSocket.on('error', async (error: Error) =>
+				Promise.resolve(error),
```
shuse2: it should use the `resolve` from input?
also i think this should be `reject`

packages/lisk-p2p/src/peer.ts
```diff
@@ -572,3 +572,149 @@ export class Peer extends EventEmitter {
...
+			);
+			outboundSocket.on('close', async () =>
+				Promise.resolve('Connection closed'),
```
shuse2: same as above, also it should rejected with `Error` object?

packages/lisk-p2p/src/peer.ts
```diff
@@ -572,3 +572,149 @@ export class Peer extends EventEmitter {
...
+			);
+			outboundSocket.on('connect', async () =>
+				Promise.resolve('Connection Successful'),
```
shuse2: should we already resolve on connection...?

packages/lisk-p2p/src/peer.ts
```diff
@@ -572,3 +572,149 @@ export class Peer extends EventEmitter {
...
+		}
+		const peers = validatePeerInfoList(responsePacket.data);
+
+		return peers;
+	} catch (error) {
+		throw error;
+	}
+};
+
+export const connectAndFetchStatus = async (
```
shuse2: should this and fetchPeer happen on the separate socket?

packages/lisk-p2p/src/peer.ts
```diff
@@ -572,3 +572,149 @@ export class Peer extends EventEmitter {
...
+				: undefined;
+			// Add a new field discovery to tell the receiving side that the connection will be short lived
+			const nodeInfoForDiscovery = { discovery: true, ...legacyNodeInfo };
+			const requestPacket = {
+				procedure,
+			};
+			const clientOptions: ClientOptionsUpdated = {
+				hostname: basicPeerInfo.ipAddress,
+				port: basicPeerInfo.wsPort,
+				query: querystring.stringify(nodeInfoForDiscovery),
```
jondubois: Does `querystring.stringify(...)` support nested objects (not just primitive types)? They should be allowed.

Maybe you can add a test case in the integration tests; maybe try adding a custom nested object to the `nodeInfo` object passed to the constructor of `P2P` instances for one of the test cases?

packages/lisk-p2p/test/integration/p2p.ts
```diff
@@ -72,7 +72,7 @@ describe('Integration tests for P2P library', () => {
 		});
 	});
 
-	describe('Partially connected network: All nodes launch at the same time. The seedPeers list of each node contains the next node in the sequence', () => {
+	describe.skip('Partially connected network: All nodes launch at the same time. The seedPeers list of each node contains the next node in the sequence', () => {
```
shuse2: if it's not relevant anymore why not remove it?

packages/lisk-p2p/src/peer.ts
jondubois: Can this be renamed to `connectAndFetchPeerInfo`?

packages/lisk-p2p/src/p2p.ts
```diff
@@ -462,15 +462,40 @@ export class P2P extends EventEmitter {
...
+	private async _fetchSeedPeerStatus(
+		seedPeers: ReadonlyArray<P2PPeerInfo>,
+	): Promise<ReadonlyArray<P2PDiscoveredPeerInfo>> {
+		const peerConfig = {
+			ackTimeout: this._config.ackTimeout,
+			connectTimeout: this._config.connectTimeout,
+		};
+		const seedPeerUpdatedInfos = await Promise.all(
+			seedPeers.map(async seedPeer =>
+				connectAndFetchStatus(seedPeer, this._nodeInfo, peerConfig),
```
jondubois: If `connectAndFetchStatus` fails for some of the peers, we should log it but it shouldn't stop us from continuing with the discovery process (so long as if succeeded for some of the peers).

Currently, the `Promise.all` call means that it's either all pass or all fail. If one of the seed peers is offline, our P2P node will fail to start completely.

packages/lisk-p2p/src/peer.ts
```diff
@@ -572,3 +572,149 @@ export class Peer extends EventEmitter {
...
+							basicPeerInfo.ipAddress,
+							basicPeerInfo.wsPort,
+						),
+					);
+				},
+			);
+		},
+	);
+
+export const connectAndFetchPeers = async (
```
shuse2: this function is not used...?

shuse2 (changes requested):

packages/lisk-p2p/src/peer.ts
```diff
@@ -572,3 +572,149 @@ export class Peer extends EventEmitter {
 		request.end(legacyNodeInfo);
 	}
 }
+
+export interface ConnectAndFetchResponse {
+	readonly responsePacket: P2PResponsePacket;
+	readonly socket?: SCClientSocket;
+}
+
+export const connectAndRequest = async (
```
jondubois, line 613: About `connectAndRequest`: I think it would be safer if the socket was destroyed at the end of this function after it has fetched whatever data was required; that way the entire lifecycle of the socket will be encapsulated within this function and it avoids the possibility that someone will forget to destroy the socket after calling this function.

packages/lisk-p2p/src/peer.ts
```diff
@@ -572,3 +572,149 @@ export class Peer extends EventEmitter {
...
+	try {
+		const { responsePacket, socket } = await connectAndRequest(
+			basicPeerInfo,
+			REMOTE_RPC_GET_ALL_PEERS_LIST,
+			nodeInfo,
+			peerConfig,
+		);
+
+		if (socket) {
+			socket.destroy();
```
jondubois: Maybe this can be moved inside the `connectAndRequest` function (see previous comment).

packages/lisk-p2p/src/peer.ts
```diff
@@ -572,3 +572,149 @@ export class Peer extends EventEmitter {
...
+): Promise<P2PDiscoveredPeerInfo> => {
+	try {
+		const { responsePacket, socket } = await connectAndRequest(
+			basicPeerInfo,
+			REMOTE_RPC_GET_NODE_INFO,
+			nodeInfo,
+			peerConfig,
+		);
+		if (socket) {
+			socket.destroy();
```
jondubois: Maybe this can be moved inside the `connectAndRequest` function (see previous comment).

packages/lisk-p2p/src/peer_discovery.ts
```diff
@@ -18,6 +18,7 @@ import { constructPeerIdFromPeerInfo, Peer } from './peer';
 export interface FilterPeerOptions {
 	readonly blacklist: ReadonlyArray<string>;
 }
+
 // TODO later: Implement LIPS to handle fixed and white list
 export const discoverPeers = async (
```
jondubois, line 23: We need to completely stop using `peer.fetchPeers()` for discovery or else we have two different discovery protocols. We should always use the new single-socket discovery approach.

It might seem more efficient to reuse the existing socket on the peer if it exists, but it's going to be a very difficult to manage these two different discovery approaches.

jondubois (changes requested):

packages/lisk-p2p/src/peer.ts
```diff
@@ -572,3 +572,149 @@ export class Peer extends EventEmitter {
...
+	new Promise<ConnectAndFetchResponse>(
+		(
+			resolve: (result: ConnectAndFetchResponse) => void,
+			reject: (result: Error) => void,
+		): void => {
+			const legacyNodeInfo = nodeInfo
+				? convertNodeInfoToLegacyFormat(nodeInfo)
+				: undefined;
+			// Add a new field discovery to tell the receiving side that the connection will be short lived
+			const nodeInfoForDiscovery = { discovery: true, ...legacyNodeInfo };
```
jondubois: This `discovery: true` is good, but I don't see it being used on the receiving end of the connection when the `connection` event triggers on the server.

packages/lisk-p2p/src/peer.ts
ishantiw: Ah, I was using it for debugging but forgot to change it, but now we don't need it as it will be handled within Peer object

packages/lisk-p2p/src/peer.ts
ishantiw: will be handled in Peer object now

packages/lisk-p2p/src/peer.ts
ishantiw: same as above

packages/lisk-p2p/src/peer.ts
ishantiw: Now we are reusing the socket and creating a peer object

packages/lisk-p2p/src/peer.ts
ishantiw: Removed

packages/lisk-p2p/src/peer.ts
ishantiw: I haven't tested it yet but I will cover this in #1175 as @shuse2 already pointed this out there

packages/lisk-p2p/src/peer.ts
ishantiw: I have made the changes as discussed earlier

packages/lisk-p2p/src/peer_discovery.ts
ishantiw: Changed as per discussion 👍

packages/lisk-p2p/src/peer.ts
ishantiw: This field is no longer needed with the new approach

packages/lisk-p2p/src/p2p.ts
ishantiw: I'm returning the error instead of throwing and we can log or emit it if we want to

packages/lisk-p2p/src/p2p.ts
```diff
@@ -462,15 +458,40 @@ export class P2P extends EventEmitter {
...
+			this._config.seedPeers,
+		);
+		// Add seed's peerinfos in tried peer as we already tried them to fetch status
+		seedPeerInfos.forEach(seedInfo => {
+			const peerId = constructPeerIdFromPeerInfo(seedInfo);
+			if (!this._triedPeers.has(peerId)) {
+				this._triedPeers.set(peerId, seedInfo);
+			}
+		});
 		// TODO: Once we will a new peer discovery then we can remove this typecasting to P2PDiscoveredPeerInfo
```
shuse2: maybe remove this comment?

packages/lisk-p2p/src/peer_pool.ts
```diff
@@ -253,33 +256,74 @@ export class PeerPool extends EventEmitter {
...
 		);
 
 		const discoveredPeers = await discoverPeers(peerSampleToProbe, {
 			blacklist: blacklist.map(peer => peer.ipAddress),
 		});
 
+		// Check for received discovery info and then find it in peer pool and then update it
+		discoveredPeers.forEach((peerInfo: P2PDiscoveredPeerInfo) => {
+			const peerId = constructPeerIdFromPeerInfo(peerInfo);
+			if (this.hasPeer(peerId)) {
```
shuse2: instead of calling `hasPeer` and casting `getPeer` result, maybe just call `getPeer` and check undefined?

packages/lisk-p2p/src/peer_pool.ts
```diff
@@ -293,6 +337,11 @@ export class PeerPool extends EventEmitter {
 			if (!this.hasPeer(peerId)) {
 				this.addPeer(peerInfo);
 			}
+			const existingPeer = this.getPeer(peerId) as Peer;
```
shuse2: same as above

packages/lisk-p2p/src/peer_pool.ts
```diff
@@ -253,33 +256,74 @@ export class PeerPool extends EventEmitter {
...
-					return existingPeer;
-				}
-
-				return this.addPeer(peerInfo);
-			},
-		);
+		const peersForDiscovery = knownPeers.map(peerInfo => {
+			const peerId = constructPeerIdFromPeerInfo(peerInfo);
+			if (this.hasPeer(peerId)) {
+				const existingPeer = this.getPeer(peerId) as Peer;
```
shuse2: same as below

shuse2 (changes requested):

packages/lisk-p2p/src/peer.ts
ishantiw: This issue will now be covered in this PR. Please ignore the previous comment.

packages/lisk-p2p/src/peer.ts
```diff
@@ -316,6 +316,13 @@ export class Peer extends EventEmitter {
 		this._outboundSocket.connect();
 	}
 
+	public createPeerFromOutboundConnection(
```
jondubois: We should pass the outbound socket to the peer's constructor as an optional argument instead of having a separate `createPeerFromOutboundConnection` method. This is dangerous here because it could be called multiple times. If we move this logic to the constructor then we guarantee that this cannot happen.

The logic which instantiates the `Peer` can just set the `inboundSocket` as `undefined` explicitly if there is no inbound socket.

packages/lisk-p2p/src/peer_pool.ts
```diff
@@ -253,33 +256,74 @@ export class PeerPool extends EventEmitter {
...
 	}
 
+	public async fetchStatusAndCreatePeer(
+		seedPeers: ReadonlyArray<P2PPeerInfo>,
+		nodeInfo: P2PNodeInfo,
+		peerConfig: PeerConfig,
+	): Promise<ReadonlyArray<P2PDiscoveredPeerInfo>> {
+		const seedPeerUpdatedInfosAndSocket = await Promise.all(
+			seedPeers.map(async seedPeer =>
+				connectAndFetchPeerInfo(seedPeer, nodeInfo, peerConfig),
```
jondubois: Because `connectAndFetchPeerInfo` is a pass or fail operation, it should throw; that way the function signature doesn't need to return dual types and we don't need to use `instanceof` statement to separate the two outputs.

The `fetchStatusAndCreatePeer` function can succeed even if it has errors, so returning the error should happen at this level. Maybe this function could return an object with an `errors` and a `peerInfoList` field. If no errors occurred, then the `errors` field would be an empty array. Also, it doesn't seem that we are using the errors from `connectAndFetchPeerInfo` anywhere at the moment (and this function is not returning them).

We should emit them on the top level `P2P` instance using the existing `EVENT_FAILED_TO_FETCH_PEER_INFO` event.

packages/lisk-p2p/src/peer_pool.ts
```diff
@@ -366,6 +415,42 @@ export class PeerPool extends EventEmitter {
...
+
+			return false;
+		}
+
+		const peerConfig = {
+			connectTimeout: this._peerPoolConfig.connectTimeout,
+			ackTimeout: this._peerPoolConfig.ackTimeout,
+		};
+		const peer = new Peer(peerInfo, peerConfig);
+		peer.createPeerFromOutboundConnection(socket);
```
jondubois: As mentioned in a previous comment. It would be safer to pass the outbound socket to the `Peer` constructor.

jondubois (changes requested):

packages/lisk-p2p/src/peer_pool.ts
```diff
@@ -366,6 +422,41 @@ export class PeerPool extends EventEmitter {
...
+			return false;
+		}
+
+		const peerConfig = {
+			connectTimeout: this._peerPoolConfig.connectTimeout,
+			ackTimeout: this._peerPoolConfig.ackTimeout,
+		};
+		const peer = new Peer(peerInfo, peerConfig, { outbound: socket });
+
+		// Throw an error because adding a peer multiple times is a common developer error which is very difficult to itentify and debug.
```
shuse2: where does it throw an error?

shuse2 (commented):
Just small question

packages/lisk-p2p/src/peer_pool.ts
ishantiw: Oh, it was a mistake, we don;t need to throw an error here as we are already checking above if it's present in peer pool or not. I corrected it

packages/lisk-p2p/src/peer.ts
```diff
@@ -713,6 +713,6 @@ export const connectAndFetchPeerInfo = async (
 
 		return { peerInfo, socket };
 	} catch (error) {
-		return error;
+		throw error;
```
jondubois: Since the error is just being re-thrown as-is and we're not doing anything else with it here, we don't need the try-catch block in this function.

packages/lisk-p2p/src/peer_pool.ts
```diff
@@ -253,33 +256,79 @@ export class PeerPool extends EventEmitter {
 		});
 	}
 
+	public async fetchStatusAndCreatePeer(
```
jondubois: This method name should be plural because it's fetching and adding multiple peers.

jondubois (changes requested):

packages/lisk-p2p/src/peer.ts
ishantiw: I think you commented on the outdated code, but now I'm throwing an error but capturing peer ip and port for better logging

jondubois: It appears that you already fixed this in a later commit so ignore.

jondubois (approved):

shuse2 (approved):

PR #1172, opened 2019-03-11
Experiment for new blockchain/dpos module and DB idea

shuse2:
### Description
This branch is mainly  used for forming idea and benchmarks for database options to process blocks.
Also, to be used as very initial idea of blockchain/dpos library

To run the sample app,
```
npm install
npm run build
cd packages/lisk-node
npx ts-node src
```


packages/lisk-node/src/application.ts
```diff
@@ -0,0 +1,104 @@
...
+		);
+		this._logger.info('Genesis block file obtained');
+		const genesis = JSON.parse(genesisStr);
+		this._p2p = new P2P({
+			blacklistedPeers: [],
+			connectTimeout: 5000,
+			ackTimeout: 15000,
+			seedPeers: [
+				{
+					ipAddress: '83.136.254.92',
```
pablitovicente: shouldn't this use a configuration value?

packages/lisk-node/src/application.ts
```diff
@@ -0,0 +1,104 @@
...
+					wsPort: 8001,
+				},
+				{
+					ipAddress: '83.136.249.76',
+					wsPort: 8001,
+				},
```
pablitovicente, line 44: shouldn't these values use configuration?

packages/lisk-node/src/application.ts
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	public async init(): Promise<void> {
+		this._logger.info('Starting initialization');
+		await this._blockchain.init();
+		this._logger.info('Blockchain initialized.');
+		await this._dpos.init(this._blockchain.lastBlock);
+		this._logger.info('DPOS initialized.');
+		this._initialized = true;
+		this._logger.info('Finished initialized.');
```
pablitovicente, line 89: 'Finished Initialisation.'

PR #1178, opened 2019-03-12
Improve error message and not throw in constructor - Closes #1164

### LINKED ISSUE #1164: Improve returned errors
### Expected behavior

When calling `[transaction type class].validate()` if errors are detected it should return descriptive errors instead of only indicating what property is at fault. 

The error array contains a lot of useful information for the user about what's wrong with a transaction.
Example: 

```
{ Transaction: 222675625422353767 failed at .id: Invalid transaction id, actual: 7707413704731182881, expected: 222675625422353767
    ...
    jse_shortmsg: 'Invalid transaction id',
    jse_info: {},
    message: 'Invalid transaction id',
    name: 'TransactionError',
    id: '222675625422353767',
    dataPath: '.id',
    actual: '7707413704731182881',
    expected: '222675625422353767' },
```
`validate()` should expose all information that might be relevant for the user.

### Actual behavior

Currently `[transaction type class].validate()` is not returning good enough errors.

### Steps to reproduce

Run an script with: 

```
import { TransferTransaction } from './src/index';
const transferTrs1 = { id: '222675625422353767',
  type: 0,
  amount: '100',
  fee: '10',
  senderPublicKey:
   '2ca9a7143fc721fdc540fef893b27e8d648d2288efa61e56264edf01a2c23079',
  recipientId: '12668885769632475474L',
  timestamp: 28227090,
  asset: {},
  signature:
  '2821d93a742c4edf5fd960efad41a4def7bf0fd0f7c09869aed524f6f52bf9c97a617095e2c712bd28b4279078a29509b339ac55187854006591aa759784c205' 
 };
const trs1 = new TransferTransaction(transferTrs1);
console.log(trs1.validate().errors);
```
and check available output

shuse2:
### What was the problem?
Error message was inconsistent, and not having enough information.
Also, constructor was throwing error with multiple error, which was harder to handle.

### How did I fix it?
- Change constructor not to throw
- Added expected and actual where it make sense

### How to test it?
Create invalid transaction and check if it doesn't throw

### Review checklist

* The PR resolves #1164 
* Linting passes
* Tests pass

### DIFF
--- a/packages/lisk-transactions/src/0_transfer_transaction.ts
+++ b/packages/lisk-transactions/src/0_transfer_transaction.ts
@@ -20,7 +20,7 @@ import {
 	StateStorePrepare,
 } from './base_transaction';
 import { MAX_TRANSACTION_AMOUNT, TRANSFER_FEE } from './constants';
-import { TransactionError, TransactionMultiError } from './errors';
+import { TransactionError } from './errors';
 import { TransactionJSON } from './transaction_types';
 import {
 	validateAddress,
@@ -36,15 +36,6 @@ export interface TransferAsset {
 	readonly data: string;
 }
 
-export const transferAssetTypeSchema = {
-	type: 'object',
-	properties: {
-		data: {
-			type: 'string',
-		},
-	},
-};
-
 export const transferAssetFormatSchema = {
 	type: 'object',
 	properties: {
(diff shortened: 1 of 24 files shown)


pablitovicente (dismissed):

packages/lisk-transactions/src/0_transfer_transaction.ts
```diff
@@ -161,20 +150,34 @@ export class TransferTransaction extends BaseTransaction {
...
 			this.recipientPublicKey &&
 			this.recipientId !== getAddressFromPublicKey(this.recipientPublicKey)
 		) {
-			errors.push(
-				new TransactionError(
-					'recipientId does not match recipientPublicKey.',
-					this.id,
-					'.recipientId',
-				),
+			const calculatedAddress = getAddressFromPublicKey(
```
mitsuaki-u, line 164: maybe we can remove `this.recipientId !== getAddressFromPublicKey(this.recipientPublicKey)` check in above if statement since we are again checking it below.

packages/lisk-transactions/src/base_transaction.ts
```diff
@@ -134,33 +133,32 @@ export abstract class BaseTransaction {
...
-				'Invalid field types',
-				rawTransaction.id,
-				convertToTransactionError(rawTransaction.id || '', validator.errors),
-			);
+	// tslint:disable-next-line cyclomatic-complexity
+	public constructor(rawTransaction: unknown) {
+		const tx = (typeof rawTransaction === 'object' && rawTransaction !== null
+			? rawTransaction
+			: {}) as Partial<TransactionJSON>;
+		this.amount = new BigNum(tx.amount || '0');
```
mitsuaki-u: should we check typeof `amount` (and also `fee` below) before instantiating BigNum, as there is no longer a type check and it could be a non-number string which throws an error.

mitsuaki-u (changes requested):

packages/lisk-transactions/src/base_transaction.ts
```diff
@@ -138,8 +138,18 @@ export abstract class BaseTransaction {
 		const tx = (typeof rawTransaction === 'object' && rawTransaction !== null
 			? rawTransaction
 			: {}) as Partial<TransactionJSON>;
-		this.amount = new BigNum(tx.amount || '0');
-		this.fee = new BigNum(tx.fee || '0');
+		this.amount = new BigNum(
+			(typeof tx.amount === 'string' && tx.amount !== '') ||
```
mitsuaki-u: also check for `new BigNum('notanumber')` as that seems to throw error at least on non-forked original version.

mitsuaki-u (approved):

PR #1180, opened 2019-03-12
Fix timestamp validation - Closes #1179

### LINKED ISSUE #1179: Negative timestamp should be allowed in transaction
### Expected behavior
Currently protocol allows negative timestamp

### Actual behavior
timestamp has to be positive integer

2.1.0-alpha.2-

shuse2:
### What was the problem?
Current protocol allows negative timestamp within int32

### How did I fix it?
Update the schema to allow negative number

### How to test it?
validate transaction with negative timestamp

### Review checklist

* The PR resolves #1179
* Linting passes
* Tests pass

### DIFF
--- a/packages/lisk-transactions/src/utils/validation/schema.ts
+++ b/packages/lisk-transactions/src/utils/validation/schema.ts
@@ -103,7 +103,7 @@ export const baseTransaction = {
 		},
 		timestamp: {
 			type: 'integer',
-			minimum: 0,
+			minimum: -2147483648,
 			maximum: 2147483647,
 		},
 		senderId: {
@@ -122,7 +122,7 @@ export const baseTransaction = {
 			type: 'string',
 		},
 		recipientPublicKey: {
-			type: ['string'],
+			type: 'string',
 			format: 'emptyOrPublicKey',
 		},
 		signature: {


packages/lisk-transactions/src/utils/validation/schema.ts
```diff
@@ -103,7 +103,7 @@ export const baseTransaction = {
 		},
 		timestamp: {
 			type: 'integer',
-			minimum: 0,
+			minimum: -2147483648,
```
SargeKhan, line 106: Does it mean that we always accept transactions with negative timestamps or do we allow negative timestamps for a few transactions which are exceptions in the blockchain?

shuse2: It seems like there is no exception and allowing negative number per current implementation.
https://github.com/LiskHQ/lisk/blob/development/framework/src/modules/http_api/schema/swagger.yml#L1278-L1283
https://github.com/LiskHQ/lisk/blob/development/framework/src/modules/chain/logic/transaction.js#L688-L696

SargeKhan (approved):

pablitovicente (approved):

PR #1183, opened 2019-03-13
Create addMultisignature function in BaseTransaction - Closes #1176

### LINKED ISSUE #1176: Create addMultisignature function in BaseTransaction
### Expected behavior
The transaction instances should include a function definition for verifying and adding a multisignature for a transaction.
### Actual behavior
Transactions only include `addVerifiedMultisignature` which expects that the signature is already verified and a `processMultisignature` function which verifies all signatures for the transaction.

2.1.0

pablitovicente:
### What was the problem?

There was no method for adding signatures to pending transactions.

### How did I fix it?

By adding the method `addMultisignatures` to `BaseTransaction` and `MultisignatureTransaction`

### How to test it?

- Tests in `test/base_transaction.ts`
- Build should be green

### Review checklist

* The PR resolves #1176
* Linting passes
* Tests pass


packages/lisk-transactions/src/4_multisignature_transaction.ts
```diff
@@ -331,4 +332,65 @@ export class MultisignatureTransaction extends BaseTransaction {
 
 		return [];
 	}
+
+	public addMultisignature(
+		store: StateStore,
```
shuse2, line 340: for registration of multisignature, it should be `_: StateStore` and not used

packages/lisk-transactions/src/4_multisignature_transaction.ts
```diff
@@ -331,4 +332,65 @@ export class MultisignatureTransaction extends BaseTransaction {
...
+						signatureObject.signature
+					}' already present in transaction.`,
+					this.id,
+				),
+			]);
+		}
+
+		// Validate signature key belongs to pending multisig registration transaction
+		// tslint:disable-next-line
+		const keysgroup = (this as any).asset.multisignature.keysgroup.map(
```
shuse2: why do we need `this as any`? prefer not to use any

packages/lisk-transactions/src/4_multisignature_transaction.ts
```diff
@@ -331,4 +332,65 @@ export class MultisignatureTransaction extends BaseTransaction {
...
 	}
+
+	public addMultisignature(
+		store: StateStore,
+		signatureObject: SignatureObject,
+	): TransactionResponse {
+		// Get the account
+		const account = store.account.get(this.senderId);
+		// Check if signature is not already there
+		if (this.signatures.includes(signatureObject.signature)) {
```
shuse2: this check is done in `validateMultisignatures`

packages/lisk-transactions/src/4_multisignature_transaction.ts
```diff
@@ -331,4 +332,65 @@ export class MultisignatureTransaction extends BaseTransaction {
...
+						signatureObject.publicKey
+					}' is not a member for account '${account.address}'.`,
+					this.id,
+				),
+			]);
+		}
+
+		// Validate the signature using the signature sender and transaction details
+		const trsSignature = validateMultisignatures(
+			[signatureObject.publicKey],
```
shuse2: it should include all the keysgroup here

packages/lisk-transactions/src/4_multisignature_transaction.ts
```diff
@@ -331,4 +332,65 @@ export class MultisignatureTransaction extends BaseTransaction {
...
+			[signatureObject.signature],
```
shuse2: all the signatures and `signatureObject.signature` here

packages/lisk-transactions/src/4_multisignature_transaction.ts
```diff
@@ -331,4 +332,65 @@ export class MultisignatureTransaction extends BaseTransaction {
...
+			1,
```
shuse2: min from asset here

packages/lisk-transactions/src/4_multisignature_transaction.ts
```diff
@@ -331,4 +332,65 @@ export class MultisignatureTransaction extends BaseTransaction {
...
+					`Public Key '${
+						signatureObject.publicKey
+					}' is not a member for account '${account.address}'.`,
+					this.id,
+				),
+			]);
+		}
+
+		// Validate the signature using the signature sender and transaction details
+		const trsSignature = validateMultisignatures(
```
shuse2:
```suggestion
		const { valid, errors } = validateMultisignatures(
```

packages/lisk-transactions/src/4_multisignature_transaction.ts
```diff
@@ -331,4 +332,65 @@ export class MultisignatureTransaction extends BaseTransaction {
...
+			[signatureObject.publicKey],
+			[signatureObject.signature],
+			1,
+			this.getBasicBytes(),
+		);
+		// If the signature is valid for the sender push it to the signatures array
+		if (trsSignature.valid) {
```
shuse2: need to check `valid || (errors.length > 0 && instanceof errors[0] === TransactionPendingError)`.
maybe 
```
if (!valid && instanceof errors[0] !== TransactionPendingError) {
 return createResponse(this.id, errors)
}
this.signatures.push(signatureObject.signature);
return this.processMultisignatures();
```

packages/lisk-transactions/src/base_transaction.ts
```diff
@@ -305,6 +307,65 @@ export abstract class BaseTransaction {
...
+			return createResponse(this.id, [
+				new TransactionError(
+					`Public Key '${
+						signatureObject.publicKey
+					}' is not a member for account '${account.address}'.`,
+					this.id,
+				),
+			]);
+		}
+
```
shuse2, line 329: maybe if it's not included in the signatue already, we can just added to the `signatures` and `return this.processMulltisignature`?

shuse2 (changes requested):

packages/lisk-transactions/src/base_transaction.ts
pablitovicente: I'm not sure I understand the issue; if the PK that was sent is not one of the registered with the accounts should be rejected.

packages/lisk-transactions/src/4_multisignature_transaction.ts
pablitovicente: Was needed in base transaction but not here. Removing.

packages/lisk-transactions/src/4_multisignature_transaction.ts
pablitovicente: Removed.

packages/lisk-transactions/src/4_multisignature_transaction.ts
pablitovicente: Changed

pablitovicente:
@shuse2 as this method only adds one signature to the transaction trying to use `validateMultisignatures()` causes issues as that method use case is to check a complete set of signatures for a multi-signature transaction instead of signatures arriving one at a time to a pending transaction that was sent without any signatures or only some of the signatures. Please review the last changes

pablitovicente:
After our talk I will refactor part of validateMultisignatures() into new methods for sharing the validations

pablitovicente:
After moving some of the validations in `validateMultisignatures()` to its own functions still decided to keep the code as we're dealing here with validating a single incoming signature. calling `validateMultisignatures()` should happen after all signatures are available as a final complete validation. @SargeKhan could you please review?

packages/lisk-transactions/src/4_multisignature_transaction.ts
```diff
@@ -331,4 +332,60 @@ export class MultisignatureTransaction extends BaseTransaction {
 
 		return [];
 	}
+
+	public addMultisignature(
+		_: StateStore,
```
SargeKhan: since we are using this variable, we should assign it a proper name.

packages/lisk-transactions/src/base_transaction.ts
SargeKhan: I don't think we should add the signature to the array, and return the `processMultisignature` function because `processMultisignature`  will set the status to invalid for the transaction if the new signature is invalid. Which is not what we want.

packages/lisk-transactions/src/base_transaction.ts
```diff
@@ -305,6 +306,65 @@ export abstract class BaseTransaction {
...
+			signatureObject.signature,
+			this.getBasicBytes(),
+			this.id,
+		);
+		// If the signature is valid for the sender push it to the signatures array
+		if (trsSignature.valid) {
+			this.signatures.push(signatureObject.signature);
+			this.processMultisignatures(store);
+		}
+		// Else pupulate errors
```
SargeKhan: *populate

packages/lisk-transactions/src/base_transaction.ts
```diff
@@ -305,6 +306,65 @@ export abstract class BaseTransaction {
...
+		const trsSignature = validateSignature(
+			signatureObject.publicKey,
+			signatureObject.signature,
+			this.getBasicBytes(),
+			this.id,
+		);
+		// If the signature is valid for the sender push it to the signatures array
+		if (trsSignature.valid) {
+			this.signatures.push(signatureObject.signature);
+			this.processMultisignatures(store);
```
SargeKhan: should we return from here?

packages/lisk-transactions/test/4_multisignature_transaction.ts
```diff
@@ -407,4 +409,99 @@ describe('Multisignature transaction class', () => {
...
+					'bb7ef62be03d5c195a132efe82796420abae04638cd3f6321532a5d33031b30c',
+				signature:
+					'eeee799c2d30d2be6e7b70aa29b57f9b1d6f2801d3fccf5c99623ffe45526104b1f0652c2cb586c7ae201d2557d8041b41b60154f079180bb9b85f8d06b3010c',
+			});
+
+			const expectedError =
+				'Failed to add signature eeee799c2d30d2be6e7b70aa29b57f9b1d6f2801d3fccf5c99623ffe45526104b1f0652c2cb586c7ae201d2557d8041b41b60154f079180bb9b85f8d06b3010c.';
+
+			expect(status).to.eql(Status.FAIL);
+			expect(errors[0].message).to.be.eql(expectedError);
```
SargeKhan, line 485: can you check in this test that the status of the Multisignature transaction is still PENDING and has not changed to FAIL?

packages/lisk-transactions/src/4_multisignature_transaction.ts
pablitovicente: Good catch; I failed to renamed on the last refactor. Will rename.

SargeKhan (commented):
Great work Pablo!

packages/lisk-transactions/test/4_multisignature_transaction.ts
pablitovicente: @SargeKhan added a new test for this as `_multisignatureStatus` is 0 until at least one valid signature is added. Should I create an issue so when a multi-signature transaction arrives without signatures its `_multisignatureStatus` is set to 2 -Pending-?

packages/lisk-transactions/src/base_transaction.ts
```diff
@@ -305,6 +306,66 @@ export abstract class BaseTransaction {
 		return createResponse(this.id, errors);
 	}
 
+	public addMultisignature(
+		store: StateStore,
+		signatureObject: SignatureObject,
```
shuse2, line 311: maybe destructure here?

packages/lisk-transactions/src/base_transaction.ts
```diff
@@ -305,6 +306,66 @@ export abstract class BaseTransaction {
...
+					`Signature '${
+						signatureObject.signature
+					}' already present in transaction.`,
+					this.id,
+				),
+			]);
+		}
+
+		// Validate the signature using the signature sender and transaction details
+		const trsSignature = validateSignature(
```
shuse2: also destructure here?

packages/lisk-transactions/test/4_multisignature_transaction.ts
shuse2: ` when a multi-signature transaction arrives without signatures its _multisignatureStatus is set to 2 -Pending-?`
in case of type 4, i think we should set the status to at least pending because we know it is multi-sig. Lets create an issue for that

packages/lisk-transactions/test/base_transaction.ts
```diff
@@ -628,6 +630,122 @@ describe('Base transaction class', () => {
 		});
 	});
 
+	describe('#addMultisignature', () => {
+		let transferFromMultiSigAccountTrs: TransferTransaction;
+		let multisigMember: SignatureObject;
+		beforeEach(() => {
```
shuse2:
```suggestion
		beforeEach( async () => {
```

packages/lisk-transactions/test/base_transaction.ts
```diff
@@ -628,6 +630,122 @@ describe('Base transaction class', () => {
...
+			multisigMember = {
+				transactionId: transferFromMultiSigAccountTrs.id,
+				publicKey:
+					'542fdc008964eacc580089271353268d655ab5ec2829687aadc278653fad33cf',
+				signature:
+					'f223799c2d30d2be6e7b70aa29b57f9b1d6f2801d3fccf5c99623ffe45526104b1f0652c2cb586c7ae201d2557d8041b41b60154f079180bb9b85f8d06b3010c',
+			};
+		});
+
+		it('should add signature to transaction from multisig account', () => {
```
shuse2:
```suggestion
		it('should add signature to transaction from multisig account', async () => {
```
same for below

packages/lisk-transactions/test/4_multisignature_transaction.ts
```diff
@@ -407,4 +409,119 @@ describe('Multisignature transaction class', () => {
...
+				publicKey:
+					'cba7d88c54f3844bbab2c64b712e0ba3144921fe7a76c5f9df80b28ab702a35b',
+				signature:
+					'35d9bca853353906fbc44b86918b64bc0d21daf3ca16e230aa59352976624bc4ce69ac339f08b45c5e926d60cfa81276778e5858ff2bd2290e40d9da59cc5f0b',
+			};
+
+			multisigTrs.addMultisignature(store, nonMemberSignature);
+
+			expect(arrangeStatus).to.eql(Status.PENDING);
+			expect(multisigTrs['_multisignatureStatus']).to.eql(Status.PENDING);
```
shuse2:
```suggestion
			expect((multisigTrs as any)._multisignatureStatus).to.eql(Status.PENDING);
```

packages/lisk-transactions/test/4_multisignature_transaction.ts
```diff
@@ -407,4 +409,119 @@ describe('Multisignature transaction class', () => {
...
+				transactionId: multisigTrs.id,
+				publicKey:
+					'cba7d88c54f3844bbab2c64b712e0ba3144921fe7a76c5f9df80b28ab702a35b',
+				signature:
+					'35d9bca853353906fbc44b86918b64bc0d21daf3ca16e230aa59352976624bc4ce69ac339f08b45c5e926d60cfa81276778e5858ff2bd2290e40d9da59cc5f0b',
+			};
+
+			multisigTrs.addMultisignature(store, nonMemberSignature);
+
+			expect(arrangeStatus).to.eql(Status.PENDING);
```
shuse2, line 527: Also, it would be nice to check length of `signatures` didn't change

shuse2 (changes requested):

packages/lisk-transactions/test/4_multisignature_transaction.ts
pablitovicente: Created https://github.com/LiskHQ/lisk-elements/issues/1192

packages/lisk-transactions/test/4_multisignature_transaction.ts
pablitovicente: I originally had it casted as any but I thought we wanted to avoid the use of `any` as much as possible?

shuse2: only for `test` it's ok =)

packages/lisk-transactions/src/base_transaction.ts
pablitovicente: You mean the function parameters? It seems that in TS using destructuring assignment is more verbose than in JS should we reserve its use to methods with many parameters only?

shuse2 (approved):

mitsuaki-u (approved):

SargeKhan (approved):

PR #1187, opened 2019-03-14
PeerPool should not respond to get peers list - Closes#1186

### LINKED ISSUE #1186: PeerPool should not respond to get peers list
### Expected behavior

Whenever we receive `list` RPC call we should respond with a peer list that includes the peers from `newPeers` and `triedPeers`.

### Actual behavior

Currently, PeerPool is replying to this RPC with all the connected nodes that don't include sometimes the peers from `newPeers` or `triedPeers` that are not connected.

ishantiw:
### Description

Whenever we receive `list` RPC call we should respond with a peer list that includes the peers from `newPeers` and `triedPeers` and this RPC request should be handled in P2P.

This PR should go after #1170 

### Review checklist

* The PR resolves #1186 
* Linting passes
* Tests pass


packages/lisk-p2p/src/p2p.ts
```diff
@@ -92,6 +101,11 @@ export const DEFAULT_DISCOVERY_INTERVAL = 30000;
 
 const BASE_10_RADIX = 10;
 
+const selectRandomPeerSample = (
```
jondubois, line 106: This function is now in two places; here and in peerPool. We should try to reuse. Maybe we can export the one from the PeerPool and use that one?

packages/lisk-p2p/src/p2p.ts
```diff
@@ -480,6 +497,61 @@ export class P2P extends EventEmitter {
...
+				.map(
+					(peerInfo: P2PDiscoveredPeerInfo): ProtocolPeerInfo | undefined => {
+						const peerDiscoveredInfo = peerInfo;
+
+						if (!peerDiscoveredInfo) {
+							return undefined;
+						}
+
+						// The options property is not read by the current legacy protocol but it should be added anyway for future compatibility.
+						return {
```
jondubois: This is not specific to this PR but this logic can be simplified as we discussed (this object should also include custom properties).

jondubois (changes requested):

packages/lisk-p2p/src/p2p.ts
ishantiw: Oh actually there is a small difference, this one is using `P2PDiscoveredPeer` and the other one is using `Peer` objects. Or maybe, I can make it a general function that only takes `P2PDiscoveredPeer` and in PeerPool I pass the list of peerinfos

packages/lisk-p2p/src/p2p.ts
```diff
@@ -15,14 +15,20 @@
 
 import { EventEmitter } from 'events';
 import * as http from 'http';
+// tslint:disable-next-line no-require-imports
+import shuffle = require('lodash.shuffle');
```
SargeKhan, line 19: why don't we do something like: `import { shuffle } from 'lodash';`?

packages/lisk-p2p/src/p2p.ts
```diff
@@ -480,6 +497,61 @@ export class P2P extends EventEmitter {
...
+								: undefined,
+							wsPort: peerDiscoveredInfo.wsPort,
+						};
+					},
+				)
+				.filter(
+					(peerDetailedInfo: ProtocolPeerInfo | undefined) =>
+						!!peerDetailedInfo,
+				)
+				.map(
```
SargeKhan: Will typescript remove this `map` function after it transpiles it to javascript?

packages/lisk-p2p/src/p2p.ts
ishantiw: It was used in `PeerPool` like this and I just moved it here, but you are right I will change it to `from` syntax and update `package.json`

packages/lisk-p2p/src/p2p.ts
ishantiw: No, it doesn't

SargeKhan: Then isn't it better to remove this map function and cast the whole array object as: 
```
peers: (...computation ) as ReadonlyArray< ProtocolPeerInfo> 
```
?

ishantiw: Good catch, I removed this map and typecasted it directly :)

packages/lisk-p2p/src/p2p.ts
shuse2: @SargeKhan @ishantiw 
so there are two reason
1. we want to minimize the usage of library/lodash
2. if we use only `lodash.shuffle`, `import shuffle` doesn't work because of the type definition, so this was the only way ='(

ishantiw:
Addressed

packages/lisk-p2p/src/p2p.ts
```diff
@@ -480,6 +497,52 @@ export class P2P extends EventEmitter {
...
+	): ReadonlyArray<P2PDiscoveredPeerInfo> {
+		const discoveredPeerList: ReadonlyArray<P2PDiscoveredPeerInfo> = [
+			...this._newPeers.values(),
+			...this._triedPeers.values(),
+		]; // Peers whose values has been updated atleast once.
+
+		return selectRandomPeerSample(discoveredPeerList, count);
+	}
+
+	private _handleGetAllPeersRequest(request: P2PRequest): void {
```
shuse2: Not strictly related to this but function name should be `_handleGetPeersRequest`?

packages/lisk-p2p/src/p2p.ts
```diff
@@ -480,6 +497,52 @@ export class P2P extends EventEmitter {
...
+		// TODO later: Remove fields that are specific to the current Lisk protocol.
+		const protocolPeerInfoList: ProtocolPeerInfoList = {
+			success: true,
+			peers: this._pickRandomDiscoveredPeers(MAX_PEER_LIST_BATCH_SIZE)
+				.map(
+					(peerInfo: P2PDiscoveredPeerInfo): ProtocolPeerInfo | undefined => {
```
shuse2: would it be possible to have `undefined` here? seems like `_pickRandomDiscoveredPeers` must return array

packages/lisk-p2p/src/p2p.ts
```diff
@@ -480,6 +497,52 @@ export class P2P extends EventEmitter {
...
+						if (!peerDiscoveredInfo) {
+							return undefined;
+						}
+						const { ipAddress, ...peerInfoWithoutIp } = peerDiscoveredInfo;
+
+						// The options property is not read by the current legacy protocol but it should be added anyway for future compatibility.
+						return {
+							...peerInfoWithoutIp,
+							ip: ipAddress,
+							broadhash: peerInfoWithoutIp.broadhash
```
shuse2: do we need to overwrite these property?
it seems like if it exists, it will be deconstructed from `peerInfoWithoutIp`

packages/lisk-p2p/src/p2p.ts
```diff
@@ -480,6 +497,52 @@ export class P2P extends EventEmitter {
...
+	}
+
+	private _handleGetAllPeersRequest(request: P2PRequest): void {
+		// TODO later: Remove fields that are specific to the current Lisk protocol.
+		const protocolPeerInfoList: ProtocolPeerInfoList = {
+			success: true,
+			peers: this._pickRandomDiscoveredPeers(MAX_PEER_LIST_BATCH_SIZE)
+				.map(
+					(peerInfo: P2PDiscoveredPeerInfo): ProtocolPeerInfo | undefined => {
+						const peerDiscoveredInfo = peerInfo;
```
shuse2: why changing the veriable name?

shuse2 (changes requested):

packages/lisk-p2p/src/p2p.ts
ishantiw: `ip` is by the name `ipAddress`, and `broadhash` and `nonce` are not required fields, they are unknown fields and we want to make sure to send these values as part of handshake.

packages/lisk-p2p/src/p2p.ts
ishantiw: Yeah it was needed when we had two types peer types but not now, actually, this function was not updated, I just moved it from PeerPool but it had to change during previous PRs related to peer type changes.

ishantiw:
Addressed @shuse2 @jondubois Waiting for your review

shuse2 (approved):

packages/lisk-p2p/src/p2p.ts
```diff
@@ -494,6 +511,44 @@ export class P2P extends EventEmitter {
...
+				const { ipAddress, ...peerInfoWithoutIp } = peerInfo;
+
+				// The options property is not read by the current legacy protocol but it should be added anyway for future compatibility.
+				return {
+					...peerInfoWithoutIp,
+					ip: ipAddress,
+					broadhash: peerInfoWithoutIp.broadhash
+						? (peerInfoWithoutIp.broadhash as string)
+						: '',
+					nonce: peerInfoWithoutIp.nonce
```
jondubois, line 538: Is it necessary to default `broadhash` and `nonce` to empty strings if not provided?

jondubois (approved):
Please check the comment. If you're happy with it then it's good to merge.

packages/lisk-p2p/src/p2p.ts
ishantiw: Actually, `ProtocolPeerInfo` has broadhash and nonce as required field so I need to default  it to accept because its unknown type in P2PDiscoveredPeer.

ishantiw: So either we can allow it to accept `undefined` also or else we provide a default value,  if `undefined.`

PR #1189, opened 2019-03-15
Check for network, version, protocol compatibility during handshake - Closes#1137

### LINKED ISSUE #1137: Check for network, version, protocol compatibility during handshake
### Expected behavior

We need to check network and protocol compatibility during the handshake in lisk-p2p.

### Actual behavior

Right now we don't have a check for network and protocol compatibility during the handshake.

### Steps to reproduce

N/A

ishantiw:
### Description

Add a field in P2P config to allow compatibility check functions to be inserted in the constructor. The checks will cover,

- Network compatibility (`nethash`)
- Version compatibility (`version`)
- Protocol version compatibility (`protocolVersion`)

### Review checklist

* The PR resolves #1137 
* Linting passes
* Tests pass

### DIFF
--- a/packages/lisk-p2p/src/p2p.ts
+++ b/packages/lisk-p2p/src/p2p.ts
@@ -25,12 +25,8 @@ interface SCServerUpdated extends SCServer {
 import { constructPeerId, constructPeerIdFromPeerInfo } from './peer';
 
 import {
-	INCOMPATIBLE_NETWORK_CODE,
-	INCOMPATIBLE_NETWORK_REASON,
-	INCOMPATIBLE_PROTOCOL_VERSION_CODE,
-	INCOMPATIBLE_PROTOCOL_VERSION_REASON,
-	INCOMPATIBLE_VERSION_CODE,
-	INCOMPATIBLE_VERSION_REASON,
+	INCOMPATIBLE_PEER_CODE,
+	INCOMPATIBLE_PEER_UNKNOWN_REASON,
 	INVALID_CONNECTION_QUERY_CODE,
 	INVALID_CONNECTION_QUERY_REASON,
 	INVALID_CONNECTION_URL_CODE,
@@ -40,6 +36,7 @@ import {
 import { PeerInboundHandshakeError } from './errors';
 
 import {
+	P2PCheckPeerCompatibility,
 	P2PClosePacket,
 	P2PConfig,
 	P2PDiscoveredPeerInfo,
@@ -47,9 +44,6 @@ import {
 	P2PNetworkStatus,
 	P2PNodeInfo,
 	P2PPeerInfo,
-	P2PPeerNetworkCompatibility,
-	P2PPeerProtocolVersionCompatibility,
-	P2PPeerVersionCompatibility,
 	P2PPenalty,
 	P2PRequestPacket,
 	P2PResponsePacket,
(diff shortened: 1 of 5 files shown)


2snEM6 (dismissed):

packages/lisk-p2p/src/p2p_types.ts
```diff
@@ -113,6 +114,18 @@ export type P2PPeerSelectionForConnection = (
 	nodeInfo?: P2PNodeInfo,
 ) => ReadonlyArray<P2PDiscoveredPeerInfo>;
 
+export interface P2PPeerCheckCompatibility {
+	readonly networkCompatible?: P2PPeerNetworkCompatibility;
+	readonly protocolVersionCompatible?: P2PPeerProtocolVersionCompatibility;
+	readonly versionCompatible?: P2PPeerVersionCompatibility;
+}
+
+export type P2PPeerNetworkCompatibility = (nethash: string) => boolean;
```
shuse2: instead of specifying one by one, maybe
`P2PPeerNetworkCompatibility = (header: QueryObject) => boolean`
and i think we can provide the default by
checking nethash, version, and protocol version, and override them if it's provided?

packages/lisk-p2p/src/p2p.ts
```diff
@@ -342,6 +350,76 @@ export class P2P extends EventEmitter {
 					return;
 				}
 
+				if (this._peerHandShakeChecks) {
+					if (this._peerHandShakeChecks.networkCompatible) {
```
shuse2: one `if` for both checking and executing should be fine?

shuse2 (changes requested):

packages/lisk-p2p/src/p2p_types.ts
ishantiw: Although I have added this check now. But by default, all the nodes need to be compatible with each other in terms of minVersion, version and nethash which should be handled in module level in IMO. Now anyone using p2p library needs to pass these values. @jondubois What do you think?

jondubois: I think it would be simpler if there was a single compatibility function instead of three different ones. The compatibility function could either throw or return a custom error on failure which can be handled inside the P2P class; then the `error.message` from the function could be used as the disconnect reason; that way the user can make the error reason be anything; so it gives them more flexibility.

Also, I think that instead of `QueryObject`, we could go one step further and just pass `incomingPeerInfo` (`P2PDiscoveredPeerInfo`) to the function.

ishantiw: After the discussion with @jondubois, we decided to keep `networkCompatibility` and `protocolVersionCompatibility` checks so that we avoid peers from other networks to connect by default. Also, we are going to have one function for all the compatibility checks so that user is not confined to only these 3 checks but also could provide its own check.

packages/lisk-p2p/src/p2p.ts
```diff
@@ -228,6 +231,10 @@ export class P2P extends EventEmitter {
 		this._discoveryInterval = config.discoveryInterval
 			? config.discoveryInterval
 			: DEFAULT_DISCOVERY_INTERVAL;
+
+		this._peerHandshakeChecks = config.peerhandShakeChecks
```
jondubois: Inconsistent capitalization: `peerhandShakeChecks` should be `peerHandshakeChecks`.

I prefer `peerHandshakeCheck` (singular) but this is minor.

jondubois (changes requested):

jondubois (approved):

shuse2 (approved):

PR #1204, opened 2019-03-28
Convert optional field null to undefined

### LINKED ISSUE #1203: Non required properties for transactions should be allowed to be null/undefined
### Expected behavior
The current protocol allows non-required properties to have null/undefined value for all non-required properties whereas lisk-transactions throws an error in case, say if a property value is `null` it was expected to be `string`. 
### Actual behavior
The `validate` step for transaction throws an error if a property is `null` or `undefined`. 
release/2.1.0

shuse2:
### What was the problem?
Optional field should accept null and undefined for the validation

### How did I fix it?
In constructor, change the null to undefined

### How to test it?
npm t

### Review checklist

* The PR resolves #1203 
* Linting passes
* Tests pass

### DIFF
--- a/packages/lisk-transactions/src/5_dapp_transaction.ts
+++ b/packages/lisk-transactions/src/5_dapp_transaction.ts
@@ -27,12 +27,15 @@ const TRANSACTION_DAPP_TYPE = 5;
 export interface DappAsset {
 	readonly dapp: {
 		readonly category: number;
-		readonly description?: string;
-		readonly icon?: string;
 		readonly link: string;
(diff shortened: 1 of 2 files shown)


SargeKhan (approved):

SargeKhan:
@shuse2 , did you check other transaction types? I think we should do this for all transaction definitions.

shuse2:
@SargeKhan 
Yes, base transaction is covered, and for asset, I think there is no optional field for other transactions

mitsuaki-u (approved):

PR #1211, opened 2019-04-04
Integration tests failing randomly due to low timeout - Closes#1103

### LINKED ISSUE #1103: P2P Integration test randomly fail
### Expected behavior
Integration test should be deterministic and not fail

### Actual behavior
Jenkins still fails at below
```
Integration tests for P2P library
Partially connected network which becomes fully connected: All nodes launch at the same time. The seedPeers list of each node contains the next node in the sequence. Discovery interval runs multiple times.
Peer discovery
should discover all peers in the network after a few cycles of discovery:
AssertionError: expected [ Array(9) ] to deeply equal [ Array(10) ]

at p2pNodeList.forEach.p2p (test/integration/p2p.ts:198:30)
at Array.forEach (<anonymous>)
at Context.it (test/integration/p2p.ts:192:17)
```

ishantiw:
### What was the problem?

Integration tests are failing randomly in p2p.

### How did I fix it?

For partial network test, the value of `ackTimeout` and `connectTimeout` was very low (100ms) and sometimes it was not enough time to connect with other nodes if the network is a bit slow. We have increased the timeouts to 1000ms.

### How to test it?

`npm t`

### Review checklist

* The PR resolves #1103 
* Linting passes
* Tests pass

### DIFF
--- a/packages/lisk-p2p/test/integration/p2p.ts
+++ b/packages/lisk-p2p/test/integration/p2p.ts
@@ -330,6 +330,7 @@ describe('Integration tests for P2P library', () => {
 				return new P2P({
 					blacklistedPeers: [],
 					connectTimeout: 5000,
+					ackTimeout: 5000,
 					seedPeers,
 					wsEngine: 'ws',
 					nodeInfo: {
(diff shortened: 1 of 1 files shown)


ishantiw:
Now its throwing,

`@liskhq/lisk-p2p: ResponseAlreadySentError: A response has already been sent for this request`

when running network test locally. I'm looking into it.

jondubois (approved):

shuse2 (approved):

PR #1216, opened 2019-04-10
Fix validation of asset data - Closes #1214

### LINKED ISSUE #1214: Transaction validate incorrectly throws error from assetToBytes
### Expected behavior

`validate` method for each transaction type should return a `TransactionResponse` with errors array instead of throwing.

### Actual behavior

When asset includes invalid data, `validate` --> `getBasicBytes` --> `assetToBytes` throws an error. 

### Steps to reproduce
```
const invalidTx = { "type": 0, "asset": { "data": { } }, ... };
const transferTx = new TransferTransaction(invalidTx);
transferTx.validate()
```

mitsuaki-u:
### What was the problem?

1. `assetToBytes` and `getBasicBytes` called after schema checks failed causing error.

2. `assetToJSON` was destructuring invalid assets in initialization stage, causing errors.

3. `.dataPath` in errors were missing when asset themselves were of invalid type

4. Validation process for dapp and vote transactions were throwing error 

### How did I fix it?

Early return if errors produced from schema and asset validation. 

`assetToJSON` now simply returns itself.

New function specifically for filling missing `.dataPath` in errors.

Fixed initialization and validation issues for dapp and vote transactions. 

### How to test it?

npm run rest

### Review checklist

* The PR resolves #1214 
* Linting passes
* Tests pass

### DIFF
--- a/packages/lisk-transactions/src/5_dapp_transaction.ts
+++ b/packages/lisk-transactions/src/5_dapp_transaction.ts
@@ -145,9 +145,7 @@ export class DappTransaction extends BaseTransaction {
 	}
 
 	public assetToJSON(): object {
-		return {
-			...this.asset,
-		};
+		return this.asset;
 	}
 
 	public async prepare(store: StateStorePrepare): Promise<void> {
--- a/packages/lisk-transactions/src/5_dapp_transaction.ts
+++ b/packages/lisk-transactions/src/5_dapp_transaction.ts
@@ -98,13 +98,14 @@ export class DappTransaction extends BaseTransaction {
 		const tx = (typeof rawTransaction === 'object' && rawTransaction !== null
 			? rawTransaction
 			: {}) as Partial<TransactionJSON>;
-
 		this.asset = (tx.asset || { dapp: {} }) as DappAsset;
-		// If Optional field contains null, converts to undefined
-		this.asset.dapp.description = this.asset.dapp.description || undefined;
-		this.asset.dapp.icon = this.asset.dapp.icon || undefined;
-		this.asset.dapp.tags = this.asset.dapp.tags || undefined;
 		this.containsUniqueData = true;
+		if (this.asset && this.asset.dapp && typeof this.asset.dapp === 'object') {
+			// If Optional field contains null, converts to undefined
+			this.asset.dapp.description = this.asset.dapp.description || undefined;
+			this.asset.dapp.icon = this.asset.dapp.icon || undefined;
+			this.asset.dapp.tags = this.asset.dapp.tags || undefined;
+		}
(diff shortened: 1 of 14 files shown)


packages/lisk-transactions/src/errors.ts
```diff
@@ -87,9 +87,9 @@ export const convertToTransactionError = (
 	return errors.map(
 		error =>
 			new TransactionError(
-				`'${error.dataPath}' ${error.message}`,
+				`'${error.dataPath || '.asset'}' ${error.message}`,
```
shuse2, line 108: At this point, I think non existent `dataPath` doesn't necessarily mean `.asset`

shuse2 (dismissed):

packages/lisk-transactions/src/5_dapp_transaction.ts
```diff
@@ -260,6 +261,11 @@ export class DappTransaction extends BaseTransaction {
 			);
 		}
 		const validLinkSuffix = ['.zip'];
+
+		if (errors.length > 0) {
```
shuse2, line 263: why return here?

packages/lisk-transactions/src/base_transaction.ts
```diff
@@ -233,10 +233,20 @@ export abstract class BaseTransaction {
 	}
 
 	public validate(): TransactionResponse {
-		const errors = [...this._validateSchema(), ...this.validateAsset()];
+		const validateAssetErrors = this.validateAsset();
+		if (validateAssetErrors.length > 0) {
+			return createResponse(this.id, validateAssetErrors);
+		}
+
+		const validateSchemaErrors = this._validateSchema();
```
shuse2: i thought we are not going to split this? what is the purpose of separating this?
I think it's more useful to know all the invalid field (ie: base schema and asset schema validation result together)

shuse2 (dismissed):

packages/lisk-transactions/src/base_transaction.ts
mitsuaki-u: `_validateSchema` calls `toJSON` which in turn calls `assetToJSON`.  If validation of `asset` fails, calling `assetToJSON` throws an error. Therefore we separate here.

packages/lisk-transactions/src/5_dapp_transaction.ts
mitsuaki-u: If asset does not pass validation we should not check fields like `this.asset.dapp.link` below, otherwise it will throw an error.

shuse2 (approved):

shuse2 (approved):

